PR #3, opened 2014-01-22
Timeout granularity

mtstickney:
This change enables less-than-second timeout granularity for wait-for-process-with-timeout by using get-internal-real-time instead of get-universal-time (the same approach is used in Lisp-Dep/mp-nil.lisp).

This change is based on a re-indent and whitespace fixup, which I just did to get a clean patch. Feel free to just cherry-pick the tip commit.

[Review on Reviewable](https://reviewable.io/reviews/robert-strandh/mcclim/3)


dkochmanski:
this PR should be closed (it doesn't work when adapted to bordeaux-threads)

PR #4, opened 2014-01-22
Multiple evaluation

mtstickney:
This commit fixes a small macro bug where with-lock-held and with-recursive-lock-held would evaluate the state form more than once.

This change is based on a re-indent and whitespace fixup, which I really only did to get a clean patch. Feel free to cherry-pick the tip commit.

[Review on Reviewable](https://reviewable.io/reviews/robert-strandh/mcclim/4)


dkochmanski:
this PR may be closed, because now McCLIM uses bordeaux-threads

PR #5, opened 2014-03-05
Don't depend on swank, minor CLX fix

deterministic-arts:
The mcclim.asd file can now be loaded in a mode, which makes it not depend on swank in any form. To use this, push ":clim-without-swank" onto "*features*" prior to asdf:load-op'ing (or ql:quickload'ing) the system.

The CLX backend now passes the ":size 16" parameter do xlib:draw-glyphs in medium-draw-text*. This was done since forever already be medium-draw-glyph, but it seems to have been missing from draw-text*. I think, that this fix is more important than the swank issue, and I can create a separate pull-request, if you don't want to have both topics intermixed.


dkochmanski:
this pull request may be closed – backported these changes in the asd-cleanup branch

PR #6, opened 2014-03-05
The color table is now weak on the values

deterministic-arts:
The hash table used to cache color objects uses weak references for the values on implementations, which support this (and where I know how to implement it), currently CCL and SBCL.


dkochmanski:
this should depend on `trivial-garbage`

dkochmanski:
This should be closed due to merge request #58 (the idea is very good, just replaced the dependency with a portability layer)

PR #17, opened 2016-02-02
Fix on Scigraph

mgi:


gabriel-laddel:
Does this fix the whole of scigraph or are there still issues?

mgi:
Still plenty of issues I think. I've just tested with
(gr:make-demo-frame) and I'm able to see graph, annotate data point with
left click on point but "zoom in" and "describe point", for instance,
end up in the debugger with a xlib:query-pointer on nil.
-- 
Manuel Giraud

PR #18, opened 2016-02-14
Fix several problem with standard gadgets

gas2serra:


gabriel-laddel:
I've not run the tests and demos with these changes applied yet, but having read over all commits: looks good.

gas2serra:
The "List Test" is working.
Improved "Scroll Test".
Improved "Font Test".
Fix activation/deactivation behavior of several standard gadgets.
Fix value and editable-p initargs for text and field pane.

PR #25, opened 2016-05-26
Major asd systems cleanup and codebase housekeeping

dkochmanski:


earl-ducaine:
Oh no! McCLIM isn't standardising on hash colon, e.g.   `(defsystem` #:clouseau blah blah....`

That's almost as hideous, not to mention evil, as using tabs rather than spaces!  :-)

dkochmanski:
What's wrong with not interning the system name in the keyword package?

earl-ducaine:
Haha.... Because it's one extra character to type.  Also, the hash code is hideously ugly and using it practically never makes a difference on keywords. (Mostly, as far as I can remember the convention was introduced to support Lisps like Allegro running in 'modern' mode where the default case of symbols' interned name  is something other than all caps.  Every time I see the convention I feel the need to remind myself why it was sometimes necessary, which I instantly forget again.) And anyway, when the hash is used with keywords and when it does make a difference, you should really consider rewriting your code so that it doesn't.  (To accommodate *my* laziness and stupidity, see above)  Also, pulling a Linus Torvald, The K&R of CL TI and Symbolics never used it in ordinary Lisp code, so it must be the work of Satan. 

But, by far the most important reason not to use it is that some people have an irrational hatred of it (myself) and obviously those people feel their preferences should take precedent over everyone else's.

robert-strandh:
Actually, this convention is a direct consequence of a much more
general rule of software engineering, namely that you should
use the most precise construct that will do the job.  In this case,
The fact that the symbol has no package is a direct signal to the
maintainer that the package is of no importance.

On Fri, May 27, 2016 at 8:50 AM, Earl DuCaine <notifications@github.com>
wrote:

> Haha.... Because it's one extra character to type. Also, the hash code is
> hideously ugly and using it practically never makes a difference on
> keywords. (Mostly, as far as I can remember the convention was introduced
> to support Lisps like Allegro running in 'modern' mode where the default
> case of symbols' interned name is something other than all caps. Every time
> I see the convention I feel the need to remind myself why it was sometimes
> necessary, which I instantly forget again.) And anyway, when the hash is
> used with keywords and when it does make a difference, you should really
> consider rewriting your code so that it doesn't. (To accommodate *my*
> laziness and stupidity, see above) Also, pulling a Linus Torvald, The K&R
> of CL TI and Symbolics never used it in ordinary Lisp code, so it must be
> the work of Satan.
>
> But, by far the most important reason not to use it is that some people
> have an irrational hatred of it (myself) and obviously those people feel
> their preferences should take precedent over everyone else's.
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub
> <https://github.com/robert-strandh/McCLIM/pull/25#issuecomment-222073095>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABYcQYiy8wXjThhg63D2XBEG-xnaROpfks5qFpQZgaJpZM4InOYp>
> .
>



-- 
Robert Strandh

earl-ducaine:
> In this case, The fact that the symbol has no package is a direct signal to the maintainer that the package is of no importance.

That's practically always case with keywords, which makes the hash redundant.  

But, I was mostly kidding.  If someone really hates them so much they can always write an emacs mode to strip them out when editing asd files and add them back in when saving a file that uses the convention.

The changes look great. I'm hoping to check out the new code this weekend.

PR #46, opened 2016-07-01
remove mirrored-sheet-mixin from basic-pane

gas2serra:
I'm working for a backend with a single top-level window. 
The first step was to remove MIRRORED-SHEET-MIXIN from the superclasses of BASIC-PANE. CLX backend creates a dummy class for each BASIC-PANE subclass.

### DIFF
--- a/Backends/CLX/frame-manager.lisp
+++ b/Backends/CLX/frame-manager.lisp
@@ -79,13 +79,42 @@
 ;;; frame-manager. However, in the CLX case, we don't expect there to
 ;;; be any CLX specific panes. CLX uses the default generic panes
 ;;; instead.
+
+;;; if the pane is a subclass of basic-pane and it is not mirrored we create a new class.
+(defun maybe-mirroring (concrete-pane-class)
+  (when (and (not (subtypep concrete-pane-class 'mirrored-sheet-mixin))
+	     (subtypep concrete-pane-class 'basic-pane))
+    (let* ((concrete-pane-class-symbol (if (typep concrete-pane-class 'class)
+                                          (class-name concrete-pane-class)
+                                          concrete-pane-class))
+	   (concrete-mirrored-pane-class (concatenate 'string
+						      "CLX-"
+						      (symbol-name concrete-pane-class-symbol)
+						      "-DUMMY"))
+	   (concrete-mirrored-pane-class-symbol (find-symbol concrete-mirrored-pane-class
+							     :clim-clx)))
+      ;;(format *debug-io* "use dummy mirrored class ~A~%" concrete-mirrored-pane-class)
+      (unless concrete-mirrored-pane-class-symbol
+	(setf concrete-mirrored-pane-class-symbol
+	      (intern concrete-mirrored-pane-class :clim-clx))
+	(eval
+	 `(defclass ,concrete-mirrored-pane-class-symbol
+	      (mirrored-sheet-mixin
+	       ,concrete-pane-class-symbol)
+	    ()
+	    (:metaclass ,(type-of (find-class concrete-pane-class-symbol))))))
+	;;(format *debug-io* "create class ~A~%" concrete-mirrored-pane-class-symbol))
+      (setf concrete-pane-class (find-class concrete-mirrored-pane-class-symbol))))
+  concrete-pane-class)
+
 (defmethod make-pane-1 ((fm clx-frame-manager) (frame application-frame) type &rest args)
   (apply #'make-instance
-	 (find-concrete-pane-class type)
+	 (maybe-mirroring (find-concrete-pane-class type))
 	 :frame frame
 	 :manager fm
 	 :port (port frame)
 	 args))
+
 
 (defmethod adopt-frame :before ((fm clx-frame-manager) (frame menu-frame))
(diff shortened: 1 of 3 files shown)


robert-strandh:
I would like for Daniel to have a look at this pull request before I
merge it, because he is also working on removing MIRRORED-SHEET-MIXIN
from the superclasses of BASIC-PANE.  I need to make sure there is no
conflict with what he is doing.

gas2serra:
Ok.
> x Daniel
Hi Daniel, can I submit you my pull requests? have you a branch where we can collaborate?

dkochmanski:
Hey, sure. Basically it's https://github.com/dkochmanski/McCLIM/tree/develop (note, that some top commits are marked as dirty). How can I contact you? Are you on IRC? Or maybe an e-mail?

dkochmanski:
I see that your approach is making CLX use mirrored sheets by making them on the fly if they aren't mirrored. What I'm trying to do is to make CLX use only one mirror for a top-level sheet. I'm not sure which option do we prefer.

@robert-strandh – what do you think?

Basically your approach will give us what we want at instant, but the CLX backend will use mirrored sheets for everything (unless I've misunderstood the concept), while with my WIP we'll have to fix event handling, scrolling etc etc., but we will have only one mirror.

robert-strandh:
We definitely need the kind of backend that Daniel is working on,
i.e., a CLX backend that uses a single X11 window, so that only the
top-level sheet is mirrored.  

Such a backend is important because it will be simpler.  No coordinate
swizzling is needed, for instance.  Also, it can be a model for other
backends that use display servers that do not allow nested windows.
I am also planning to use it for yet another CLX backend that uses an
X11 image as a frame buffer.  Such a backend would allow finer control
over drawing primitives, allowing things like transparency and
anti-aliasing.

However, I don't see any particular reason to remove the CLX backend
that uses mirrored sheets for all panes and gadgets.  So if at all
possible, I would like for the work that gas2serra did to serve as a
first step in creating those new backends.  This also makes it
possible for existing applications to run, using the existing
backend.

As Daniel pointed out on the #clim IRC channel on freenode, there
might be considerable code duplication between the existing CLX
backend and the one he is working on.  I am willing to temporarily
accept such duplication if that means that the two of you can continue
to work in parallel.  Later, I would like to see some refactoring of
such duplication into (say) a clx-common system.  But there is no rush
with that. 

Finally, let me just say that I am delighted to have two people
working on this code, and I would really like to see a collaboration
where appropriate.  If there is no "natural" thing to collaborate on
at the moment, at least it would be great if you could avoid Git
conflicts as much as possible.  I also don't want to micro-manage this
collaboration (I hired Daniel to work on this so that I could free up
time to work on other things), so if you could resolve differences
between the two of you as much as possible, without my involvement,
then that would be great.

gas2serra:
My idea was to leave untouched CLX backend. It is good. I'm working on CLXv2 backend [1] (it is a subclass of CLX) where we can have 1 or more mirrored sheets and an event manager that can deal with unmirrored panes.
Anyway, I will look at your branch. My email is gas2serra@gmail.com.


[1] https://github.com/gas2serra/McCLIM/tree/CLXv2

robert-strandh:
Would you consider joining us in the #clim IRC channel?  It's very low
traffic, and almost all traffic is relevant.  I fully understand if the answer
is "no", though.

ailisp:
@dkochmanski @gas2serra
Have you finished efforts on the only top-level-sheet-pane is mirrored clx backend? Looks that concrete pane classes like `climi::push-button-pane`, if want it to work correctly, depends on `clim-standard::standard-full-mirrored-sheet-mixin`. When I tried on Windows we must:
 ```lisp
(defclass push-button-pane-dummy (standard-full-mirrored-sheet-mixin push-button-pane) ())
``` 
(like what @gas2serra did but manually)
and when `(make-pane 'push-button)` must actually create `dummy-push-button-pane` but not `push-button-pane` otherwise the text displayed on push-button-pane would disappear. 
**My conclusion is as `push-button-pane` should work properly as a concrete pane, some methods for `standard-full-mirrored-sheet-mixin` may be, actually for `basic-pane`.** If so, we can just make push-button-pane when on only top-level-sheet-pane mirrored backends. Thanks for your fantastic works in 2016 that we now have a better clim. Happy new year :-)

dkochmanski:
thanks, happy new year to you too! :) For not mirrored CLX backend see:
https://github.com/robert-strandh/McCLIM/wiki/CLX-backend(s)

dkochmanski:
as a side note, I think we should slowly start to migrate to the unmirrored version as a default.

gas2serra:
Hi @ailisp,
you can create your mirrored classes manually.

`(defclass push-button-pane-dummy (standard-full-mirrored-sheet-mixin push-button-pane) ())`

or better:

`(defclass window-push-button-pane (standard-single-mirrored-sheet-mixin push-button-pane) ())`
or 
`(defclass window::push-button-pane (standard-single-mirrored-sheet-mixin push-button-pane) ())` 

I think that 'single-mirrored' is better and simple to implement. An example of simple mirrored backend is CLX-v3.

Happy new year :-)
Alessandro

PR #73, opened 2016-08-20
fix fontconfig on Debian

ailisp:
This code is provided by @dkochmanski .Thanks for his help. After this fix, McCLIM demos can be run on Debian after use `CONTINUE` restart from a warning:
```
NOTE:
* McCLIM was unable to configure itself automatically using
  fontconfig. Therefore you must configure it manually.
```


dkochmanski:
Fantastic! Thanks

PR #76, opened 2016-08-22
Implement MEDIUM-DRAW-TEXT* when medium is an instance of STRING-STREAM

lokedhs:
The implementation of `CLIM-INTERNALS:PRESENT-TO_STRING` calls `PRESENT` on a `STRING-STREAM`. Graphics functions that outputs text needs to have methods that handle this.

This fix implements such method for `MEDIUM-DRAW-TEXT*`, but these kinds of methods are probably needed for all the other gaphics functions.

### DIFF
--- a/Core/clim-basic/graphics.lisp
+++ b/Core/clim-basic/graphics.lisp
@@ -888,6 +888,7 @@ position for the character."
                               start end
                               align-x align-y
                               toward-x toward-y transform-glyphs)
+  (declare (ignore x y start end align-x align-y toward-x toward-y transform-glyphs))
   (write-string string medium))
 
 ;;; Some image junk...
--- a/Core/clim-basic/graphics.lisp
+++ b/Core/clim-basic/graphics.lisp
@@ -884,6 +884,12 @@ position for the character."
 			       align-x align-y
 			       toward-x toward-y transform-glyphs))
 
+(defmethod medium-draw-text* ((medium string-stream) string x y
+                              start end
+                              align-x align-y
+                              toward-x toward-y transform-glyphs)
+  (write-string string medium))
+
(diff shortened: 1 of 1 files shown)


dkochmanski:
```
13:39 < jackdaniel> loke__: I think it's not the right thing to do (to write-string that way), I'm not sure what is the right thing though, maybe providing separate display
                    function for the second pane?
13:40 < jackdaniel> beach: any opinions regarding https://github.com/robert-strandh/McCLIM/pull/76 ?
14:33 < jackdaniel> OK, I'm afk for now (I'll lurk at IRC, but probably w/o access to keyboard) – see you o)
15:37 < beach> jackdaniel: I don't think draw-text* is the right thing to do on a stream.
15:38 < beach> But I may be wrong.
15:42 < beach> Oh, I think I know....
15:42 < beach> loke should probably specialize on the VIEW parameter so that his PRESENT method is applicable only for the graphic view and then have a different one for a textual
               view.
15:44 < beach> ... and he should create a view class for his default stream.
15:45 < beach> Then he can have two different PRESENT methods.  The one he has, specialized to his own stream view, and a different one, specialized to textual-view.
```

dkochmanski:
```
15:45 < beach> Then he can have two different PRESENT methods.  The one he has, specialized to his own stream view, and a different one, specialized to textual-view.
16:10 < beach> loke: http://paste.lisp.org/+6XPM/1
16:10 < beach> loke: I created a potato-view and specialized on it.  I also added a PRESENT method specialized to textual-view.
16:11 < beach> loke: You can now use DRAW-TEXT* on the graphic stream and FORMAT on the textual stream.
16:13 < beach> Views are under-used and under-appreciated.  I use them all the time myself.
```

PR #89, opened 2016-09-18
Improve autoconfiguration of TTF fonts

deterministic-arts:
Out of the box, the current McCLIM seems to have some problems getting the truetype font configuration right, at least on Mac OS/X (10.11.6, El Capitan). This PR tries to improve on the situation slightly.

1. In `clim-clx::text-style-to-X-font`, the code examines the resolved pathname; if it is already absolute, it doesn't even try to `merge-pathname` it against `*truetype-font-path*`.
2. In `mcclim-truetype::find-fontconfig-font`, we use `uiop:run-program` now instead of the deprecated `asdf:run-shell-command`. Also, we try to use a fully qualified path for the `fc-match` program. I am not sure why that would be, but neither `run-shell-program` nor `run-program` (regardless of whether `:force-shell` is true) seems to be able to run `fc-match` otherwise.

I am not entirely sure, whether all of these changes are appropriate for all machines; in particular, I do not currently have access to Linux desktop machine, so I cannot test these changes on configurations other than my Mac right now. Feel free to request changes, if this PR breaks other environments.

### DIFF
--- a/Extensions/fonts/fontconfig.lisp
+++ b/Extensions/fonts/fontconfig.lisp
@@ -58,13 +58,25 @@
 * McCLIM was unable to configure itself automatically using
   fontconfig. Therefore you must configure it manually.~%"))
 
+(defparameter *fc-match-program*
+  (find-if #'probe-file
+	   '("/usr/bin/fc-match"
+	     "/usr/local/bin/fc-match"
+	     "/opt/X11/bin/fc-match")))
+
 (defun find-fontconfig-font (font-fc-name)
-  (with-input-from-string
-      (s (with-output-to-string (asdf::*verbose-out*)
-	   (let ((code (asdf:run-shell-command "fc-match -v \"~A\"" font-fc-name)))
-	     (unless (zerop code)
-	       (warn "~&fc-match failed with code ~D.~%" code)))))
-    (parse-fontconfig-output s)))
+  (and *fc-match-program*
+       (multiple-value-bind (output errors code)
+	   (uiop:run-program (list *fc-match-program* "-v" font-fc-name)
+			     :input nil :output :string :error-output nil
+			     :force-shell nil :ignore-error-status t)
+	 (declare (ignore errors))
+	 (if (not (zerop code))
+	     (progn
+	       (warn "~&fc-match failed with code ~D.~%" code)
+	       nil)
+	     (with-input-from-string (stream output)
+	       (parse-fontconfig-output stream))))))
 
 (defun fontconfig-name (family face) 
   (format nil "~A:~A" family face))
--- a/Extensions/fonts/fontconfig.lisp
+++ b/Extensions/fonts/fontconfig.lisp
@@ -58,25 +58,16 @@
 * McCLIM was unable to configure itself automatically using
   fontconfig. Therefore you must configure it manually.~%"))
 
-(defparameter *fc-match-program*
-  (find-if #'probe-file
-	   '("/usr/bin/fc-match"
-	     "/usr/local/bin/fc-match"
-	     "/opt/X11/bin/fc-match")))
-
 (defun find-fontconfig-font (font-fc-name)
-  (and *fc-match-program*
-       (multiple-value-bind (output errors code)
-	   (uiop:run-program (list *fc-match-program* "-v" font-fc-name)
-			     :input nil :output :string :error-output nil
-			     :force-shell nil :ignore-error-status t)
-	 (declare (ignore errors))
-	 (if (not (zerop code))
-	     (progn
-	       (warn "~&fc-match failed with code ~D.~%" code)
-	       nil)
-	     (with-input-from-string (stream output)
-	       (parse-fontconfig-output stream))))))
+  (multiple-value-bind (output errors code)
+      (uiop:run-program (list "fc-match" "-v" font-fc-name)
+			:input nil :output :string :error-output nil
+			:force-shell t :ignore-error-status t)
+    (declare (ignore errors))
+    (if (not (zerop code))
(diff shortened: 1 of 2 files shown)


deterministic-arts:
Hm. Should have scrolled though the issues list first. This could be related to [issue 75](https://github.com/robert-strandh/McCLIM/issues/75)

dkochmanski:
Hey, thank you for working on this. fwiw I'm currently refactoring fonts architecture (https://github.com/robert-strandh/McCLIM/tree/font-autoconfigure) – regarding absolute/relatvie pathnames, I've fixed it like this: https://github.com/robert-strandh/McCLIM/commit/02f777c0b62d84f13b70437bd3c9dd9a37b201d9 .

I'm not sure, if we really want to call `fc-match` from absolute paths, because it will break if it's installed on some non-standard directory (for instance `/usr/local/bin/fc-match`) which is in PATH. Is your `fc-match` in directory covered by `PATH`? you can verify that with `which fc-match` from the shell.

Extensions/fonts/fontconfig.lisp
```diff
@@ -58,13 +58,25 @@
...
+  (and *fc-match-program*
+       (multiple-value-bind (output errors code)
+	   (uiop:run-program (list *fc-match-program* "-v" font-fc-name)
+			     :input nil :output :string :error-output nil
+			     :force-shell nil :ignore-error-status t)
+	 (declare (ignore errors))
+	 (if (not (zerop code))
+	     (progn
+	       (warn "~&fc-match failed with code ~D.~%" code)
+	       nil)
```
dkochmanski: warn returns `NIL` anyway, we may get rid of `progn`

deterministic-arts: Ah! I wasn't sure, and was too lazy to look it up. Thanks for catching this.

deterministic-arts:
> Is your `fc-match` in directory covered by `PATH`?

That's the strange thing, it actually is (it's in `/usr/local/bin/fc-match`, and my path contains `/usr/local/bin`). Further more, when I start it on the command line, things work just fine. But for a reason I couldn't discern (and I think would require a deep dive into the `uiop` sources), neither the deprecated `asdf:run-shell-command` nor the `uiop:run-program` with (or without) `:force-shell` seem to be able to run it successfully. The result is always an error code of 127.

dkochmanski:
I think this should be fixed upstream (i.e in `sbcl`).

@pipping is one of the `ASDF` hackers and may be interested in this issue, since he works on `run-program`.

It seems, that this fails on my host too from `sbcl`, but `ecl` works fine.

deterministic-arts:
> It seems, that this fails on my host too from sbcl, but ecl works fine.

And so does CCL, so yeah, this looks like an SBCL issue.

deterministic-arts:
Ugh -- wait a minute. I actually loaded my patched McCLIM into CCL. So now I know, I didn't break something (which is good...), but I didn't really test, what I wanted to test. Just a moment.

dkochmanski:
Regarding this pull request, I would want you to squash it into two commits and rebase it against https://github.com/robert-strandh/McCLIM/tree/font-autoconfigure (and make a PR against that branch):

- replacement of run-shell-command to run-program (without probing for absolute pathnames)
- adding new `ttf` directory to the search list
- drop the fix for relatvie/absolute pathnames, because mentioned branch contains such fix already

Can you do that?

deterministic-arts:
Sure, I can do that. But despite what I just said, CCL seems to have a problem here, too. So, it's not just SBCL, which fails to autoconfigure the fonts.

dkochmanski:
Thanks. I've checked with CCL and it seems to find `fc-match` here (SBCL can't do that"):
`(ccl:run-program "fc-match" '("-v") :output *standard-output*)` returns good results

deterministic-arts:
Hu. Yeah.

dkochmanski:
Regarding `sbcl`:
```
17:55 < jackdaniel> hey! is it me, or SBCL's sb-ext:run-program can't run "fc-match" even if it's in the PATH?
                    (giving absolute pathname works fine)
17:56 < jackdaniel> (sb-ext:run-program "fc-match" '("-v")) ; -> Couldn't execute fc-match: no such file or
                    directory
17:56 < stassats> you've never used sb-ext:run-program before?
17:56 < jackdaniel> no, not really
17:57 < stassats> well, it needs :search
17:58 < jackdaniel> ah, OK, when I give the pathname to :search it seems to work, thanks. So basically I should
                    get environment variable PATH, splice it and feed :search with it?
17:59 < stassats> huh?
17:59 < jackdaniel> if I want to run applications found in paths put in PATH environment variable
18:00 < jackdaniel> ah, search is a boolean, sorry! and thanks (I've called it like (sb-ext:run-program
                    "fc-match" '("-v") :search "/usr/bin/")
18:01 < jackdaniel> and assumed you put there paths, nvm. Thank you
```

dkochmanski:
@pipping : you may be interested in `:search t` keyword for sbcl's run-program:
http://sbcl.org/manual/index.html#Running-external-programs

deterministic-arts:
I think, things could even be embarrassingly more simple here... I am used to do all development from within emacs with slime. Just now, I discovered, that the `PATH` seen by subprocesses of emacs is different from the one seen in the shell, where I add `/usr/local/bin` in `.bash_profile`, at least, when I start Emacs via click in the Dock. 

Sorry for the noise. I think, this "error" was entirely my own broken set-up here. Nonetheless, I will make the other changes as requested.

dkochmanski:
Not a problem, mistakes happen. If you could close this PR and make rebased one against the branch I've talked about

PR #91, opened 2016-09-18
fix undo

gas2serra:
Fix issues 83 about undo [1]. I'm using sbcl version 1.3.9.
Alessandro

[1] https://github.com/robert-strandh/McCLIM/issues/83


dkochmanski:
I can confirm this works and doesn't break examples nor sample applications, thanks!

Issue #94, opened 2016-09-21
End of output not displayed


robert-strandh:
In the following example, when the command "fill" has been executed,
the last two lines of the display are not shown.  Instead a black area
is shown:
```common-lisp
(cl:in-package #:common-lisp-user)

(defpackage #:pane-size
  (:use :clim-lisp :clim)
  (:export #:pane-size))

(in-package #:pane-size)

(define-application-frame pane-size ()
  ()
  (:panes
   (text :application
	 :scroll-bars nil
	 :width 800
	 :display-time nil)
   (inter :interactor :width 800 :height 100))
  (:layouts
   (:default (vertically ()
	       (scrolling (:scroll-bars t) text)
	       inter))))
			   
(defun pane-size ()
  (run-frame-top-level (make-application-frame 'pane-size)))

(define-pane-size-command (com-quit :name t) ()
  (frame-exit *application-frame*))

(define-pane-size-command (com-fill :name t) ()
  (let* ((out (find-pane-named *application-frame* 'text))
	 (medium (sheet-medium out))	      
	 (text-style (make-text-style :sans-serif :roman nil)))
    (window-clear out)
    (with-text-style (medium text-style)
      (with-text-size (medium :huge)
	(loop for i from 1 to 20
	      do (format out "~d~%" i))))))
```

dkochmanski:
confirmed

robert-strandh:
In the example, REPLAY-OUTPUT-RECORD is never called on the displayed
output record containing "20".

REPLAY-OUTPUT-RECORD is called on the the displayed output record
containing "19", but with a restrictive clipping region.  The Y
coordinate of that output record starts at 4311/8 (around 539) and the
clipping region has its maximum Y coordinate at 544.  This is why only
a few pixels are shown of this output record.

Also, when REPLAY-OUTPUT-RECORD is called on the output history itself
(which is an instance of the class STANDARD-TREE-OUTPUT-HISTORY), the
clipping region also has a maximum Y coordinate at 544, so this
clipping region is going to restrict the replay of the children, which
is why the result is what it is.

I suspect that there is nothing wrong with the code for the output
history and instead, it has to do with whatever function calls
REPLAY-OUTPUT-RECORD on the output history.  The clipping region that
is passed is too small.

robert-strandh:
Further investigation reveals that REPLAY-OUTPUT-RECORD is called from
REPLAY which is called from STREAM-REPLAY and STREAM-REPLAY is already
[...]

Issue #66, opened 2016-08-16
pane is resized too much on output


dkochmanski:
Steps to reproduce:
type `(climi::with-text-size (t 714) (princ "?"))` in the Listener – that will draw a big letter and resize the pane to fit. But it resizes too much (twice as much – the prompt should be at the very bottom of the pane, while it's in the middle).

gas2serra:
Hi,
the problem is in  "seos-write-newline".
I'm going to send you a patch.

Alessandro

PR #100, opened 2016-10-21
fix issues

gas2serra:
Hi Daniel,
I hope I had fixed two issues: #66, #94. I did not tested them deeply.
Alessandro


dkochmanski:
Thanks, I've just confirmed that it solves the mentioned issues and run against all tests in clim-examples. It seems there are no regressions.

dkochmanski:
I'll probably merge font-autoconfigure branch to master tomorrow as well

PR #103, opened 2016-10-27
Unified code style by making long argument fields uniformly on their …

otjura:
…own line. Stripped trailing spaces. Untabified everything. Abolished most over-80-char lines.


dkochmanski:
Thanks!

PR #105, opened 2016-10-28
make mcclim-fonts system independent from clx backend

gas2serra:
Hi Daniel,
is it better?


dkochmanski:
yes, thanks! is this change for your `png` branch? What will be `png` backend – something like `postscript` one?

gas2serra:
Yes, it is for a png backend (like postscript). My goal is also to build a render that is able to output to an image. I hope it will be a little step towards a full framebuffer backend.
At the moment I'm using vecto.

Issue #108, opened 2016-11-05
bug: region-contains-position-p ((self standard-rectangle-set)


gas2serra:
```
(clim:REGION-CONTAINS-POSITION-P
	  (clim:REGION-UNION 
                             (clim:MAKE-RECTANGLE* 0 0 100 100)
			     (clim:MAKE-RECTANGLE* 10 10 110 110))
	  105 105)
-->
nil
```

I think the problem is in the following function:

```
(defmethod region-contains-position-p ((self standard-rectangle-set) x y)
  (block nil
    (map-over-bands (lambda (y1 y2 isum)
                      (when (<= y1 y y2)
                        (when (isum-member x isum)
                          (return t)))
                      (when (> y y2)
                        (return nil)))
                    (standard-rectangle-set-bands self))
    nil))
```
My hypo is that we need to change ` (when (> y y2)` with ` (when (< y y2)`.
I'm not sure because I don't know what "standard-rectangle-set" is. 

Alessandro

dkochmanski:
It is a subclass of region-set (apparently mcclim-specific), I think the problem is that second return jumps of the function instead of continuing mapping over band (but skipping the current rectangle)

dkochmanski:
Fixed in master branch

PR #109, opened 2016-11-07
fix region-contains-position-p

gas2serra:


dkochmanski:
Related to #108.

(see my comment in the issue)

PR #110, opened 2016-11-07
Raster Image Backend

gas2serra:
I improved the performance. 
There is a bug with the size of the font. Maybe you can solve it.

Alessandro


dkochmanski:
Thanks!

PR #112, opened 2016-11-10
fix region-contains-position-p

gas2serra:
Hi Daniel,
the solution that you proposed is fine.
I have seen that the bands are always ordered so we can quit if we y is smaller than y2.

The example "10) Clipping Region" of drawing tests uses this method.

Alessandro


dkochmanski:
Thanks!

PR #120, opened 2016-12-06
Update medium.lisp

ailisp:
This fixme is not very accurate. As on Windows, there is an api for average character width for given text-style, not necessarily the width of `#\m`.


dkochmanski:
Thanks

Issue #35, opened 2016-06-10
clx: input: english layout


dkochmanski:
McCLIM input (interactor, listener) are set to the english layout (even if other layout is set on the X window system), so the user can't insert characters like "łęść". If truetype is loaded these characters are rendered without a problem, just can't insert them.

dkochmanski:
This bug has assigned a 100$ bounty: https://www.bountysource.com/teams/mcclim/bounties

ailisp:
Is this include the basic text-field and text-editor panels? After switch to a layout such as pl, characters with right alt doesn't work. I also tried es, in es layout, Spanish symbols typed out directly and with shift can work, such as  !"·$%&/()= (these are shift+1 2 ... 0), but with right alt ones doesn't work. May be a awkward solution is to add right alt+character commands in Libraries/Drei/basic-commands.lisp, as input to any text field/editor in the end call com-self-insert in line 514, basic-commands.lisp to insert the character.

dkochmanski:
it generally affects everything what takes keyboard input in McCLIM. You can render any character, but you can't type them. I suspect it's a trouble in `clx` (not a backend, but a library)

gabriel-laddel:
```
;;; To switch your keyboard layout between Russian & Eng, run the following and
;;; press ALT-SHIFT.
;;;
;;; "setxkbmap -option grp:switpch,grp:alt_shift_toggle,grp:led:scroll us,ru"
;;;
;;; Eval these expressions, switch to russian and enter it into drei - works.
;;; This is only a proof-of-concept, capital letters won't work, and there
;;; are several languages other than Russian that eg, sbcl supports.

(in-package clim-xcommon)

(loop for (a b c)
      in '((DEFINE-KEYSYM :CYRILLIC_SMALL_LETTERF_SOFT_SIGN 1759) (DEFINE-KEYSYM :CYRILLIC_SMALL_LETTER_SOFT_SIGN 1752)
  (DEFINE-KEYSYM :CYRILLIC_CAPITAL_LETTER_SOFT_SIGN 1784) (DEFINE-KEYSYM :CYRILLIC_CAPITAL_LETTER_HARD_SIGN 1791)
[...]

PR #121, opened 2016-12-08
[wip] address issue #35: have mcclim handle keymap change

fiddlerwoaroof:
Basically, I adjusted the code in port.lisp from the CLX backend to
listen for the :mapping-notify X event and, when received, call
(xlib:mapping-notify) with the appropriate arguments.

One note about this commit remains: I'm not sure what the appropriate
argument for the start parameter of mapping-notify is (mapping-notify
ignores the parameter anyways, so I don't think its value matters, except
that mapping-notify declares its type to be a card8, so I just pass 0 to it).

### DIFF
--- a/Backends/CLX/keysyms.lisp
+++ b/Backends/CLX/keysyms.lisp
@@ -125,27 +125,27 @@
 			 (clim-xcommon:keysym-to-keysym-name keysym)
 			 state)))
 
-	(let ((*standard-output* cl-user::*xxx*))
-	  (pprint-logical-block (*standard-output* (list keycode event-key state))
-	    (pprint-indent :block 4)
-	    (format t "Keycode: ~a" keycode)
-	    (pprint-newline :mandatory)
+	;; (let ((*standard-output* cl-user::*xxx*))
+	;;   (pprint-logical-block (*standard-output* (list keycode event-key state))
+	;;     (pprint-indent :block 4)
+	;;     (format t "Keycode: ~a" keycode)
+	;;     (pprint-newline :mandatory)
 
-	    (format t "event-key: ~a" event-key)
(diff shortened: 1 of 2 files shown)


fiddlerwoaroof:
Should fix #35

dkochmanski:
It doesn't seem to work for me. I've loaded `clim-examples`, run `(clim-demo:demodemo)`, "Text size" example, ran `setxkbmap pl` in a separate terminal, but I can't type characters like łęść.

fiddlerwoaroof:
I'll need to test with some more unusual characters, I think, the way I tested this was: start a clim application with test input, type the first four characters on the second row `qwerty`, change the keymap to french with `setxkbmap fr`, type the first four characters on the second row again `azerty`. In the french layout, some of the accented characters work but the combining characters seem to throw an error (in the text-size demo, but not in one of my test applications, for some reason)

dkochmanski:
I've added `[wip]` tag to the PR topic.

dkochmanski:
bump :)

dkochmanski:
@fiddlerwoaroof are you planning to finish this issue? It is a very important improvement which I'd love to see merged, but I don't want to finish it myself (it has $150 bounty assigned and you are a person, who came up with a solution, it wouldn't feel right).

MatthewRock:
Hey @fiddlerwoaroof ! Are you planning to finish this soon? I would like to fix that bug if no one else is going to, but since you provided partial solution, I don't want to "steal" it from you.

dkochmanski:
@MatthewRock : @gabriel-laddel  works on that afaik. You should ask him here if he plan to finish it:
https://github.com/robert-strandh/McCLIM/issues/35

PR #127, opened 2016-12-24
clx fb

gas2serra:
Hi Daniel,
I try to build a frame-buffered backend. It runs! It is really slow... in my i7 ;(
At the moment It is not usable and may be it will never be.
I think it will be useful sharing the code with you, may be in a new branch.
It requires to split CLX backend in different systems (in order to share the same code).
Alessandro


dkochmanski:
Hey, sorry for late answer (christmas holiday). Thanks for sharing! I'll test it after finishing the current task at hand (I'm refactoring frames in `core` atm).

even if its slow, its a very good starting point (having something what works and improving it gradually).

dkochmanski:
I have loaded it successfully, some examples doesn't work (it's too slow maybe? – for instance logic cube). Everything is a bit larger than in clx backend.

Merging your changes locally doesn't seem to break clx whatsoever – I'll do some more testing and if there is no regressions I'll merge it today. Looks very cool :-)

dkochmanski:
it seems that fb backend can't work on X fonts, so it should filter them out (see `Backends/CLX/fonts.lisp:176`) – you may verify that with `font-selector` demo

dkochmanski:
In drawing-tests which compare raster backend with the currently used one fb backend has some problems with point scaling

gas2serra:
Hi Daniel,
I did not see the problem with point transformation and scaling... thanks.
In logic cube there is a problem with flushing and syncing. I have not find a good solution yet.
I used a lot of your code for the fonts, I think I broke it somewhere.
There are also other small problems.

I agree "Everything is a bit larger than in clx backend". At the moment this is a big problem. 
Another problem is that is is too slow. 
I think the biggest problems are:
* At the moment, an image is generated in the memory and then it is sent to the X server by socket. I have seen that there is a shared memory extension (MIT-SHM) for X... that is used by gtk, qt, ecc ecc
* At the moment, the paths of a char is cached ant not the an image.
* There are a lot of useless repainting (that is true also with the others backends)
* Asynchronous repainting

dkochmanski:
regarding shm – maybe it's worth a try to use sdl framebuffer context, or to try out with wayland buffers with more lispy https://github.com/malcolmstill/ulubis ?

gas2serra:
I had a lot of problems with cl-sdl2 library. I'll look at ulbis...
Thanks,
Alessandro

PR #131, opened 2017-01-12
Devel robert

gas2serra:
Hi Daniel,
I change the rendering (from sheets to mirrors) in order to fix the with-double-buffering issue in CLX-fb. 
I had to introduce a new function in CLX: "sheet-xmirror" and replace every call to sheet-mirror with sheet-xmirror". I cannot avoid that.

Now, almost all demos and applications are working with CLX-fb. It is still slow.

I found a few problems with recording:
a) medium-draw-rectangles are not recorded (a bug)
b) medium-copy-area are not recorded (may be we  don't want record them)

Regards,
Alessandro


Backends/CLX-fb/mirror.lisp
```diff
@@ -0,0 +1,66 @@
...
+						 (xlib::put-image xmirror
+								  gcontext
+								  clx-image
+								  :src-x (round (max min-x 0)) :src-y (round (max min-y 0))
+								  :x (round (max min-x 0)) :y (round (max min-y 0))
+								  :width  (- width (min 0 (- min-x)))
+								  :height (- height (min 0 (- min-y))))))))
+				       mcclim-render::dirty-region)
+	    ;;(format *debug-io* "R: ~A ~A ~A ~A~%" (round min-x)
+	    ;;(round min-y) (round max-x) (round max-y))
```
dkochmanski: I think we should use `log4cl` with configurable log levels for such debug information (so we don't have to comment/uncomment/#+ them every time). What do you think?

gas2serra: I agree. It will be important having a standard method for logging. Usually I add and remove them every time. Sometime I forget to remove them ;( sorry.

PR #147, opened 2017-02-20
fontconfig: windows fix

fabriceleal:
Add Window's system font directory; also ensure you have dejavusans installed, for this to work correctly


dkochmanski:
Thanks!

defunkydrummer:
This thing still fails here (Windows 10, CCL64), apparently because it will still try to find the DejaVu* fonts (which don't exist by default on Windows). Thus, it will try to `build-font/family-map`, which relies on an external utility ("fc-match") that doesn't exist on Windows either.

PR #148, opened 2017-02-20
Clx backend: windows fix

fabriceleal:
clx backend works when using host "localhost" and protocol "internet"

### DIFF
--- a/Backends/CLX/port.lisp
+++ b/Backends/CLX/port.lisp
@@ -71,6 +71,9 @@
 	    :protocol protocol))))
 
 (defun helpfully-automagic-clx-server-path ()
+  #+win32
+  (parse-clx-server-path '(:clx :host "localhost" :protocol :internet))
+  #-win32
   (restart-case (automagic-clx-server-path)
     (use-localhost ()
       :report "Use local unix display"


npatrick04:
Is there a particular reason why this pull request hasn't been merged?  I've tested it on my system and was happier with it than my inappropriate hack.

dkochmanski:
That is completely my fault. I have started review, but apparently I have to also "submit" it. This GitHub functionality is new to me. I'm terribly sorry (I made this quick review same dame the PR came...)

npatrick04:
Good point about the trivial-features solution.  Are there other places in the code-base that would benefit from trivial-features?

dkochmanski:
I'm not aware of such parts of code, but if we want to feature separate code branches for different platforms, then trivial-features are way to go. If we use some implementation-specific feature names, which describe something more common between implementations, then it should be considered being a bug.

I'm semi-regularly checking whenever McCLIM works on SBCL, CCL and ECL, so even if we have such bugs they are not critical to inner working at the moment though.

npatrick04:
Sweet, now my next question is where to put the dependency.  It could go in the mcclim-clx system since that's where port.lisp is specified, or one of the top-levels in anticipation of future uses.  What's your preference?

dkochmanski:
toplevel, yes. I plan to rearrange systems a bit in near future (because we have some weird dependency graph, which is a bit ugly) – put the dependency in clim-basic system (`Core/clim-basic/clim-basic.asd`) which is a root of other modules.

npatrick04:
OK, I will try updating the pull request later unless @fabriceleal gets to it first.

fabriceleal:
Won't be able to do anything soon (sorry, i know i promised i would add trivial-features :\ )

dkochmanski:
merged in #155

Issue #21, opened 2016-03-02
bug: (setf frame-command-table)


gas2serra:
Changing the frame’s command table does NOT redisplay, and rebuild, the menu bar.

From spec:

> (setf frame-command-table) command-table frame   [Generic Function]
> Sets the command table for the frame frame to command-table. Changing the frame’s command
> table will redisplay the command menus (or menu bar) as needed.

dkochmanski:
This bug has assigned a 100$ bounty: https://www.bountysource.com/teams/mcclim/bounties

gabriel-laddel:
Wouldja stop being useless already and assign bounties to the parts of the project that are ~ difficult ~ to solve?

@robert-strandh thoughts?

nsrahmad:
## Recreating the bug

The easiest way to reproduce the bug is to use `clim-listener` (it has all we need; Menu bar and REPL). Load the clim-listener and execute following code.

1.  Copy the current active command-table and modify it.

```lisp
    (defparameter *new-commnad-table* (frame-command-table *application-frame*))

    (add-menu-item-to-command-table *new-command-table* "Dummy" :divider :label)
```
1.  Now set the frame command-table to *new-command-table*.

```lisp
    (setf (frame-command-table *application-frame*) *new-command-table*)
```

According to the spec, this last step should have updated the menu-bar with our new menu item.

## What is the problem?

To understand the problem, You need to first understand how McClim creates the menu bar. 

-   When McClim starts creating the application-frame, It first builds the menu bar and hard-wires it to the top of the frame.
-   Then it creates every other panes and adds them to the frame.
-   Even if we change the contents of menu bar, We just can’t repaint or rebuild the menu bar. Being hard-wired to the frame on creation, It cannot be updated independently.
-   Only way to solve the problem is to regenerate the whole frame, which will build the menu bar again.

Luckily, McClim provides the method (generate-panes) that does just that.

## Problem with `generate-panes`

So to solve the problem, you just need to create an `:after` method to `(setf (frame-command-table))` method, which calls the `generate-panes`. Doing that actually works but there seems to be a problem. `generate-panes` should not be clearing the whole output history in the clim-listener pane. Fortunately, Just looking at the git log, I found the commit a74e3960b9bffb9 by @dkochmanski , which fixed “generate-panes repainting problem” by creating every pane from scratch. That’s not the way to solve this problem.

## Fixing `generate-panes`

[...]

PR #153, opened 2017-04-08
Fix menu bar

nsrahmad:
I have posted the details at #21. My apologies for putting all that stuff in a single post.


Core/clim-core/frames.lisp
```diff
@@ -232,9 +232,20 @@ documentation produced by presentations.")
...
+    (let ((prev-menu (get-menu (frame-command-table frame)))
+	  (current-menu (get-menu new-command-table)))
+      (cond
+	((null current-menu) (call-next-method))
+	((null prev-menu) (call-next-method))
+	(t (let ((result (call-next-method)))
+	     (generate-panes *default-frame-manager* *application-frame*)
+	     result))))))
+
 (defmethod generate-panes :before (fm  (frame application-frame))
```
dkochmanski: Could you use `previous` (full word) for a lexical variable name? Also, this `cond` should be more readable with:
```
(prog1 (call-next-method)
  (when (or (null current-menu)
	    (null previous-menu))
    (generate-panes *default-frame-manager* *application-frame*)))
```

dkochmanski:
Except the cosmetic fix I've mentioned all is fine. Really good job (especially with problem analysis and documentation). I've tested the changes and the problem with layout change is indeed fixed and adding element to menu works as well.

nsrahmad:
1) your first proposed change is good, and I have updated it.

2) `(setf (frame-command-table))` should return the newly-set command-table, which is what the primary method returns. But when we are using an `:around` method, what this `:around` method returns will be the return value of the original call. Which means we either have to call `call-next-method` at the very end (clause 1 and 2) or call `(call-next-method)`, store it’s return value in a variable, do other operations and return the stored value (clause 3 does exactly that). We cannot call it at the end because `generate-panes` must come after `(call-next-method)`.

Your proposed change doesn't take this into account and the last call is to `generate-panes` which will return the top-level `frame-pane` as the call to `(setf (frame-command-table))`. That's what I meant (but forgot to explain) in my #21 explanation, when I called this `:around` method totally transparent.

dkochmanski:
Hey, note the prog1 in the around method. It will return desired result, so the semantics should be preserved.

On April 8, 2017 12:42:02 PM CEST, Nisar Ahmad <notifications@github.com> wrote:
>1) your first proposed change is good, and I have updated it.
>
>2) `(setf (frame-command-table))` should return the newly-set
>command-table, which is what the primary method returns. But when we
>are using an `:around` method, what this `:around` method returns will
>be the return value of the original call. Which means we either have to
>call `call-next-method` at the very end (clause 1 and 2) or call
>`(call-next-method)`, store it’s return value in a variable, do other
>operations and return the stored value (clause 3 does exactly that). We
>cannot call it at the end because `generate-panes` must come after
>`(call-next-method)`.
>
>Your proposed change doesn't take this into account and the last call
>is to `generate-panes` which will return the top-level `frame-pane` as
>the call to `(setf (frame-command-table))`. That's what I meant (but
>forgot to explain) in my #21 explanation, when I called this `:around`
>method totally transparent. 
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/robert-strandh/McCLIM/pull/153#issuecomment-292709844

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

nsrahmad:
Thanks @dkochmanski, I mistook `prog1` with simple `progn`. looks like `prog1` is exactly what we need here.

dkochmanski:
why do you still keep `cond` there then?
```
(prog1 (call-next-method)
  (when (and current-menu previous-menu)
    (generate-panes *default-frame-manager* *application-frame*)))
```
would get rid of it.

Also, as far as I see, this could be written as `after` method which whole body is
```
(defmethod (setf frame-command-table) :after (new-command-table frame)
  (flet ((get-menu (x) (slot-value x 'menu)))
    (when (and (get-menu (frame-command-table frame))
               (get-menu new-command-table))
      (generate-panes *default-frame-manager* *application-frame*))))
```
because return value of after methods are discarded (they are added just for side effects).

nsrahmad:
1) `cond` is replaced with `if`, because It remains `:around` method (see below) we had to manually call `call-next-method` regardless of condition.

2 ) It has to be `:around` method, because In `:after` method
```lisp
(get-menu (frame-command-table frame))
(get-menu new-command-table)
```
will refer to Same table!. (`:after` method will run after primary method, which happens to be a `setf`, so value of frame-command-table is updated with new-command-table already before calling `:after`.

dkochmanski:
I see now, thanks! I've increased the bounty yesterday, because you did a very fine job with the analysis and documentation too. Don't forget to claim it :)

nsrahmad:
Thanks you @dkochmanski for showing me `prog1` which I have not seen before and being so helpful.

PR #155, opened 2017-04-17
Use internet socket w/ localhost when connecting to CLX on Windows.

npatrick04:
Also added :trivial-features to clim-basic.asd.

I tested this in Windows with SBCL and CCL and XMING server 6.9.0.31 (looks fairly out-of-date).  
SBCL works fine.  

I found that CLX w/ CCL is giving an error related to the X server version being incompatible.  I'm going to attempt to find out why SBCL works, but CCL doesn't on my machine, but I think that's outside of this issue.


dkochmanski:
Looks good, thanks!

PR #168, opened 2017-04-26
Don't throw an error when trying to load images with a single channel

lokedhs:
Fix for https://github.com/robert-strandh/McCLIM/issues/167


dkochmanski:
Looks good, thanks!

PR #178, opened 2017-05-08
Bezier drawing fixes

slyrus:
* moved bezier stuff into Extensions

 * fix bezier curve drawing when replaying output records. Before,
   when we invoked the bezier drawing code in a context where
   replay-output-record would get called with non-zero
   output-record-position, we weren't adjusting the coordinates of the
   bezier curve when drawing it.

   The other drawing primitives have it a bit easier as they all just
   take simple values and the code in the particular setf
   output-record-position method was free to just hack the slots of
   the output record. Since the bezier designs are stored as objects,
   we can't/don't want to modfify those objects when replaying
   them. So... we make a new translation for the output record and add
   new slots to bezier-design-output-record
   and bezier design to store this translation. Then we apply the
   translation as necessary in medium-draw-bezier-design*.

   Question: if we have both an output-record-translation and a
   medium-transformation, which do we apply first?!?


Extensions/bezier/bezier-test-comments.lisp
```diff
@@ -0,0 +1,15 @@
...
+(defparameter *r1* (make-bezier-area* '(10 10 20 20 30 20 40 10 30 5 20 5 10 10)))
+
+(defparameter *r2* (make-bezier-area* '(15 10 20 12 30 15 35 10 30 8 20 8 15 10)))
+
+(defparameter *r3* (region-difference *r1* *r2*))
+
+(defparameter *r4* (make-bezier-curve* '(100 100 120 150 160 160 170 160)))
+
+(defparameter *r5* (convolute-regions *r2* *r4*))
+|#
```
dkochmanski: I would like to have it in `clim-examples` as a drawing test (see https://github.com/robert-strandh/McCLIM/blob/master/Examples/drawing-tests.lisp) instead of having dangling file with content which isn't loaded.

dkochmanski: Additionally to that, I'd like to see a demo https://github.com/slyrus/bezier-test/blob/master/bezier-test.lisp adjusted and merged to clim-exampls as well, so it may be selectable from `(clim-demo:demodemo)` menu. Should I fix it by myself, or you want to do it?

mcclim.asd
```diff
@@ -72,7 +72,8 @@ interface management system."
 (defsystem #:mcclim/extensions
   :depends-on (#:mcclim-bitmaps
                #:conditional-commands
-               #:mcclim-layouts/tab))
+               #:mcclim-layouts/tab
+               #:mcclim-bezier))
```
dkochmanski: I'm glad we have crafted it as an extension. Are you planning to write documentation for it? If not, I'll add issue assigned to myself and make one.

dkochmanski:
Regarding question (copied from irc)

```
Question: if we have both an output-record-translation and a
medium-transformation, which do we apply first?!?
```

```
* nyef points out that the output-record necessarily captures the medium-transformation at recording time.
```

dkochmanski:
We have also misc tests:
https://github.com/robert-strandh/McCLIM/blob/master/Examples/misc-tests.lisp if drawing isn't well suited.

slyrus:
See updated pull-request.

PR #179, opened 2017-05-08
with-drawing-options: call invoke-with-drawing-options directly

slyrus:
* before, we would call (apply #'invoke-with-drawing-options
   ... (list ,@drawing-options). Now we just
   do (invoke-with-drawing-options ... ,@drawing-options).

   This avoids a compiler warning from SBCL about the third argument
   not being a constant (and why cons up a new list and call apply
   anyway?).


dkochmanski:
looks good, thanks!

PR #180, opened 2017-05-08
Bezier extension clean 3

slyrus:
I've rewritten the history with two commits. One that just moves the bezier stuff, as is, into Extensions, and then a single commit with my fix for the output record stuff.


dkochmanski:
Merged, thanks!

PR #184, opened 2017-05-09
mcclim-render::image-sheet-image -> image-mirror-image

slyrus:
* in output-to-png.lisp


dkochmanski:
will it work with a single-mirror version of clx (clxv3)[1]? What does this change fix?

[1] https://github.com/robert-strandh/McCLIM/wiki/CLX-backend(s)

slyrus:
image-sheet-image doesn't exist anymore, AFAICT. I think this is what we want.

PR #187, opened 2017-05-10
use mcclim-render backend for drawing bezier curves

slyrus:
* add methods that specialize on the render-medium-mixin and call the
   mcclim-render backend code to do the bezier curve drawing.


dkochmanski:
works a lot faster, thanks! merging

PR #188, opened 2017-05-10
Bezier package improvements

slyrus:
moves all the bezier code into mcclim-bezier and import-from's the appropriate symbols from the other packages, instead of doing (in-package :...) to get access to symbols from null, render, and postscript backends.

Also, fix up the bezier-area drawing-test example and add a couple FIXMEs.

### DIFF
--- a/Extensions/bezier/bezier.lisp
+++ b/Extensions/bezier/bezier.lisp
@@ -843,8 +843,6 @@
 
 ;;; NULL backend support
 
-(in-package :clim-null)
-
 ;;; FIXME: need these to stop the default method attempting to do
 ;;; pixmaps, which it appears the null backend doesn't support yet.
 (defmethod mcclim-bezier:medium-draw-bezier-design*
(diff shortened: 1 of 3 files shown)


Extensions/bezier/bezier.lisp
```diff
@@ -209,6 +209,8 @@
 			(transform-segment transformation segment))
 	       (%segments path))))
 
+;;; FIXME! This overrides region-equal in Core/clim-basic/regions.lisp
+;;; -- we probably don't want that!
 (defmethod region-equal ((p1 point) (p2 point))
```
dkochmanski: I wonder, if we should use commented code from `Core/clim-basic/coordinates.lisp`, which does the same thing (that is - it works with epsilon instead of exact coordinates), what makes "default" `region-equal` behave exactly like this method. What do you think?

slyrus:
Since @robert-strandh wrote the region-equal code, it would be great if he weighed in :) My gut is that we should offer coordinate-epsilon= and friends, and then use that in range-epsilon-equal with epsilon being an optional or kw arg to each of those methods. thanks for merging!

PR #192, opened 2017-05-11
More bezier stuff

slyrus:
Added relative-to-absolute-coord-seq and some more bezier drawing tests


dkochmanski:
looks good, thanks!

PR #193, opened 2017-05-12
More bezier stuff

slyrus:
Updated the semantics of relative-to-absolute-coord-seq and added an :angle arg to draw-arrow and draw-arrow*


dkochmanski:
thanks!

PR #194, opened 2017-05-13
remove transformation and original-region slots

slyrus:
* remove slots that are either unused or poorly used. It's not at all
   clear to me that these are supposed to do anything.

@robert-strandh can you review this one? This stuff seems like cruft to me, but perhaps you're using it. thanks!

### DIFF
--- a/Extensions/bezier/bezier.lisp
+++ b/Extensions/bezier/bezier.lisp
@@ -79,11 +79,7 @@
   (- (* (realpart z) (point-y v))
      (* (imagpart z) (point-x v))))
 
-(defclass bezier-design (design)
-  ((%or :accessor original-region :initform nil)
-   (%trans :initarg :transformation
-	   :reader transformation
-	   :initform +identity-transformation+)))
+(defclass bezier-design (design) ())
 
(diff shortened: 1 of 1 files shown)


Extensions/bezier/bezier.lisp
```diff
@@ -287,53 +283,29 @@ second curve point, yielding (200 50)."
...
+  (make-instance (class-of design)
+                 :segments (mapcar (lambda (s)
+                                     (transform-segment transformation s))
+                                   (segments design))))
 
 (defmethod compute-bounding-rectangle* ((design bezier-design))
-  (multiple-value-bind (lx ly ux uy) (call-next-method)
-    (let ((tr (transformation design)))
-      (transform-rectangle* tr lx ly ux uy))))
+  (call-next-method))
```
dkochmanski: do we need to specialize method, if all what we do is call-next-method?

slyrus: good point. I think I just had this in there because I didn't want to go through the hassle of trying to delete that specific method,

dkochmanski:
merged, thanks!

PR #196, opened 2017-05-15
Fix warnings

slyrus:
Lots of warnings fixed. Please go through one by one and check.


Examples/gadget-test.lisp
```diff
@@ -132,6 +132,7 @@
                  :orientation :vertical
                  :current-color +black+
                  :value 0)
+     #+nil
      (slider-v1  :slider
```
dkochmanski: please use `#+(or)`

Examples/gadget-test.lisp
```diff
@@ -151,6 +152,7 @@
                  :orientation :vertical
                  :current-color +black+
                  :value 0)
+     #+nil
      (radar      (make-pane 'radar-pane :name 'radar))
```
dkochmanski: please use `#+(or)`

Examples/drawing-tests.lisp
```diff
@@ -475,7 +481,8 @@
 	    (with-drawing-options (stream :line-thickness lt :line-joint-shape lj :line-dashes ld)
 	      (draw-text* stream (format nil "~A ~A ~A" ld lj lt) 20 y)
 	      (let ((x-positions '(150 300)))
-		  (draw-rectangle* stream 200 (+ 10 y) 450 (+ 35 y)
+                (declare (ignore x-positions))
+                (draw-rectangle* stream 200 (+ 10 y) 450 (+ 35 y)
 				   :filled nil))
```
dkochmanski: if x-position is ignored, we may get rid of a whole let and move body one level up, right?

Examples/drawing-tests.lisp
```diff
@@ -536,6 +543,7 @@
 	  (with-drawing-options (stream :line-thickness lt :line-dashes ld)
 	    (draw-text* stream (format nil "~A ~A" ld lt) 20 y)
 	    (let ((x-positions '(150 300)))
+              (declare (ignore x-positions))
```
dkochmanski: remove let block if all vars unused

Examples/drawing-tests.lisp
```diff
@@ -602,6 +610,7 @@
...
 	      (draw-circle* stream 200 (+ 10 y) 30
```
dkochmanski: remove let block if all vars unused

Examples/draggable-graph.lisp
```diff
@@ -114,10 +115,8 @@
 (define-presentation-to-command-translator record-dragging-translator
     (t com-drag-node draggable-graph-demo
        :tester ((object presentation)
-                (declare (ignore object))
                 (find-graph-node presentation)))
```
dkochmanski: why is this a bogus declaration? (same question goes for the one below)

Core/clim-core/presentations.lisp
```diff
@@ -1310,6 +1310,7 @@ and used to ensure that presentation-translators-caches are up to date.")
               ,@body)))
         (t
          `(lambda (,(car args) &key ,@(cdr args) &allow-other-keys)
+            (declare (ignorable ,(car args)))
             ,@body))))
```
dkochmanski: whet about the key arguments? Is it possible that they are not referenced in body? Then it would be
`(declare (ignorable ,@args))`

Core/clim-basic/regions.lisp
```diff
@@ -90,8 +90,12 @@
...
 ;;; coordinate is defined in coordinates.lisp
 
-(defvar +everywhere+ (make-instance 'everywhere-region))
-(defvar +nowhere+ (make-instance 'nowhere-region))
+(defconstant +everywhere+ (make-instance 'everywhere-region))
+(defconstant +nowhere+ (make-instance 'nowhere-region))
+
+(setf everywhere* +everywhere+)
+(setf nowhere* +nowhere+)
+
```
dkochmanski: I don't get this change. Why can't we use `+everywhere+` and `+nowhere+` from the very beginning?

NB `defconstant` in sbcl is patched anyway to `defvar` for sake of getting rid of problems when we recompile things.

Backends/RasterImage/output-to-png.lisp
```diff
@@ -33,14 +33,15 @@
...
+		(save-mirror-image-to-file (sheet-mirror sheet) ,file ,(extract-format file)))
 	      (,enter-fn (sheet stream)
 		(declare (ignore sheet stream))
 		nil))
 	 (declare (dynamic-extent #',cont))
 	 (invoke-with-output-to-raster-image #',cont #',enter-fn #',exit-fn
 					     :opticl-image
-					     (extract-format ,file)
+					     ,(extract-format file)
 					     ,@options)))))
```
dkochmanski: @gas2serra could you look into it?

Examples/draggable-graph.lisp
slyrus: So the convoluted macro-hackery behing define-presentation-to-command-translator creates a bunch of nested forms before including the body. The upshot is that we end up trying to ignore a variable that was declared in an outer let block. I fixed this by changing the macro such that its kw args to tester are ignorable. And, in so doing, I've removed these declarations that were, justifiably, triggering warnings.

Core/clim-core/presentations.lisp
slyrus: maybe, but if somebody's going through the trouble of specifying the keyword args, they probably should use them. I'd leave this as is for now.

dkochmanski:
Looks good, thanks!

PR #197, opened 2017-05-15
More bezier stuff

slyrus:
With these changes we can no take point-seq's and coord-seq's instead of just lists of points and coords. Furthermore, segments can, internally, be either a list or a vector (or other sequence, presumably).

### DIFF
--- a/Extensions/bezier/bezier.lisp
+++ b/Extensions/bezier/bezier.lisp
@@ -147,16 +147,21 @@
 (defgeneric compute-bounding-rectangle* (object))
 
 (defmethod compute-bounding-rectangle* ((segments-mixin segments-mixin))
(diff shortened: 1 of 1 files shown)


Extensions/bezier/bezier.lisp
```diff
@@ -79,6 +79,12 @@
   (- (* (realpart z) (point-y v))
      (* (imagpart z) (point-x v))))
 
+(defun ensure-list (sequence)
+  (if (listp sequence)
+      sequence
+      (map 'list #'identity sequence)))
+
```
dkochmanski: we already pulled dependency on alexandria, please use that

dkochmanski: alexandria has ensure-list

PR #200, opened 2017-05-15
Fix warnings

slyrus:


dkochmanski:
Looks good, no remarks from my side. Thanks!

PR #210, opened 2017-05-21
tab-layout: draw-line* takes x1 y1 x2 y2 parameters, fix alloc-space

dkochmanski:
Previously we have taken pane width as x2. Since we count from 0,
(width,17) was one more than last pixel on the pane what caused
alloc-space to make the pane bigger. i.e pane width is 5, then last
point is (4,y).

cosmetic: use accessor instead of raw slot value.

### DIFF
--- a/Examples/font-selector.lisp
+++ b/Examples/font-selector.lisp
@@ -160,9 +160,9 @@
              (y2 (+ y1 height))
              (ybase (+ y1 baseline)))
         (when ok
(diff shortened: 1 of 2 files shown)


dkochmanski:
I've added a few more fix with add page logic.

slyrus:
Looks good to me. Do you want all of these individual commits in the history, or do you want to squash some and rebase for a more abbreviated history? I'm OK either way.

dkochmanski:
I'd like to keep individual commits if that's OK, thanks!

PR #211, opened 2017-05-21
Bezier render 2

slyrus:
This branch uses the mcclim-raster-image routines to draw bezier shapes on the CLX backend. Still not the ideal approach, but a temporary stopgap that makes rendering less bad.

### DIFF
--- a/Core/clim-basic/graphics.lisp
+++ b/Core/clim-basic/graphics.lisp
@@ -895,7 +895,9 @@ position for the character."
       (when medium
 	(medium-free-image-design medium design))
       (setf medium current-medium)
-      (setf medium-data nil))))
+      #+nil
+      (when medium-data
+        (setf (car medium-data) nil)))))
 
 (defmethod medium-draw-image-design*
     ((medium sheet-with-medium-mixin) design x y)
--- a/Extensions/bezier/bezier.lisp
+++ b/Extensions/bezier/bezier.lisp
@@ -83,7 +83,7 @@
 
 (defclass bezier-design (design) ())
 
-(defgeneric medium-draw-bezier-design* (stream design))
+(defgeneric medium-draw-bezier-design* (stream design options))
 
 
 ;;; recording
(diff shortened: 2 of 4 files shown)


Core/clim-basic/graphics.lisp
```diff
@@ -895,7 +895,9 @@ position for the character."
       (when medium
 	(medium-free-image-design medium design))
       (setf medium current-medium)
-      (setf medium-data nil))))
+      #+nil
+      (when medium-data
+        (setf (car medium-data) nil)))))
```
dkochmanski: why do we comment out this block? please use `#+(or)` here.

Extensions/bezier/bezier.lisp
```diff
@@ -844,25 +844,110 @@ second curve point, yielding (200 50)."
...
 
+;;; The real CLX method
+
+(defun my-compute-rgb-image-pixmap (drawable image)
+  (let* ((width (climi::image-width image))
+         (height (climi::image-height image))
+         (depth (xlib:drawable-depth drawable))
+         (im (clim-clx::image-to-ximage-for-drawable drawable image)))
+    (setf width (max width 1))
+    (setf height (max height 1))
```
dkochmanski: if image width/height is lower than 1 (i.e 0) wouldn't it make sense to just skip whoe computation and skip putting pixmap (and invoking whole xlib machinery here)?

Extensions/bezier/bezier.lisp
```diff
@@ -844,25 +844,110 @@ second curve point, yielding (200 50)."
...
+         (im (xlib:create-image :width width
+                                :height height
+                                :depth 1
+                                :data xdata)) )
+    (dotimes (y width)
+      (dotimes (x height)
+        (if (> (aref idata x y) #x80000000)
+            (setf (aref xdata x y) 0)
+	    (setf (aref xdata x y) 1))
+        ;; (print (aref idata x y) *terminal-io*)
```
dkochmanski: it may be github rendering code improperly, but this `if` clauses - are they indented properly?

Extensions/bezier/bezier.lisp
```diff
@@ -844,25 +844,110 @@ second curve point, yielding (200 50)."
...
+        (when (= (aref idata x y) #x00FFFFFF)
+          (setf (aref xdata x y) 0))))
+    (unless (or (>= width 2048) (>= height 2048)) ;### CLX breaks here
```
dkochmanski: interesting. afair it butchered too big images into scanlines and send them one-by-one (put-image that is), so it means that such big image scanline exceeds one request limit.

PR #215, opened 2017-05-31
dispatch-repaint directly on the mirrored-sheets

gas2serra:
A small fix that improves the performance of single-mirrored backends.


dkochmanski:
Thanks!

Issue #172, opened 2017-04-26
Listener fails to display prompt for non-CLIM interaction code at bottom of window


abridgewater:
When using Y-OR-N-P (or, it seems, any other mechanism that prints a prompt, does a FINISH-OUTPUT, and then reads input) at the bottom of the listener window, the prompt does not display until there is subsequent output.
![y-or-n-p at bottom of window](https://cloud.githubusercontent.com/assets/469380/25452346/51f56700-2a93-11e7-9a95-130d3eeb5878.png)

dkochmanski:
fix in master confirmed. closing.

Issue #174, opened 2017-04-26
Listener "Clear Output History" command fails to reset scrollbars


abridgewater:
This is probably a matter of not having reset the viewport size when clearing the history.
![listener with cleared output](https://cloud.githubusercontent.com/assets/469380/25452490/bb8bb6f6-2a93-11e7-97a6-24a82a7aa33b.png)

gas2serra:
Hi,
do you use the command menu "Listener>Clear Output History" or do you call it in the prompt ",Clear Output History"?

abridgewater:
I had been using the command menu, but it turns out that the effect occurs when being invoked via the prompt as well.

gas2serra:
The problem is in the function 'fit-pane-to-output'. It never decreases the min-height.

` :min-height (max fit-height min-height)`

@dkochmanski Is it a clim standard function? Do you know why we need it?


```
(defmethod fit-pane-to-output ((stream clim-stream-pane))
  ;; Guard against infinite recursion of size is set to :compute, as
  ;; this could get called from the display function. We'll call
  ;; compose-space here, which will invoke the display function
  ;; again..
  (when (and (sheet-mirror stream)
                     (not (or (eq (pane-user-width stream) :compute)
                                  (eq (pane-user-height stream) :compute))))
    (let* ((output (stream-output-history stream))
           (fit-width  (bounding-rectangle-max-x  output))
           (fit-height (bounding-rectangle-max-y output)))
      (multiple-value-bind (width min-width max-width 
                            height min-height max-height)
          (space-requirement-components (compose-space stream))
	(format *debug-io* "##==> ~A ~A ~A ~%" fit-width fit-height (compose-space stream))
      (change-space-requirements stream
                                 :min-width (max fit-width min-width)
                                 :min-height (max fit-height min-height)
                                 :width (max fit-width width)
                                 :height (max fit-height height)
                                 :max-width max-width
                                 :max-height max-height)))))
```

gas2serra:
@dkochmanski  What does ':compute' mean?

dkochmanski:
it's a size-spec:
```
CLIM application frames accept the \cl{:foreground}, \cl{:background},
\cl{:text-style}, and layout pane options.  The space requirement options
(\cl{:width}, \cl{:height}, and so forth) can also take a size specification of
\cl{:compute}, which causes CLIM to run the display function for the pane, and
make the pane large enough to hold the output of the display function.
```

dkochmanski:
[...]

PR #219, opened 2017-06-03
using change-space-requirements instead of fit-pane-to-output

gas2serra:
#174 
I think that it is a bad idea changing the user space preferences.


gas2serra:
It also partially fix #172.
The prompt is displayed. No scrolling.

dkochmanski:
As far as I can tell, we don't use `fit-pane-to-output` anywhere anymore. It's not part of the spec either, so I believe we should remove this generic function whatsoever if its incorrect.

Issue #171, opened 2017-04-26
Listener presentation "Menu" windows are created at the wrong size


abridgewater:
When opening a context menu on a presentation in the Listener, the window for the menu is created overlarge and then sized down to the required space for the menu, causing a noticeable flicker.

gas2serra:
The problem is in '(defmethod adopt-frame ((fm frame-manager) (frame menu-frame))'. 
The top-sheet is adopted by the graft before calling allocate-space.

abridgewater:
That doesn't quite make sense. The critical bit would be to set the sheet region based on the result of COMPOSE-SPACE before enabling it, which can happen post-graft... Though now I'm having trouble remembering my substructure redirection semantics, so there may be further subtleties here.

gas2serra:
"The critical bit would be to set the sheet region based on the result of COMPOSE-SPACE before enabling it".
I agree. The following method seem to do better. The mirror is created with the wrong size. The size is updated. And, only in the end, the mirror is shown (sheet-enabled-p).
I have a fast PC and I cannot see any difference. Can you test it?

```
(defmethod adopt-frame ((fm frame-manager) (frame menu-frame))
  (setf (slot-value fm 'frames) (cons frame (slot-value fm 'frames)))
  (setf (frame-manager frame) fm)
  (let* ((t-l-s (make-pane-1 fm *application-frame*
                             'menu-unmanaged-top-level-sheet-pane
			     :name 'top-level-sheet
			     :enabled-p nil)))
    (setf (slot-value frame 'top-level-sheet) t-l-s)
    (sheet-adopt-child t-l-s (frame-panes frame))
    (let ((graft (find-graft :port (port fm))))
      (sheet-adopt-child graft t-l-s)
      (setf (graft frame) graft))
    (let ((pre-space (compose-space t-l-s))
          (frame-min-width (slot-value frame 'min-width)))
      (multiple-value-bind (width min-width max-width height min-height max-height)
          (space-requirement-components pre-space)
        (flet ((foomax (x y) (max (or x 1) (or y 1))))
          (let ((space (make-space-requirement :min-width  (foomax frame-min-width min-width)
                                               :width      (foomax frame-min-width width)
                                               :max-width  (foomax frame-min-width max-width)
                                               :min-height min-height
                                               :height     height
                                               :max-height max-height)))
            (allocate-space (frame-panes frame)
                            (space-requirement-width space)
                            (space-requirement-height space))
[...]

PR #220, opened 2017-06-03
unmanaged-top-level-sheet-pane requires to be enabled

gas2serra:
It solves #171.
unmanaged-top-level-sheet-panes are created disabled as top-level-sheet-panes.


dkochmanski:
Looks good, thanks!

PR #221, opened 2017-06-03
Bezier clx draw rewrite

slyrus:
Now we properly set a pixmap-based mask based on the positive and negative areas of a bezier difference when drawing with the CLX backend.


dkochmanski:
Looks good, tested it with clim-demo and works very well, thanks!

Issue #144, opened 2017-02-09
push-button panes ignore width and height


jakykong:
It appears that creating a push-button pane with width and height, the width and height are ignored. I talked briefly with jackdaniel on #clim and he suggested :y-spacing/:x-spacing as a workaround. 

Here is a minimal example demonstrating the problem. The expected outcome is that the button labeled "Huge Button" should be taller than my screen. The actual outcome is it is about 32px tall. (Width is similar - it is actually a couple hundred px wide, instead of 60.)

```
(define-application-frame bug-app ()
  ()
  (:panes 
    (app :application)
    (btn (make-pane 'push-button 
                       :height 1000 :width 60 
                       :label "Huge Button")))
  (:layouts (default (vertically () app btn))))

(defun bug-main ()
  (run-frame-top-level (make-application-frame 'bug-app)))
```

gas2serra:
The problem is that push-button set 'max-height' to the exact size.

@dkochmanski do you think we can set it to +fill+?

```
(defmethod compose-space ((gadget push-button-pane) &key width height)
  (declare (ignore width height))
  (space-requirement+* (space-requirement+* (compose-label-space gadget)
                                            :min-width (* 2 (pane-x-spacing gadget))
                                            :width (* 2 (pane-x-spacing gadget))
                                            :max-width (* 2 (pane-x-spacing gadget))
                                            :min-height (* 2 (pane-y-spacing gadget))
                                            :height (* 2 (pane-y-spacing gadget))
                                            :max-height (* 2 (pane-y-spacing gadget)))				    
					    )
                       :min-width (* 2 *3d-border-thickness*)
                       :width (* 2 *3d-border-thickness*)
                       :max-width (* 2 *3d-border-thickness*)
                       :min-height (* 2 *3d-border-thickness*)
                       :height (* 2 *3d-border-thickness*)
                       :max-height (* 2 *3d-border-thickness*)))
```

dkochmanski:
fixed by #222.

PR #222, opened 2017-06-04
set max-width/height to +fill+

gas2serra:
Issue #144.

PS: The look and feel of several apps changes.

### DIFF
--- a/Core/clim-core/gadgets.lisp
+++ b/Core/clim-core/gadgets.lisp
@@ -795,8 +795,8 @@ and must never be nil."))
            (ds (text-style-descent (pane-text-style gadget) gadget))
            (w  (+ (text-size gadget label :text-style (pane-text-style gadget)) wider))
            (h  (+ as ds higher)))
-      (make-space-requirement :width w  :min-width w  :max-width  w
-                              :height h :min-height h :max-height h))))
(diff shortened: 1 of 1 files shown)


dkochmanski:
I wonder myself whenever max-height should be always `+fill+` – that definetely depends on the gadget.

For instance, taking example from material.io, button height is fixed (unit is dp), see:
https://material.io/guidelines/components/buttons.html#buttons-style

dkochmanski:
what is arguably the right thing- then button is always a button, if someone wants to have "touchable pane", then a new gadget may be created.

dkochmanski:
using fixed sizes (in dp so it looks good on different kind of screens) would make our life easier I think (this isn't the right place to discuss it, just thinking out lould)

gas2serra:
> that definetely depends on the gadget.
Yes.
> using fixed sizes (in dp so it looks good on different kind of screens) would make our life easier
I agree. 

The problem is that user cannot change it.
If we use +fill+ the user has more freedom.

If the user want be conservative her can use:
` :max-width '(1.0 :point :relative)` 
or adding +fill+ in the container.

dkochmanski:
If we implement material design it is up to implementer to provide "fixed" values for this style.

Issue #218, opened 2017-06-02
climacs window moving


gas2serra:
Every time you type a key in climacs the window change its position.

gas2serra:
Broken by:
> commit ee53f1e00c495ba3d1dd036ae2eeeffd852edf6e (HEAD)
> Author: Nisar Ahmad <nisarahmad1324@gmail.com>
> Date:   Fri Apr 7 17:11:13 2017 +0530
> 
>     frames.lisp : implement menu-bar update after changes
>     
>     since (setf frame-command-table) is also called by accepting-values
>     dialog, we need to detect and ignore them because generating-panes
>     call is expensive.
It was ok:
> commit 91fa8c88be1a241fa75e9fc2382197bb622acec3
> Author: Nisar Ahmad <nisarahmad1324@gmail.com>
> Date:   Thu Apr 6 15:34:27 2017 +0530
> 
>     frames.lisp : fixes generate-panes fails to repaint properly
>     
>     This should be the correct solution of #146. See #Issue-21 on github
>     repo for detailes.
>

gas2serra:
Esa-top-level does '(setf (frame-command-table ,frame) command-table)' 
```
 `(defmethod esa-top-level (,frame &key
                             (,command-parser 'esa-command-parser)
                             ;; FIXME: maybe customize this?  Under what
                             ;; circumstances would it be used?  Maybe try
                             ;; turning the clim listener into an ESA?
                             (,command-unparser  'command-line-command-unparser)
                             (,partial-command-parser 'esa-partial-command-parser)
                             (,prompt "Extended Command: "))
     ,(let ((frame (unlisted frame)))
           `(with-slots (windows) ,frame
              (let ((*standard-output* (car windows))
                    (*standard-input* (frame-standard-input ,frame))
                    (*minibuffer* (minibuffer ,frame))
                    (*print-pretty* nil)
                    (*abort-gestures* *esa-abort-gestures*)
                    (*command-parser* ,command-parser)
                    (*command-unparser* ,command-unparser)
                    (*partial-command-parser* ,partial-command-parser)
                    (*extended-command-prompt* ,prompt)
                    (*pointer-documentation-output*
                     (frame-pointer-documentation-output ,frame))
                    (*esa-instance* ,frame))
                (unless (eq (frame-state ,frame) :enabled)
                  (enable-frame ,frame))
                (redisplay-frame-panes ,frame :force-p t)
                (loop
                   do (restart-case
                          (handler-case
[...]

PR #224, opened 2017-06-04
[WIP] fix setf of frame-command-table

gas2serra:
For #218.

I added '%change-box-pane-contents'. 
What do you think?

@dkochmanski Sorry for "spacing" difference. How can I avoid them? How can I configure my emacs?


dkochmanski:
put `(setq-default indent-tabs-mode nil)` in your emacs

Also, you may call:
`M-x whitespace-cleanup`
which cleans all tabs in the current buffer.

gas2serra:
Thanks!

PR #230, opened 2017-06-04
fix setf of frame-command-table

gas2serra:
For #218.

PS: It modifies a macro! You need to clear your cache.

### DIFF
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -113,6 +113,7 @@ input focus. This is a McCLIM extension."))
 		    :reader frame-top-level-sheet)
    (menu-bar :initarg :menu-bar
 	     :initform nil)
+   (menu-bar-pane :initform nil)
    (state :initarg :state
 	  :initform :disowned
 	  :reader frame-state)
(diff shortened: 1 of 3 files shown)


Core/clim-core/frames.lisp
```diff
@@ -240,7 +241,9 @@ documentation produced by presentations.")
     (if (and (get-menu (frame-command-table frame))
 	     (get-menu new-command-table))
 	(prog1 (call-next-method)
-	  (generate-panes *default-frame-manager* *application-frame*))
+	  (when (slot-value frame 'menu-bar-pane)
+	    (update-menu-bar (slot-value frame 'menu-bar-pane)
+			     new-command-table)))
```
dkochmanski: if there is no possibility, that slot-value chagnes between condition check and call update-menu-bar, then please use

```
(alexandria:when-let ((menu-bar-pane (slot-value frame -menu-bar-pane)))
  (update-menu-bar menu-bar-pane …))
```

Core/clim-core/frames.lisp
```diff
@@ -113,6 +113,7 @@ input focus. This is a McCLIM extension."))
 		    :reader frame-top-level-sheet)
    (menu-bar :initarg :menu-bar
 	     :initform nil)
+   (menu-bar-pane :initform nil)
    (state :initarg :state
```
dkochmanski: since we set it etc., maybe it is worth to give it a proper accessor?

dkochmanski:
thanks!

PR #231, opened 2017-06-04
with-clx-graphics macro and bezier drawing fixes

slyrus:
These changes fix the with-clx-graphics macro and improve bezier drawing. see commit logs for details.


dkochmanski:
looks good, thanks!

PR #235, opened 2017-06-05
Minor fixes

slyrus:
default argument fix and a variable to ignore


Core/clim-basic/stream-output.lisp
```diff
@@ -455,7 +455,7 @@ STREAM-STRING-WIDTH will be called."))
                                    &key (first-quadrant t)
                                         height
                                         (move-cursor t)
-                                        (record-type ''standard-sequence-output-record))
+                                        (record-type 'standard-sequence-output-record))
                                   &body body)
```
dkochmanski: I don't think this is a typo – the goal is to pass quoted argument from macro to function, so the function call is
`(fun x y 'my-type)` which is expanded from `'(fun ,my-x ,my-y ,my-type)`

PR #242, opened 2017-06-08
improve space requirement protocol

gas2serra:
Make the space req protocol closer to the spec.
- the user space requirements are never changed 
-  'redisplay-frame-pane' calls 'change-space-requirements' only if a user req is equal to :compute. This improves the overall performance of clim.
- stream-output does not depend on pane protocol.
- 'window-clear' does not change user space width and height to zero.This improves the overall performance of clim.


dkochmanski:
Looks good, thanks!

PR #243, opened 2017-06-08
fix clear completition record

### LINKED ISSUE #31: [$100] menu for input-prompt in lisp-listener does not disapear after use
After clicking the "Filesystem" menu item and then "Display Image" I right-click on the input prompt to get a list of available files. After clicking on the desired file, the menu does not disappear. The menu is also sometimes drawn on top of the input-prompt, making it difficult to cancel the operation without choosing a file first.

The result can look like the image attached
screenshot from 2016-06-07 20-11-06

gas2serra:
Fix #31


dkochmanski:
fix confirmed, please claim a bounty at https://www.bountysource.com/issues/34948387-100-menu-for-input-prompt-in-lisp-listener-does-not-disapear-after-use . Thanks!

PR #253, opened 2017-06-11
fix compose-space with :compute requirement

### LINKED ISSUE #251: When using incremental-redisplay, the size of the pane doesn't change when the content changes
To reproduce this problem, run the test case below. Issue the command `Rows` and type a number that causes the pane to display more content than fits in the window, for example 100.

Then, type the same command again, but choose a smaller number. Notice how the size of the pane does not change and you can scroll into a large blank area below.

Another problem which this test program does not reproduce is that if you use `UPDATING-OUTPUT` and an output record changes such that its vertical dimension becomes larger, the pane will also not accommodate and in this case the problem is that some content at the end of the pane will now be hidden.

```
(clim:define-application-frame resize-test-frame ()
  ((rows :initform 0
         :accessor resize-test-frame/rows))
  (:panes (content :application
                   :display-function 'display-content
                   :display-time t
                   :incremental-redisplay t)
          (interaction-pane :interactor))
  (:layouts (default content
                     interaction-pane)))

(defun display-content (frame stream)
  (loop
    for i from 0 below (resize-test-frame/rows frame)
    do (format stream "Row number: ~a~%" i)))

(defun open-resize-test-frame ()
  (let ((frame (clim:make-application-frame 'resize-test-frame)))
    (clim:run-frame-top-level frame)))

(define-resize-test-frame-command (rows :name "Rows")
    ((n 'integer))
  (let ((frame clim:*application-frame*))
    (setf (resize-test-frame/rows frame) n)
    (setf (clim:pane-needs-redisplay (clim:find-pane-named frame 'content)) t)))
```

gas2serra:
fix #251.

### DIFF
--- a/Core/clim-core/panes.lisp
+++ b/Core/clim-core/panes.lisp
@@ -2491,8 +2491,13 @@ order to produce a double-click")
   (declare (ignore width height))
   (flet ((compute (val default)
 	   (if (eq val :compute) default val)))
-    (if (or (eq (pane-user-width pane) :compute)
-            (eq (pane-user-height pane) :compute))
+    (if (or
+         (eql :compute (pane-user-width pane))
+         (eql :compute (pane-user-min-width pane))
+         (eql :compute (pane-user-max-width pane))
+         (eql :compute (pane-user-height pane))
+         (eql :compute (pane-user-min-height pane))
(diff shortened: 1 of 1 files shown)


dkochmanski:
Thanks!

PR #256, opened 2017-06-12
Added output recording code for draw-rectangles*

slyrus:
* Fixes PostScript output with draw-rectangles*, which was totally
   broken before.

 * also, may fix Drawing Test "04) Rectangle Basic"


dkochmanski:
Looks good to me, merging. CLX still has problems, as seen on screenshots:

Before fix:
04-rectangle-basic-before

After fix:
04-rectangle-basic-after

PR #260, opened 2017-06-13
wip: Cl pdf backend

slyrus:
First pass at a CL-PDF based backend to generate proper PDF files. There's probably a ton of missing functionality (particularly designs/pixmaps), but this is a good place to start.

### DIFF
--- a/Backends/PostScript/font/package.lisp
+++ b/Backends/PostScript/font/package.lisp
@@ -1,7 +1,7 @@
 ;;; -*- Mode: Lisp; Package: CLIM-POSTSCRIPT -*-
 
-;;;  (c) copyright 2002 by
-;;;           Alexey Dejneka (adejneka@comail.ru)
+;;;  (c) copyright 2017 by
+;;;           Cyrus Harmon (cyrus@bobobeach.com)
 
 ;;; This library is free software; you can redistribute it and/or
 ;;; modify it under the terms of the GNU Library General Public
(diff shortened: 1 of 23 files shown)


slyrus:
Don't merge this yet. I still need to a little package hackery on the clim-postscript-font stuff.

slyrus:
Ok, htat has been fixed by 9f223f5.

Backends/PostScript/font/package.lisp
```diff
@@ -0,0 +1,40 @@
...
+                #:ellipse-normal-radii*
+
+                #:get-transformation
+                #:untransform-angle
+                #:with-transformed-position
+
+                #:maxf
+
+                #:port-text-style-mappings))
+
```
dkochmanski: one "extra" newline would do here, or even better, comment describing why this block of imports is separated

Backends/PostScript/font/standard-metrics.lisp
```diff
@@ -1,5 +1,6 @@
 
-(in-package :clim-postscript) 
+(in-package :clim-postscript-font)
+
```
dkochmanski: why do we have new package clim-postscript-font? because it is reused by pdf backend? If yes, shouldn't it be named `clim-document-font` or something, and put in `Backends/Standard`?

Backends/PostScript/graphics.lisp
```diff
@@ -468,8 +510,8 @@ setmatrix")
 (defmethod postscript-set-graphics-state (stream medium
                                           (kind (eql :text-style)))
   (let* ((font-name (medium-font medium))
-         (font (%font-name-postscript-name font-name))
-         (size (%font-name-size font-name)))
+         (font (clim-postscript-font::%font-name-postscript-name font-name))
+         (size (clim-postscript-font::%font-name-size font-name)))
     (pushnew font (slot-value (medium-sheet medium) 'document-fonts)
```
dkochmanski: shouldn't we export these, if they are used from the outside of the package?

Backends/PostScript/font/package.lisp
slyrus: yeah, that's fine. we'll need to export more stuff over time (see below). I wouldn't worry about the lines, ATM.

Backends/PostScript/font/standard-metrics.lisp
slyrus: Yeah, it's used by both the postscript and PDF backends. The reason it's not clim-document-font, is that I thought we were dealing with postscript fonts, but I guess you're right. Could be renamed.

Backends/PostScript/graphics.lisp
slyrus: Yes, eventually.

slyrus:
I have a new branch that is off of the post-scigraph-cleanup master.

PR #262, opened 2017-06-14
add image extension

gas2serra:
I moved all code about images into a new package. The goal is to provide a single extension as described in  McCLIM User’s Manual based on opticl.


Extensions/image/package.lisp
```diff
@@ -0,0 +1,63 @@
...
+   #:make-pattern-from-bitmap-file
+   #:medium-data
+   ))
+
+(in-package :mcclim-image)
+
+;;;
+;;; import symbols into climi
+;;;
+
```
dkochmanski: why can't we import these in the package definition?

dkochmanski:
```
ADE> (export 'xxx)
T
ADE> (in-package cl-user)
#<PACKAGE "COMMON-LISP-USER">
CL-USER> (import 'eu.turtleware.ade:xxx )
T
CL-USER> (defun xxx () "bam")
XXX
CL-USER> (in-package eu.turtleware.ade)
#<PACKAGE "EU.TURTLEWARE.ADE">
ADE> (xxx)
"bam"
```

gas2serra:
Hi, 
in [1], I found "Drawing Two-Dimensional Images " chapter. This chapter is not present in the current documentation. Is it deprecated? Do you think Robert need it?

When the API will be stable I'll add all symbols into 'clim-extensions'.

[1] https://common-lisp.net/project/mcclim/static/documents/mcclim.pdf

dkochmanski:
Documentation is build from Texinfo sources. I think it was never present in Latex sources. This doc duplication is annoying, I'd gladly get rid of the LaTeX one, but Robert insists to have it, so all I do is that I try to backport changes from one to another from time to time.

See file Documentation/Manual/Texinfo/extensions.txi

Extensions/image/package.lisp
dkochmanski: what I have meant is that can't these symbols be imported in
`:import-from` clause in the defpackage above?

gas2serra: > can't these symbols be imported in :import-from clause in the defpackage above?
No, I want an ':import-to'.

These symbols are imported into clim-internals to make the old code running. I hope to be able to fix the situation in the future.

I think the correct place for the public API is in the 'clim-extensions' package.

gas2serra:
Ok.

gas2serra:
I try to fix all symbols and than I'll submit.

PR #266, opened 2017-06-15
Fix FIND-STROKE-CONTAINING-OFFSET & [$200] Interactor CLI prompt print problem

gabriel-laddel:
Fixes https://github.com/robert-strandh/McCLIM/issues/265

### DIFF
--- a/Core/clim-core/presentation-defs.lisp
+++ b/Core/clim-core/presentation-defs.lisp
@@ -1286,11 +1286,12 @@ protocol retrieving gestures from a provided string."))
       (princ object stream)))
 
 (define-presentation-method accept ((type symbol) stream (view textual-view)
-                                    &key)
+                                                  &key (default-type type)
+                                                  default)
   (let ((read-result (accept-using-read stream type)))
-    (if (and (null read-result) default)
-	(values default type)
-	(values read-result type))))
+    (if (and (null read-result) default) 
+        (values default type)
+        (values read-result type))))
 
 (define-presentation-type keyword () :inherit-from 'symbol)
 
@@ -1385,13 +1386,14 @@ protocol retrieving gestures from a provided string."))
         (*print-radix* radix))
     (princ object stream)))
 
-(define-presentation-method accept ((type real) stream (view textual-view)
-                                    &key)
+(define-presentation-method accept ((type real) stream (view textual-view) &key 
+                                                (default-type type)
+                                                default)
   (let ((*read-base* base)
-	(read-result (accept-using-read stream type)))
-    (if (and (null read-result) default)
-	(values default type)
-	(values read-result type))))
+         (read-result (accept-using-read stream type)))
+    (if (and (null read-result) default) 
+        (values default type)
+        (values read-result type))))
 
 ;;; Define a method that will do the comparision for all real types.  It's
 ;;; already determined that that the numeric class of type is a subtype of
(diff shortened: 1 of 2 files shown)


Core/clim-core/presentation-defs.lisp
```diff
@@ -1285,7 +1287,10 @@ protocol retrieving gestures from a provided string."))
 
 (define-presentation-method accept ((type symbol) stream (view textual-view)
                                     &key)
-  (accept-using-read stream type))
+  (let* ((read-result (accept-using-read stream type)))
+    (if (and (null read-request) default)
```
dkochmanski: shouldn't it be read-result (instead of read-request)?

dkochmanski: also, if it is only one-clause let, please make it `let` not `let*` (cosmetic, but since you change this part of the code anyway it would be easy).

gabriel-laddel: Oh ffs, yeah sorry about that.

gabriel-laddel: Do I have to submit a new commit, or do you have the ability to pull this with those changes?

dkochmanski: if you push to your branch which PR was issued from, pull request should be updated automatically

gabriel-laddel: Done.

Core/clim-core/presentation-defs.lisp
```diff
@@ -1285,7 +1287,10 @@ protocol retrieving gestures from a provided string."))
 
 (define-presentation-method accept ((type symbol) stream (view textual-view)
                                     &key)
-  (accept-using-read stream type))
+  (let ((read-result (accept-using-read stream type)))
+    (if (and (null read-result) default)
+	(values default type)
```
dkochmanski: still default is unbound

Core/clim-core/presentation-defs.lisp
```diff
@@ -1382,8 +1387,11 @@ protocol retrieving gestures from a provided string."))
 
 (define-presentation-method accept ((type real) stream (view textual-view)
                                     &key)
-  (let ((*read-base* base))
-    (accept-using-read stream type)))
+  (let ((*read-base* base)
+	(read-result (accept-using-read stream type)))
+    (if (and (null read-result) default)
+	(values default type)
```
dkochmanski: same probably applies here (default unbound)

gabriel-laddel: Oh ffs. I'm going to re-create the repo, apply the changes again so there don't have to be extra commits in master.

What 'indenter' do you use? The only reason we're here is because I'm typing s-expressions out by hand rather than copy-pasting my ALREADY WORKING ones into climacs, which indents them differently than you like.

dkochmanski: no worries. You may add new commit to the same master branch, we use --no-ff option, so it won't clobber main branch anyways.

As of indenter, I use emacs+slime default indentation with McCLIM loaded (and I have `(setq-default indent-tabs-mode nil)` in my init.el)

dkochmanski:
Looks good, thanks! Please claim a bounty at https://www.bountysource.com/issues/39478750-200-interactor-cli-prompt-print-problem

PR #269, opened 2017-06-16
Pdf clip

slyrus:
These changes add rectangular clipping paths to the PDF backend.

### DIFF
--- a/Backends/PDF/graphics.lisp
+++ b/Backends/PDF/graphics.lisp
@@ -224,18 +224,22 @@
     (let ((x0 (point-x (first points)))
           (y0 (point-y (first points))))
       (with-transformed-position (tr x0 y0)
-        (cl-pdf:move-to x0 y0)
+        (pdf:move-to x0 y0)
         (loop for point in (rest points)
            do
              (let ((x1 (point-x point))
                    (y1 (point-y point)))
                (with-transformed-position (tr x1 y1)
-                 (cl-pdf:line-to x1 y1))))))
-    (cl-pdf::end-path-no-op)))
+                 (cl-pdf:line-to x1 y1))))
+        (pdf:clip-path)
+        (pdf:end-path-no-op)))))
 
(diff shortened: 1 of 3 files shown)


Backends/PDF/graphics.lisp
```diff
@@ -269,27 +336,23 @@
...
-(defmethod pdf-set-clipping-region (region)
+(defmethod pdf-set-clipping-region (medium region)
   ;; this is not yet implemented!
-  #+(or)
   (progn
-    (pdf-add-path region)
-    (pdf:clip-path)))
+    #+nil (print region)
+    (pdf-add-path medium region)
+    ))
```
dkochmanski: please add `#+(or)` instead of `#+nil` and put closing parens in the same line as the last form

`(pdf-add-path medium region)))`

Extensions/bezier/bezier.lisp
```diff
@@ -949,16 +949,39 @@ second curve point, yielding (200 50)."
...
+      (clim-pdf::pdf-actualize-graphics-state medium :color :line-style)
+      (dolist (area (areas design))
+        (%pdf-draw-bezier-curve (transform-region tr area))
+        (pdf:close-fill-and-stroke)))))
+
+(defmethod medium-draw-bezier-design*
+    ((medium clim-pdf::pdf-medium) (design bezier-difference))
+  (let ((tr (sheet-native-transformation (medium-sheet medium))))
+    (declare (ignore tr))
+    (cl-pdf:with-saved-state
```
dkochmanski: if we ignore `tr` – why do we call sheet-native-transformation on it? for side-effects?

Backends/PDF/graphics.lisp
slyrus: Fixed, thanks.

Extensions/bezier/bezier.lisp
slyrus: this wasn't really working before. Now it is.

PR #270, opened 2017-06-16
add a 1 unit margin to PDF output

slyrus:
modifies the pdf-transformation to allow for a one pixel margin. without this, text would get cutoff at the top of the page.

### DIFF
--- a/Backends/PDF/paper.lisp
+++ b/Backends/PDF/paper.lisp
@@ -35,13 +35,22 @@
       (rotatef width height))
     (make-rectangle* 0 0 width height)))
 
+(defparameter *pdf-top-margin* 1)
+(defparameter *pdf-left-margin* 1)
+(defparameter *pdf-bottom-margin* 1)
+(defparameter *pdf-right-margin* 1)
+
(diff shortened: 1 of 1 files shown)


Backends/PDF/paper.lisp
```diff
@@ -35,13 +35,22 @@
       (rotatef width height))
     (make-rectangle* 0 0 width height)))
 
+(defparameter *pdf-top-margin* 1)
+(defparameter *pdf-left-margin* 1)
+(defparameter *pdf-bottom-margin* 1)
+(defparameter *pdf-right-margin* 1)
+
```
dkochmanski: do we intend to make the margin size customizable? if not, we could just put below `(- height 2)` etc with a brief comment.

slyrus: Yes, the margin size should be customizable. I'm not sure I want to commit to a particularly exported interface yet though. I much prefer the special var's to, e.g., (- height 2) and a comment. I'd like to see this committed as is and some subsequent thought given to how spec margins for printed objects across both PS and PDF backends.

PR #274, opened 2017-06-21
add sequence-type function and use it in map calls

slyrus:
* this gets around the problem of declaring the result type from map
   based on the length of the vector in the case where the
   fill-pointer indicates that there are actually fewer elements of
   the sequence than type-of reports.

### DIFF
--- a/Extensions/bezier/bezier.lisp
+++ b/Extensions/bezier/bezier.lisp
@@ -317,10 +317,19 @@ second curve point, yielding (200 50)."
 (defmethod segments ((design bezier-design))
   (%segments design))
 
+(defun sequence-type (seq)
+  (let ((type (type-of seq)))
+    (if (and (listp type)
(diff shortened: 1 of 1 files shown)


Extensions/bezier/bezier.lisp
```diff
@@ -317,10 +317,19 @@ second curve point, yielding (200 50)."
...
 
+(defun sequence-type (seq)
+  (let ((type (type-of seq)))
+    (if (and (listp type)
+             (eq (car type) 'vector)
+             (not (equal (third type)
+                         (length seq))))
+        (car type)
+        type)))
+
```
dkochmanski: won't traversing the sequence nullify the effect of knowing the exact size of it anyway?

Maybe we could just always force on transform-region type `vector`

`(map 'vector … …)` ?

slyrus:
New PR coming.

PR #280, opened 2017-06-23
Output recording coord seq mixin

slyrus:
These changes make it so that the output-recording coord-seq can be a seq, not just a vector, and fix a bunch of bugs encountered along the way. Previously, the functions under the covers only worked with vectors, but the drawing code for draw-lines, draw-points, draw-polygon, etc... only worked if we passed in a list which was copied to a vector. Now we can pass in either a list or a vector and the code under the covers should work with both lists and vectors. We still copy the sequence, as the copy gets transformed during the playback of the output record, but things are much less broken.

draw-rectangles, OTOH, was completely broken in multiple ways. It, too, is fixed, using the coord-seq-mixin infrastructure.

In general, all this output-recording/drawing code is a big hacky mess, but it's slightly less broken now.

### DIFF
--- a/Backends/CLX/medium.lisp
+++ b/Backends/CLX/medium.lisp
@@ -847,14 +847,17 @@ time an indexed pattern is drawn.")
 				(medium-sheet medium))
                                position-seq)
     (with-clx-graphics () medium
-      (loop
-	 for (left top right bottom) on position-seq by #'cddddr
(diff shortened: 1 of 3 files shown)


dkochmanski:
I've read it, all looks good to me - no style remarks or anything. Merging. Thanks!

Issue #175, opened 2017-04-26
[$100] Caps lock affects non-alphabetic keys


abridgewater:
Seems like it's behaving as Shift Lock, not Caps Lock, so on my (US QWERTY layout) keyboard, 5 becomes % and , becomes <.

dkochmanski:
Some scraps from IRC:

```
< nyef`> The caps lock modifier needs to affect "A" but not "3". A shift lock modifier would need to affect both.
< nyef`> The implication that I'm seeing is that caps lock needs to apply only if the key name is #'ALPHA-CHAR-P, while shift-lock always applies.
< nyef`> ... Except that we then also need to consider the implications of the actual SHIFT keys.
< nyef`> ... And possibly *both* SHIFT keys (and there's an implicit assumption here that there are only *two* SHIFT keys).
< nyef`> ... Seems like there are (at least) three scenarios here. One is straight X11 core protocol, one is the X Keyboard Extension, and one is the X Input  Extension.
< nyef`> It would not surprise me in the least if CLX is limited to the X11 core protocol here.
```

abridgewater:
Small amount of missing context for the above IRC scrap (the immediately preceding line):

```<nyef`> Oh, FFS. "3" on my keyboard is keycode 12.  Keysyms 0-3 are 51 35 51 35. "A" on my keyboard is keycode 38.  Keysyms 0-3 are 97 65 97 65.```

The point here being that they have the exact same pattern of keysyms (unshifted, shifted, unshifted, shifted), but require different handling with respect to shift lock and caps lock.

**Comment by abridgewater** [comment]
More from IRC:
```
<nyef`> Aha! From the CLX manual: "When the :lock modifier is on and is interpreted as :caps-lock, it is suggested that the :shift modifier first be applied to choose a keysym, but if that keysym is lowercase alphabetic, the corresponding uppercase keysym should be used instead." Section 15.4.1 Keycodes and Keysyms in my local copy.
<nyef`> Hrm. That doesn't seem quite right.
<nyef`> Shift + caps-lock on an alpha key needs to produce minuscules, but the logic described by the CLX manual would produce majuscules.
<nyef`> Still, this should be enough for someone to hack up Backends/CLX/keysyms.lisp / X-EVENT-TO-KEY-NAME-AND-MODIFIERS to do something more reasonably sane.
```
And that's that. I'm willing to test potential fixes once someone implements one.

MatthewRock:
[...]

PR #281, opened 2017-06-27
Fix caps lock affecting non-alphabetic keys

MatthewRock:
This is the fix to the issue #175 

The problem with caps lock was that 
(a) it was not recognized when caps-lock key was pressed (instead, the shift-lock was recognized)
(b) it did not handle the casing properly

I have tested this fix on my machine. I have changed the right shift with "Shift_Lock" modifier (instructions can be found here: http://xahlee.info/linux/linux_shift_lock_key.html )

When pressing the caps-lock, it is recognized and applied to the keysym.
When pressing shift, it is recognized and applied to the keysym.
When pressing both, they are recognized and applied correctly.

I have also removed shift-lock from the files I have edited. The reason for that is the way the shift-lock works; rather than triggering shift-lock key, it actually works as if the user was holding the shift all the time. This leaves shift-lock modifier without purpose.

Side note: I have found mentions of shift-lock in Backends/CLX/standard-keysym-rules , but I don't really know how the file is used. Should I somehow try to remove the shift-lock from there, too?

### DIFF
--- a/Backends/CLX/keysyms-common.lisp
+++ b/Backends/CLX/keysyms-common.lisp
@@ -89,12 +89,12 @@
 ;;; reflect the post event state.
 
 (defun x-keysym-to-clim-modifiers (port event-key keychar keysym-name state)
-  "event-key is :key-press or :key-release"
-  (multiple-value-bind (clim-modifiers shift-lock? caps-lock? mode-switch?)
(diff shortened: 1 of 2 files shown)


dkochmanski:
Works fine for me (and code looks good). I'm merging it as is. Please claim a bounty on bountysource. Thanks!

PR #287, opened 2017-07-09
Text editor pane value changed callback fix (with bonus spelling fixes)

scymtym:
The way this was implemented previously, namely passing the value of the ``value-changed-callback`` _initarg_ to the substrate object, bypassed the ``value-changed-callback`` _method_ of the ``text-editor-pane`` instance, violating the description of ``value-changed-callback`` in [30.3 (Basic Gadget Classes)](http://bauhh.dyndns.org:8000/clim-spec/30-3.html#_6734).

With this change a value-changed callback is installed in the substrate object that invokes the ``value-changed-callback`` method for the ``text-editor-pane``.

Another (potential) benefit of this change is that the ``id`` and ``client`` initargs can be handled by the ``text-editor-pane`` instance normally instead of being passed to the substrate object. This seems more
accurate and more robust (e.g. in case the id and/or client were to change in the ``text-editor-pane`` instance for some reason, the new values would be "picked up").

### DIFF
--- a/Core/clim/text-editor-gadget.lisp
+++ b/Core/clim/text-editor-gadget.lisp
@@ -212,21 +212,21 @@ cause the activate callback to be called."))
   (:default-initargs :activation-gestures '()))
 
 (defmethod initialize-instance :after ((object text-editor-pane)
-                                       &key id client armed-callback
+                                       &key armed-callback
                                          disarmed-callback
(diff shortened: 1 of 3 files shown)


dkochmanski:
Looks very good, thanks (I have no critical remarks)! Merging.

PR #288, opened 2017-07-10
Render: new version

gas2serra:
Hi Daniel,
I redesign (a little bit - I hope for the last time) the render. I:
* make a more general image library (adding also a test application for it)
* improve the performance
* split the project in different sub projects

There are still some work to do. I hope I'll finish it when I will return from the vacation.

Regards,
Alessandro


dkochmanski:
Hey, thanks! I will test it when I'll find a bit of time, and I'll merge it then. From brief look at the code it looks OK (though the codebase grows if we take lines of code metric). Btw, I've asked Robert to add you to the contributors, so you have the necessary bits to administer the main repository (I like the peer review scheme, but if you have some obvious small fix, you may push directly to master).

Most importantly, have a nice holiday!

gas2serra:
Thanks.

PR #291, opened 2017-07-18
Add graph-toy to clim demos.

knusbaum:
This adds my little graph toy to the mcclim demos package and the demodemo list of demos.


dkochmanski:
Looks good, thanks!

PR #292, opened 2017-07-20
Spelling and other small changes.

phmarek:
Please take a look at these; not all might be usable as-is, because I'm not sure about some conflicting wording.

And no, I won't tell which changes, because if you don't know that you'll read every one carefully ;)

Thanks a lot.

### DIFF
--- a/Documentation/Specification/silica.tex
+++ b/Documentation/Specification/silica.tex
@@ -1798,8 +1798,8 @@ with which \arg{sheet} is associated.  Otherwise it returns \cl{nil}.
 \subsection {Internal Interfaces for Native Coordinates}
 
 \issue {SWM} {Do these functions work on any sheet, or only on sheets that have
-a mirror, or only on sheets that have a direct mirror?  Also, define what a
-``native coordinate'' are.  Also, do \cl{sheet-device-transformation} and
+a mirror, or only on sheets that have a direct mirror?  Also, define what
+``native coordinates'' are.  Also, do \cl{sheet-device-transformation} and
 \cl{sheet-device-region} really account for the user's transformation and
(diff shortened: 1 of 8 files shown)


Documentation/Specification/silica.tex
```diff
@@ -1249,14 +1249,14 @@
 This macro will signal a runtime error if sheet is not a subclass of
 \cl{sheet-with-medium-mixin}.
 
-The \arg{medium} argument is not evaluated, and must be a symbol that is bound
+The \arg{medium} argument is not evaluated, and must be a symbol that will be bound
 to a medium.  \arg{body} may have zero or more declarations as its first forms.
```
dkochmanski: I had doubts about that, so checked how it's implemented. All looks fine though, `will be` describes the behavior better (also, concurring interpretation would cease reason of `with-sheet-medium-bound` existance).

Documentation/Specification/silica.tex
```diff
@@ -1749,7 +1749,7 @@
 Returns the mirror of the \term{sheet} \arg{sheet}.  If the sheet is not a
 subclass of \cl{mirrored-sheet-mixin}, this will return \cl{nil}.  If the sheet
 is a subclass of \cl{mirrored-sheet-mixin} and does not currently have a mirror,
-\cl{sheet-mirror} will return \cl{nil}.
+\cl{sheet-direct-mirror} will return \cl{nil}.
```
dkochmanski: technically it is not wrong statement, but I get that given the context it may be confusing.

dkochmanski:
All looks fine to me. I had two comments, but I agree with the corrections. Merrging. Thanks!

PR #295, opened 2017-07-27
Use Alexandria-provided macros and functions

pefimo:
* PARSE-ORDINARY-LAMBDA-LIST
* MAKE-KEYWORD
  McCLIM's own implementation used to be a no-op if the argument was a
  keyword.  However, CL:INTERN handles keywords gracefully.
* ONCE-ONLY
* WITH-GENSYMS
* CLAMP

### DIFF
--- a/Libraries/ESA/packages.lisp
+++ b/Libraries/ESA/packages.lisp
@@ -27,8 +27,7 @@
   (:shadowing-import-from :clim-lisp #:describe-object)
   (:import-from #:alexandria
 		#:once-only
-		#:with-gensyms
-		)
+		#:with-gensyms)
   (:export #:with-gensyms
 	   #:once-only
            #:unlisted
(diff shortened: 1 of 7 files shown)


Libraries/ESA/packages.lisp
```diff
@@ -25,8 +25,12 @@
 (defpackage :esa-utils
   (:use :clim-lisp :clim-mop :clim)
   (:shadowing-import-from :clim-lisp #:describe-object)
+  (:import-from #:alexandria
+		#:once-only
+		#:with-gensyms
+		)
```
dkochmanski: please close sexp in the same line as last element

dkochmanski: since ESA depends on alexandria, please add explicit dependency in mcclim-esa.asd too

pefimo:
Done it is.

dkochmanski:
Thank you!

PR #304, opened 2017-08-11
Closer mop

dkochmanski:
Replace clim-mop with c2mop.


dkochmanski:
I've checked also on CCL and ECL. Added extra imports.

Works fine on ECL, on CCL we have problem with CLX, but it is unrelated to this PR. I'm merging this, thanks for the review.

PR #305, opened 2017-08-12
Pdf backend improvements

slyrus:
Various PDF enhancements. The big one (for me anyway) is that I can now make a PDF to a certain size, including just the size of the output.

 * add trim-page-to-output-size keyword argument to
   clim-pdf:with-output-to-pdf-stream

 * device-type can now be a list of the form (width height) to specify
   the page size, instead of being locked into the fix named paper
   sizes

 * landscape printing is better, but text prints rotated 90
   degrees (that is to say that it isn't rotated, as it should be).

 * add orientation accessor

 * make sheet-native-transformation be an acessor, not just a reader

### DIFF
--- a/Backends/PDF/class.lisp
+++ b/Backends/PDF/class.lisp
@@ -67,10 +67,11 @@
   ((file-stream :initarg :file-stream :reader clim-pdf-stream-file-stream)
    (title :initarg :title)
    (for :initarg :for)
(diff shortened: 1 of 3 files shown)


dkochmanski:
Looks good. It might be a good idea to write a small tutorial how to use it.

PR #309, opened 2017-08-26
Drei minor fixes

slyrus:
two minor cleanups for Drei

### DIFF
--- a/Libraries/Drei/lisp-syntax.lisp
+++ b/Libraries/Drei/lisp-syntax.lisp
@@ -81,7 +81,7 @@ syntax should be run.")
   (:name "Lisp")
   (:pathname-types "lisp" "lsp" "cl")
   (:command-table lisp-table)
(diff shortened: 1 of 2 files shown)


dkochmanski:
looks good, thanks!

Libraries/Drei/lisp-syntax.lisp
```diff
@@ -81,7 +81,7 @@ syntax should be run.")
   (:name "Lisp")
   (:pathname-types "lisp" "lsp" "cl")
   (:command-table lisp-table)
-  (:default-initargs :initial-state |initial-state |))
+  (:default-initargs :initial-state '|initial-state |))
```
dkochmanski: I'm reverting this change, it seems that `|initial-state |` is a defined parameter, not meant as a symbol:

```
(define-parser-state |initial-state | (form-may-follow) ())
```

this change breaks the listener at least.

PR #310, opened 2017-08-27
FIX #308: accepting-values: on `end' choice should return value of the body

### LINKED ISSUE #308: accepting-values: on `end' choice should return value of last form in body
Steps to reproduce:
1.  `(clim-listener:run-listener)`
2.  `(accepting-values (t) (accept 'string) 4)`

Expected: returns 4 when `OK` selected
Observed: returns `NIL`

Other "issue" is that not having `accept` in accepting-values puts us in the debugger. this cuold be no-op which expands to `(progn ,@body)` if no accept clauses are present or a window / dialog without innputs , but with OK/Cancel buttons.

Steps to reproduce:

1.  `(clim-listener:run-listener)`
2.  `(accepting-values (t) 4)`

Expected: returns 4 immedietly (or when `OK` selected- something to be discussed)
Observed: debugger complaining about no accept inside

http://bauhh.dyndns.org:8000/clim-spec/26.html#_1364

>  Builds a dialog for user interaction based on calls to accept within body. The user can select the values and change them, or use defaults if they are supplied. The dialog will also contain some sort of "end" and "abort" choices. If "end" is selected, then accepting-values returns whatever values the body returns. If "abort" is selected, accepting-values will invoke the abort restart. [annotate]

nsrahmad:


dkochmanski:
looks good, thanks! I'm nnot closing #308 yet (see my comment regarding clear error message).

PR #328, opened 2017-09-15
Fix accepting values

dkochmanski:
We provide intelligible message for malformed accepting-values call (see #308).
Alse multiple-values if present are returned from body.

### DIFF
--- a/Core/clim-basic/recording.lisp
+++ b/Core/clim-basic/recording.lisp
@@ -378,15 +378,13 @@ recording stream. If it is T, *STANDARD-OUTPUT* is used.")
 (defun highlight-output-record-rectangle (record stream state)
   (with-identity-transformation (stream)
     (multiple-value-bind (x1 y1 x2 y2)
-        (output-record-hit-detection-rectangle* record)
+        (bounding-rectangle* record)
       (ecase state
         (:highlight
          (draw-rectangle* (sheet-medium stream) (1+ x1) (1+ y1) (1- x2) (1- y2)
                           :filled nil :ink +foreground-ink+)) ; XXX +FLIPPING-INK+? 
         (:unhighlight
-	 ;; FIXME: repaint the hit detection rectangle. It could be
-	 ;; bigger than the bounding rectangle.
-	 (repaint-sheet stream record)
+         (repaint-sheet stream (bounding-rectangle record))
          ;; Using queue-repaint should be faster in apps (such as
          ;; clouseau) that highlight/unhighlight many bounding
          ;; rectangles at once. The event code should merge these into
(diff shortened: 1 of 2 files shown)


Core/clim-basic/recording.lisp
```diff
@@ -396,13 +394,11 @@ recording stream. If it is T, *STANDARD-OUTPUT* is used.")
          ;; out shortly after being drawn. So, we aren't ready for
          ;; this yet.  ..Actually, it isn't necessarily
          ;; faster. Depends on the app.
-         #+NIL
+         #+ (or)
```
slyrus: Do we really want to keep this around?

Core/clim-basic/recording.lisp
```diff
@@ -378,15 +378,13 @@ recording stream. If it is T, *STANDARD-OUTPUT* is used.")
 (defun highlight-output-record-rectangle (record stream state)
   (with-identity-transformation (stream)
     (multiple-value-bind (x1 y1 x2 y2)
-        (output-record-hit-detection-rectangle* record)
+        (bounding-rectangle* record)
```
slyrus: why does this change? is the comment about the hit detection rectangle below bogus?

Core/clim-core/dialog.lisp
```diff
@@ -245,8 +245,15 @@ accept of this query")))
                                       :record-type 'accepting-values-record)
                       (if align-prompts
                           (formatting-table (stream)
-                            (setf return-value (funcall body *accepting-values-stream*)))
-                            (setf return-value (funcall body *accepting-values-stream*)))
+                            #1=(setf return-values
+                                     (multiple-value-list
+                                      (funcall body *accepting-values-stream*))))
+                          #1#)
```
slyrus: why are we using this reader macro stuff here?

Core/clim-basic/recording.lisp
dkochmanski: I think that using `#+nil` is a bad practice becase it is wrong from semantical point of view, so I'd say yes, `#+(or)` is the only proper way to comment block of code.

Core/clim-basic/recording.lisp
dkochmanski: comment is alright (that hit detection rectangle may be bigger), but this function highlights output-record-rectangle, not its hit detection range. Also, this improves performance greatly for bigger output records with lots of repainting.

Core/clim-core/dialog.lisp
dkochmanski: I've taken this habit from Peter Norvig Lisp style guide (and I generally agree with that - if you have exactly same form repeated in a slightly different context one after another repeating text may lead to bugs, because someone may fix the form in one place but not in the another). If you don't agree, I can change that (so the same form is copied verbatim instead of using `#1#`).

Core/clim-basic/recording.lisp
dkochmanski: ah, you mean call to `queue-repaint`. I think that `queue-repaint` is the right thing here, so if we get acceptably faster at some point, I'd want to make it the default.

PR #332, opened 2017-10-10
rewrite render backend %lerp to use octets

slyrus:
* before we would occasionally call octet-mult with a negative
   number. This worked, but the function declamation says it should
   only take octets and recent SBCL compilers are smart enough to see
   that we were violating this assumption. So, rewrite %lerp to always
   call octet-mult with bona fide octets.

### DIFF
--- a/Extensions/render/color.lisp
+++ b/Extensions/render/color.lisp
@@ -35,7 +35,9 @@
 (declaim (inline %lerp)
 	 (ftype (function (octet octet octet) octet) %lerp))
 (defun %lerp (p q a)
-  (logand #xFF (+ p (octet-mult a (- q p)))))
+  (logand #xFF (if (>= q p)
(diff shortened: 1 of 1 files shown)


Extensions/render/color.lisp
```diff
@@ -35,7 +35,9 @@
 (declaim (inline %lerp)
 	 (ftype (function (octet octet octet) octet) %lerp))
 (defun %lerp (p q a)
-  (logand #xFF (+ p (octet-mult a (- q p)))))
+  (logand #xFF (if (>= q p)
+                   (+ p (octet-mult a (- q p)))
+                   (- p (octet-mult a (- p q))))))
```
dkochmanski: looks good to me. If there is no objections, I'm going to merge this in the afternoon.

Issue #164, opened 2017-04-25
[$100] keystroke accelerators may shadow keys even if inactive


dkochmanski:
You may call bug from menu, then string input box accepts letter `x`, but if you activate command with a keystroke, then `x` can't be typed (but it doesn't restart or start new instance of a command).

Expected result: if keystroke doesn't activate accelerator, it should be passed as normal character

Code reproducing issue:
```common-lisp
(defpackage #:clim.bug
  (:use #:clim-lisp #:clim)
  (:export #:display))
(in-package #:clim.bug)

(define-application-frame bug ()
  ((book :initform nil :accessor active-book))
  (:panes (display :application :display-function 'display)
          (buttons (horizontally ()
                     (make-pane :push-button)
                     (make-pane :push-button))))
  (:layouts (default
                (vertically ()
                  display
                  buttons))))

(defun display (frame pane))

(define-bug-command (foo :menu t :keystroke #\x)
    ((a string))
  "bam")

(defun gui-main ()
  (run-frame-top-level (make-application-frame 'bug)))
```

nsrahmad:
## why the difference in behaviour
 1) When invoking from menu, command was looked up in the command-table and executed. We nowhere encounter `*accelerator-gestures*` and it stays `NIL`.
 2) when `input-steam` is active, keystroke is added to `*accelerator-gestures*`.

Now in `accepting-values` (which defines its own command-table and gestures), We append `accepting-values` gestures to `*accelerator-gestures*`. In case 1, `*accelerator-gestures*` is `NIL` so existing accelerators are not active, but in case 2, existing accelerators are active and they cannot be printed.

### Solution 
Instead of appending `accepting-values` gestures to `*accelerator-gestures*`, just overwrite it. It will be restored automatically when `accepting-values` exits. and if your application requires any specific accelerator active in `accepting-values`, you can add that command to =accepting-values= manually by using `add-command-to-command-table` or you could explicitly define the command to be in the accepting-values command-table. For example:
```
(define-command (com-quit :command-table accept-values
				:keystroke (#\q :control)
				:provide-output-destination-keyword nil)
    ()
  (frame-exit *application-frame*))
```
Put the above code in `test` example above. This gesture (`Ctl-Q`) will be available in `accepting-values` and will quit the example (somewhat silly thing to do but anyway).

dkochmanski:
Thank you for the insightful analyze, I appreciate that very much.

PR #335, opened 2017-10-14
Fix-164: Do not append keystrokes to *accelerator-gestures*

nsrahmad:
Instead just overwrite them till the 'accepting-values' is active. it will be automatically restored when 'accepting-values' finishes. See #164.


dkochmanski:
Works for me, thanks! Please claim a bounty on bountysource.

PR #337, opened 2017-10-25
Improve text field gadget

### LINKED ISSUE #317: [$100] Text field pane height is too small, clipping the bottom off characters
This example shows the problem:

```
(define-application-frame text-field-frame ()
  ()
  (:pane
    (spacing (:thickness 40)
      (horizontally (:x-spacing 10)
         (make-pane 'label-pane :label "abc_def")
         (make-pane 'text-field :value "ghi_jkl" :width 250)))))

(run-frame-top-level (make-application-frame 'text-field-frame))

```
The label pane displays abc_def correctly, but the underscore character in ghi_jkl is not visible in the text field pane.

Text field height was correct in mcclim-20160929. Since then there have been changes to `box-layout-mixin/xically-allocate-space-aux*` to remove calls to `ceiling`. If I put those calls back in then this text field height problem is fixed. Could those changes be reverted?

nsrahmad:
+ I have changed button spacing because It had very low padding. (`g` and `y` will touch the button boundry. see first column of buttons in `demodemo`)
+ Block cursor is replaced with a line cursor of `3` pixel width. Tested with listener, climacs and different editor widgets.
+ Fixed #317 .

### DIFF
--- a/Core/clim-core/gadgets.lisp
+++ b/Core/clim-core/gadgets.lisp
@@ -1102,7 +1102,7 @@ and must never be nil.")
     :align-x :center
     :align-y :center
     :x-spacing 4
-    :y-spacing 2))
+    :y-spacing 4))
 
 (defmethod compose-space ((gadget push-button-pane) &key width height)
   (space-requirement+* (space-requirement+* (compose-label-space gadget)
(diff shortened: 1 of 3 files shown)


dkochmanski:
Thanks for waiting. I was busy with ellipsis code. I can confirm, that fix works for me. Regarding bigger size for buttons – I agree. Merging as is. Please claim a bounty.

PR #345, opened 2017-10-31
Fix disabled quit option in pages menu of tabdemo.

jstoddard:
The "Quit" menu option is disabled in the tabdemo.lisp example program; this commit changes the reference to nonexistent com-quit-tabdemo in the 'tabdemo-pages-menu command table to com-quit.


dkochmanski:
Looks good to me, thanks!

PR #347, opened 2017-11-01
Develop

dkochmanski:
- core improvements (design etc)
- ellipse region implementation
- clx clipping region may be arbitrary region
- we can draw ellipses which are no xy-aligned
- new demos for clim-examples with descriptions

I suggest going through commits one by one - they are cleaned up with short descriptions of changes. I have rearranged their order a few times to make them more readable and locally relatable.

### DIFF
--- a/Backends/CLX/medium.lisp
+++ b/Backends/CLX/medium.lisp
@@ -918,14 +918,58 @@ time an indexed pattern is drawn.")
             (vector-push-extend (- max-y min-y) points)))
         (xlib:draw-rectangles mirror gc points filled)))))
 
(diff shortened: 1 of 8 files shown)


dkochmanski:
new commits:
- fixups for line intersection code (it was slightly wrong for some cases)
- speed improvements for CLX backend drawing of rotated ellipses (we go down to O(n+m)
- speed improvements for ellipse clipping region (we go down to O( min(n, m))
- implementation for drawing rotated ellipsis for CLX-fb (start-end angles are distorted due to bezier approximation, bug is reported).
- fix for rectangle rotation test

dkochmanski:
do you find something to fix, or we're good to merge?

Issue #285, opened 2017-07-05
[$100] drag-test demo: dragging square to the empty position invokes the debugger


dkochmanski:
Steps to reproduce:
1. Run D&D translator
2. Drag square outside the window

Observed: debugger invoked
Expected: mouse doesn't move further, or D&D ends with failure (gets back to original position) - not sure

```
There is no applicable method for the generic function
  #<STANDARD-GENERIC-FUNCTION CLIM-INTERNALS::INVOKE-WITH-SHEET-MEDIUM-BOUND (5)>
when called with arguments
  (#<CLOSURE (LABELS #:G410 :IN CLIM-INTERNALS::INVOKE-WITH-IDENTITY-TRANSFORMATION) {7FFFF05FAB0B}>
   NIL
   #<CLIM-EXTENSIONS:VIEWPORT-PANE "(Unnamed Pane)" {10178D56B3}>).
   [Condition of type SIMPLE-ERROR]

Restarts:
 0: [RETRY] Retry calling the generic function.
 1: [ABORT] Return to application command loop
 2: [ABORT] Return to application command loop
 3: [RETRY] Retry SLIME REPL evaluation request.
 4: [*ABORT] Return to SLIME's top level.
 5: [ABORT] abort thread (#<THREAD "repl-thread" RUNNING {10030BFFA3}>)

Backtrace:
  0: ((:METHOD NO-APPLICABLE-METHOD (T)) #<STANDARD-GENERIC-FUNCTION CLIM-INTERNALS::INVOKE-WITH-SHEET-MEDIUM-BOUND (5)> #<CLOSURE (LABELS #:G410 :IN CLIM-INTERNALS::INVOKE-WITH-IDENTITY-TRANSFORMATION) {7..
  1: (SB-PCL::CALL-NO-APPLICABLE-METHOD #<STANDARD-GENERIC-FUNCTION CLIM-INTERNALS::INVOKE-WITH-SHEET-MEDIUM-BOUND (5)> (#<CLOSURE (LABELS #:G410 :IN CLIM-INTERNALS::INVOKE-WITH-IDENTITY-TRANSFORMATION) {7..
[...]

PR #349, opened 2017-11-05
Solution to  #285: drag-test demo

josrr:
Please review.

### DIFF
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -1552,6 +1552,10 @@ have a `pointer-documentation-pane' as pointer documentation,
 
 (defmethod frame-drag-and-drop-feedback
     ((frame standard-application-frame) from-presentation stream
+     initial-x initial-y x y state))
+
+(defmethod frame-drag-and-drop-feedback
+    ((frame standard-application-frame) from-presentation (stream output-stream)
      initial-x initial-y x y state)
   (with-bounding-rectangle* (fp-x1 fp-y1 fp-x2 fp-y2)
       from-presentation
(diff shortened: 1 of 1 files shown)


Core/clim-core/frames.lisp
```diff
@@ -1630,10 +1630,12 @@ have a `pointer-documentation-pane' as pointer documentation,
 		   do (return-from find-dest-translator translator))
 	     nil)
 	   (do-feedback (window x y state)
-	     (funcall feedback-fn frame from-presentation window
-			initial-x initial-y x y state))
+	     (when (typep window 'clim-stream-pane)
```
dkochmanski: it solves immediate issue for me, but are we allowed to limit d&d to only clim-stream-pane? Also, why does it break, if window is something different than clim-stream-pane?

Core/clim-core/frames.lisp
```diff
@@ -1630,10 +1630,12 @@ have a `pointer-documentation-pane' as pointer documentation,
...
+	       (funcall feedback-fn frame from-presentation window
+			initial-x initial-y x y state)))
 	   (do-hilite (presentation window state)
-	     (funcall hilite-fn frame presentation window state))
+	     (when (typep window 'clim-stream-pane)
```
dkochmanski: is this also necessary? after wrapping in when feedback-fn, things seem to work for me. why is the same thing done to do-hilite?

josrr: You are right, s not necessary; I didn't test this after changed button-release by pointer-button-release.

Core/clim-core/frames.lisp
josrr: I think is not specified; the problem is that the function frame-drag-and-drop-feedback try to draw in the stream, and this invoke INVOKE-WITH-SHEET-MEDIUM-BOUND; but there is no method defined for all classes of sheets.

dkochmanski: in that case please specialize `frame-drag-and-drop-feedback` stream argument on `output-stream class ` and add new method (above the existing one), which is not specialized on anything and is a no-op (empty body).

We'd want to specialize on `output-recording-stream`, but we have take into account also that we have `encapsulating-stream`, which isn't recording (but underlaying stream may be output-recording-stream).

dkochmanski: also please squash changes into single commit when you do

dkochmanski:
also, before you squash, please remove all indent and whitespace changes from changes

dkochmanski:
hey, I've squashed it and pushed to the repository. please claim a bounty on bountysource. Thanks for narrowing down and fixing the issue.

PR #403, opened 2017-11-12
Debugger improvements

scymtym:
* Add pointer documentation pane
* Output causing condition as presentation to make it inspectable


dkochmanski:
looks good, thanks!

PR #407, opened 2017-11-21
Table formatting

### LINKED ISSUE #405: bordered-output regression : highlight-background option is ignored
caused by commit b54eac6b
### Now
now

----

### Before
then

dkochmanski:
Fixes #405.

### DIFF
--- a/Examples/package.lisp
+++ b/Examples/package.lisp
@@ -1,4 +1,4 @@
 
 (defpackage #:clim-demo
   (:use #:clim-extensions #:clim #:clim-lisp)
-  (:export #:demodemo #:run-drawing-tests #:tabledemo #:misc-tests))
+  (:export #:demodemo #:run-demo))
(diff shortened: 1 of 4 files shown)


Examples/package.lisp
```diff
@@ -1,4 +1,4 @@
 
 (defpackage #:clim-demo
   (:use #:clim-extensions #:clim #:clim-lisp)
-  (:export #:demodemo #:run-drawing-tests #:tabledemo))
+  (:export #:demodemo #:run-drawing-tests #:tabledemo #:misc-tests))
```
nsrahmad: Idea of quick start function is fine but the exported symbol list keeps growing. How about we export `demodemo` and a single function which takes a `test-name` argument and runs the test directly?

dkochmanski: Fixed (added single function run-demo which is exported from clim-demo package).

PR #411, opened 2017-12-04
improve user experience

dkochmanski:
Scrolling improvements

- refine event class hierarchy
- make all panes with viewport scrollable
- fix bug which took whole output-history size as sheet size (even when started at negative or positive coordinates, which was resulting in bogus scrolling)
- add new scrolling test

There are more changes pending, but this part is an improvement on its own and I think it will be easier to review it now while it's still moderately small.


dkochmanski:
Hey, thank you for the review. I did test it briefly, but apparently this branch caused some regressions. I agree that we shouldn't merge it before these issues are fixed.

dkochmanski:
both bugs fixed (I hope).

nsrahmad:
Nice! 
however, Now `pointer-documentation pane` in `listener` (or `german towns` example, or `tab-layout` example) becomes too large. It remains normal in `dragndrop` example. Only visible difference is that `pointer-documentation` pane in `dragndrop` example doesn't have scrollbars.

dkochmanski:
Fixed, thanks!

PR #428, opened 2018-01-21
Develop

dkochmanski:
Makes interactor pane working without breaking bordered table demo (table always starts at the very position formatting-table is invoked in – imo a reasonable decision).


Core/clim-core/commands.lisp
```diff
@@ -847,16 +847,15 @@ examine the type of the command menu item to see if it is
...
-                                            :align-prompts t)
-                   (format ,stream
-                           "You are being prompted for arguments to ~S~%"
-                           ,command-line-name)
+               (let* ((,command-line-name (command-line-name-for-command
+                                           ,command-name
+                                           ,command-table
+                                           :errorp nil))
+                      (,label (format nil
+                                      "You are being prompted for arguments to ~S"
```
nsrahmad: Since table starts exactly where the cursor is, You may want to add a new line format directive here, Otherwise the result will look odd (To See it click on: german town example -> Get Distance)

dkochmanski: that's what I meant by terpri in the other place
```
(when label
                        (format stream label)
                        (terpri))
```
but forgot to pass `stream` to terpri (fixed in last commit).

nsrahmad: That's the second reason I hate terpri. (The first--- it got it's own homepage at http://terpri.org )

nsrahmad:
Looks good now, Merging it

PR #431, opened 2018-01-30
Ellipse

mgi:
This work fixes some ellipse issues in the render backend


dkochmanski:
works great, thank you! (I have no critical remarks regarding style whatsoever). No tests seem to break and angles are now correct.

PR #433, opened 2018-02-03
Improve CLX-fb experience by fixing distribute-event for single-mirro…

dkochmanski:
…red-sheet

Recent change in event hierarchy (introducing pointer-button-event and
pointer-scroll-event) caused some regressions in CLX-fb.

single-mirrored-sheet (like CLX-fb) requiers some juggling with sheets to
distribute events. On the other hand events are immutable, so we had to make a
new instance if we wanted to augment event sheet.

Since different pointer events have different slots (delta-x, button etc) we
need an utility to make event shallow copy (which we can modify without
violating the spec). Thanks to that scrolling started working on CLX-fb (just
like it works on normal CLX) and regression is fixed.

Extra benefit is the fact that code is simplier now (39 insertions vs 63
deletions).

How to test CLX-fb:

> (ql:quickload '(clim-examples mcclim-clx-fb))
> (let ((clim:*default-server-path* :clx-fb))
    (clim-demo:demodemo))

### DIFF
--- a/Backends/Standard/events.lisp
+++ b/Backends/Standard/events.lisp
@@ -98,44 +98,25 @@ pointer-exit and grab-pointer for non mirrored sheets"))
 	  (t	 
 	   (call-next-method)))))))
 
-
 (defmethod climi::distribute-event ((port standard-handled-event-port-mixin) (event pointer-event))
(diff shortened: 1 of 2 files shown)


gas2serra:
It's good!
Alessandro

Issue #385, opened 2017-11-10
[$100] Listener commands with "list" arguments, such as Run, cannot be executed from command history


abridgewater:
Steps to reproduce:
1. Start a fresh Listener.
2. Enter `,Run`.
3. Enter command `ls` and arguments `-l`.
4. Accept the values to run the command.
5. Hit M-p to bring up the previous command. Note that it appears as `(CLIM-LISTENER::COM-RUN "ls" ("-l"))`
6. Hit Enter to end up in the debugger, because `"-l"` isn't a valid function or macro name.

Issue #384, opened 2017-11-10
Listener command "Run" displays its output in a proprotional font


abridgewater:
Most noticeable with `ls -l`, where none of the columns line up. This should almost-certainly be a monospace font, not a proportional font.

Issue #383, opened 2017-11-10
Listener command "Run" fails to display command being run post-entry


abridgewater:
Steps to reproduce:
1. Start a fresh Listener.
2. Enter ,Run.
3. Enter `ls` as the command to run, with no arguments
4. Hit "OK", or otherwise successfully exit the ACCEPTING-VALUES dialoge.
5. Behold the complete lack of information about the command run.

PR #443, opened 2018-02-26
Improve listener

nsrahmad:
* fix for #383 , #384 , #385 
* Various UX improvements
* Replace listener:run-program with uiop:run-program


dkochmanski:
### launch program
launch program is not a good solution here, because if there is an error, we land in the debugger (instead of having error message in the debugger): try to list directory which doesn't exist

Old behavior:
obraz

(in Polish locale it means: ls: no access to Pulpit)

New behavior:
obraz

We probably want uiop:run-program here (or even better, external-program:run – but it is my personal preference, I'm ~OK with adding uiop despite disliking it).

### Command multiple arguments:

this is not strictly a regression, because it didn't work as desired before, but it got worse

1. type `,Run (Command) ls (Arguments[]) /home[NOW PRESS SPACE]`

New behavior:

command is executed right away (without waiting for next arguments), so effectively we can't pass multiple arguments to command

obraz

Previous behavior:

obraz

we can't type anything as a keyword, so it is not correct behavior either.

In Accepting Values we can't insert multiple arguments as well, because it doesn't read Arguments correctly into a list (with or without quotes, doesn't matter).

If we don't have a ticket for this issue we should fill it.

### issues

I confirm that this PR fixes all three issues reported earlier. My suggestion would be to replace launch-program  with run-program (or external-program alternative) to handle programs which print to stderr correctly (instead of being dropped into the debugger). Also it may be worth investigating why can't we correctly process list arguments in neither command prompt nor accepting values.

nsrahmad:
1) I was using `run-program` (you can see it in backtrace) but looks like internally `run-program` uses `launce-program`. I have updated `run-program`'s parameters to tune the output more to your suggestion.
I don't like UIOP either, but it's cross-platform and cross-implementation support is excellent. We could in the future replace listener's filesystem operations with UIOP and remove CL-FAD.

2) It can handle multiple arguments just fine. You need to separate them with , (comma). That's the standard `accept` way for sequences; closer to natural languages like English but feels really strange.

Apps/Listener/dev-commands.lisp
```diff
@@ -234,20 +234,29 @@
...
                                     :menu t
-				    :command-table application-commands
-				    :provide-output-destination-keyword t)
+				    :command-table application-commands)
   ((program 'string :prompt "Command")
    (args '(sequence string) :default nil :prompt "Args"))
-  (run-program program args :wait nil :output nil :input nil))
+  (if (zerop (length (car args)))
+      (uiop:launch-program program)
+      (uiop:launch-program `(,program ,@args))))
```
dkochmanski: I mean here: this is launch-program not run-program.

Core/clim/input-editing-drei.lisp
```diff
@@ -164,31 +166,62 @@ activated with GESTURE"))
...
                        (presentation-type-history accepting-type))))
     (when history
       (multiple-value-bind (object type)
           (presentation-history-next history accepting-type)
         (when type
-          (presentation-replace-input stream object type (stream-default-view stream)
-                                      :allow-other-keys t
-                                      :accept-result nil))))))
+	  (when (and (equal type 'command-or-form) (listp object))
+	    (when (alexandria:starts-with-subseq "COM-" (symbol-name `,(car object)))
```
dkochmanski: is there a reason why it is not a single when

```
(and (equal …)
     (listp …)
     (starts-with …))
```

?

Core/clim/input-editing-drei.lisp
```diff
@@ -164,31 +166,62 @@ activated with GESTURE"))
...
                        (presentation-type-history accepting-type))))
     (when history
       (multiple-value-bind (object type)
           (presentation-history-previous history accepting-type)
         (when type
-          (presentation-replace-input stream object type (stream-default-view stream)
-                                      :allow-other-keys t
-                                      :accept-result nil))))))
+	  (when (and (equal type 'command-or-form) (listp object))
+	    (when (alexandria:starts-with-subseq "COM-" (symbol-name `,(car object)))
```
dkochmanski: same question as above - when+when

Core/clim/input-editing-drei.lisp
```diff
@@ -164,31 +166,62 @@ activated with GESTURE"))
...
+	      (setf quoted-sublists (%quote-sublists object))))
+	  (when (null quoted-sublists)
+	    (setf quoted-sublists object))
+	  (let ((*print-case* :downcase))
+	    (presentation-replace-input stream quoted-sublists type
+					(stream-default-view stream)
+					:allow-other-keys t
+					:accept-result nil)))))))
 
 (defun history-yank-previous (stream input-buffer gesture numeric-argument)
```
dkochmanski: Both functions look pretty similar, could they be factored into one function with a parameter next/prev ?

Apps/Listener/dev-commands.lisp
dkochmanski: ah, I take that back, run-program in uiop doesn't cover `wait` argument.

nsrahmad: Oh, this is in com-background-run command, it is meant to run asynchronously.

Core/clim/input-editing-drei.lisp
nsrahmad: Sorry, you are right. I have updated it.

Core/clim/input-editing-drei.lisp
nsrahmad: We could, but they are Commands and have there own keyboard shortcuts.

Core/clim/input-editing-drei.lisp
```diff
@@ -164,31 +166,66 @@ activated with GESTURE"))
...
+		     (push item result)
+		     (push `(quote ,item) result))
+		 (push item result)))
+    (reverse result)))
+
+(defun %object-command-p (object type)
+  (when (and (equal type 'command-or-form)
+		     (listp object)
+		     (alexandria:starts-with-subseq "COM-" (symbol-name `,(car object))))
+    t))
```
dkochmanski: please just make a single when inside history-yank-*, having a separate function for that is an overkill

Core/clim/input-editing-drei.lisp
dkochmanski: we could make drei more aware about gesture name aware, define gesture names for yank-next and yank-prev and do if on the gesture (now it would be obscure `(equal gesture '((#\n :meta)))`, taking into account numerical argument (so we go 10 entries up for instance) wouldn't be a bad thing either.

That said I think that would be way out of scope of this PR. So let's leave this duplication be for now.

Core/clim/input-editing-drei.lisp
nsrahmad: As you wish! 
Just realized that part needs to be more intelligent, because commands are not guaranteed to start with "COM-", it can be customized by the user.

dkochmanski:
lgtm, thanks! one of the issues has bounty on it, please don't forget to claim it.

PR #445, opened 2018-02-27
accepting-values : Inherit query box background and foreground ink

### LINKED ISSUE #444: text input fields don't inherit color of pane (except behind text)
I'm setting up an app with a solarized color theme. The interactor paints as expected until you try a command that requests information:

drei-background

Is there a way to make the text input fields either inherit the color from the containing pane completely, or to keep its own colors completely, instead of this curious mix?

nsrahmad:
from the pane It is called in. Fixes #444 .


dkochmanski:
lgtm, thanks!

Issue #376, opened 2017-11-09
Listener context menu mouse-over highlighting is weird


abridgewater:
To reproduce:

1. Start a fresh CLIM-LISTENER.
2. Enter and evaluate some expression, such as (+ 3 4).
3. Right-click the entered expression to open the context menu.
4. Mouse-over the menu items, noting how they are highlighted as you do so.

The "expression as literal" menu item seems to be highlighted reasonably, but the "Describe" item is weird.

abridgewater:
Still occurs as of 2018-02-08, but after some discussion on IRC it appears that "weird" should be quantified.  First, it is rendered as comma-separated atoms, rather than as a lisp list.  Second, it involves _seven_ separate boxes when highlighted by mouseover.  Two screenshots attached (somehow not including the mouse cursor).

![screenshot at 2018-02-07 23 58 41](https://user-images.githubusercontent.com/469380/35956492-3322fbf4-0c64-11e8-8c6e-1c2acbafc93b.png)
![screenshot at 2018-02-07 23 59 08](https://user-images.githubusercontent.com/469380/35956493-3335e39a-0c64-11e8-8e80-de3e27d654a9.png)

wmannis:
I'd do a pull request for this, but it might be too blunt (the McCLIM codebase is still often a mystery to me).  But in the function `draw-standard-menu` in Core/clim-core/menu-choose.lisp, under the `:item` ecase changing this:

```common-lisp
                           (with-text-style
                               (stream (menu-item-option
                                        item :style
                                        '(:sans-serif nil nil)))
                             (with-output-as-presentation
                                 (stream
                                  item
                                  `((,name ,@params)
                                    :description ,(getf (menu-item-options item) :documentation)
                                    ,@options))
                               (funcall item-printer item stream)))))))
```
to this:
```common-lisp
                           (with-text-style
                               (stream (menu-item-option
                                        item :style
                                        '(:sans-serif nil nil)))
                             (with-output-as-presentation
                                 (stream
                                  item
                                  `((,name ,@params)
                                    :description ,(getf (menu-item-options item) :documentation)
                                    ,@options)    ;;; <----
                                  :single-box t)  ;;; <----
                               (funcall item-printer item stream)))))))
```
seems to fix the issue without introducing any new insanities I could find (but my search may not have been sufficiently comprehensive).

dkochmanski:
[...]

PR #449, opened 2018-03-10
Make lists and vectors look reasonably lispy in Listener menus

wmannis:
Addresses the second part of issue #376.

### DIFF
--- a/Apps/Listener/dev-commands.lisp
+++ b/Apps/Listener/dev-commands.lisp
@@ -162,6 +162,42 @@
           (format stream "~A bytes" object)
         (format stream "~,1F~A" (/ object (expt 1000 idx)) (nth idx suffixes))))))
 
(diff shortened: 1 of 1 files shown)


nsrahmad:
Sorry about the late reply but this can not be merge.

There is only `sequence`(presentation-type) visible on the screen (object-type could be vector or list, doesn't matter). Quoting CLIM Spec from [here](http://bauhh.dyndns.org:8000/clim-spec/23-8.html#_1284):
> The printed representation of a sequence type is the elements separated by commas. It is unspecified whether accept returns a list or a vector.

Your PR changes the way `sequence` is presented. It violates the spec. But that's not the only reason to not merge it.

A better way to solve this issue is to specify the input recieved at listener as `form` or `expression` presentation type.
screenshot_20180318_003830.
Do not attempt it yet because of bugs in presentation system. See #452. You can close this PR if you want.

PR #453, opened 2018-03-17
Fittestbits/demo fixes

fittestbits:
Change CLX/medium.lisp so that text-style-fixed-width-p checks for family :fixed in stead of :fix.

Change Examples/demodemo.lisp so that a call to apropos-list only has three arguments when using SBCL, since the third argument does not seem to be part of the common lisp spec. and is not supported by CLISP or Mezzano.

Change Extensions/render/backend/port.lisp to support gs-...-mixin. (gs = graph-state) For, graph state, medium-text-style is a list of (<family> <face> <size>) instead of a standard-text-style. Added methods text-style-to-font and text-style-size methods to support this type of medium-text-style.


dkochmanski:
lgtm, thanks!

PR #454, opened 2018-03-18
simplify mirroring class hierarchy - pass2

dkochmanski:
Second batch of cleanup (there will be the third one).

Also a new demo application useful for debugging mirrors.

### DIFF
--- a/Backends/CLX-fb/mirrored-sheets.lisp
+++ b/Backends/CLX-fb/mirrored-sheets.lisp
@@ -1,7 +1,7 @@
 (in-package :clim-clx-fb)
 
 (defclass clx-fb-mirrored-sheet-mixin (image-sheet-mixin
-				       standard-single-mirrored-sheet-mixin)
+				       standard-mirrored-sheet-mixin)
   ())
 
 (defmethod sheet-direct-xmirror ((sheet clx-fb-mirrored-sheet-mixin))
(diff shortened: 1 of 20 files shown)


nsrahmad:
Regression Alert!
menus in menu-bar are blank. menu-choose or context menu is fine.

nsrahmad:
Another problem.
co-ordinate swizzling code is broken (trying something like ,show class sub-classses t which provides lot of output blows up CLX.
Try the demo app at #150 
1) scrolling doesn't update the pane until you resize the pane.
2) around 2050 (where CLX reaches it's limits) numbers get mixed up.

nsrahmad:
Everything else is fine I guess.
Note that these regression are related to this PR only. I had tested everything against yesterday's changes and All was Fine.
BTW heirarchy-tool is cool. Is there a particular reason you choose BSD llicense? I mean everyone else (example authors) choose LLGPL to keep it consistent with McCLIM.

dkochmanski:
Nice catch. Thanks for reviewing the changes. In that case I'll just drop the third batch when finished with these issues fixed. I don't see much merit in fixing code which is going to be removed.

// menu regression was caused by overblown update-mirror-geometry which worked on some dubious assumptions.

Thanks for the praise. I have picked it because it is compatible with LGPL and I hoped nobody will ask me why did I choose LGPL :_) if it is a concern I may change it back - on the other hand I think that examples should be licensed as liberal as possible. That's what I did in ECL: whole implementaiton is LGPL-2.1+, but examples are BSD-2-Clause.

PR #456, opened 2018-03-25
[wip] unify mirrored-sheet-mixin hierarchy

dkochmanski:
I haven't spotted regressions so far. This batch collapses mirrored-sheet-mixin class hierarchy into single one (as meant by the specification, where port-* is responsible for backend-specific stuff while mirrored-sheet-mixin is part of the core). Also removes a dozen of rudimentary internal notification protocols.

Still wip because:

- dispatch-repaint in sheets.lisp calls too big region (+everywhere+) instead of uniojn  of previous and current sheet region or sheet-parent region (not sure which is better yet);
- mixed mirror sheet hierarchies are not fully supported (single-mirror and full-mirror work fine), need to fix that;
- parser for server-path should accept `:mirroring` keyword which will accept `:full`, `:single`, `:random` and `#'mirror-fn` arguments which will decide about mirroring policy.

CLXv3 is removed because CLX works fine with single-mirror now.


nsrahmad:
Good work! I only found two issues.
- [ ] Interactor Not repainting properly and slightly out of viewport
anim
- [ ] Coordinate swizzling
Scrolling does not work right if you keep moving in downward direction only. scrolling slightly upwards always syncs it properly. If I scroll all the way up and down multiple times, I get 0000 on up side and different random values at the bottom. Expected behaviour will be gettiing 4400 at the bottom every time.

You can merge when these two issues are fixed.

dkochmanski:
I see both problems, thanks for spotting them

nsrahmad:
Oops! The regression with interactor is in `master` too, so it's NOT related to this PR. I am opening an issue for it.

PR #457, opened 2018-03-31
fix Pixmap 3 drawing test by fixing args to copy-to-pixmap

slyrus:
* we need to (copy-to-pixmap stream ...) to draw on the
   stream/medium's pixmap. copy-to-pixmap takes a medium as its first
   argument, not a pixmap.

### DIFF
--- a/Examples/drawing-tests.lisp
+++ b/Examples/drawing-tests.lisp
@@ -1584,7 +1584,7 @@ clipping areas. No point should be drawn outside the blue rectangles."
 		  (draw-rectangle* m 0 0 200 200 :ink +red+)
 		  (picture01 m))))
     (copy-from-pixmap pixmap 0 0 200 200 stream 0 0)
-    (copy-to-pixmap pixmap 0 0 100 100 pixmap 100 100)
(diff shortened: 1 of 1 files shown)


Examples/drawing-tests.lisp
```diff
@@ -1584,7 +1584,7 @@ clipping areas. No point should be drawn outside the blue rectangles."
 		  (draw-rectangle* m 0 0 200 200 :ink +red+)
 		  (picture01 m))))
     (copy-from-pixmap pixmap 0 0 200 200 stream 0 0)
-    (copy-to-pixmap pixmap 0 0 100 100 pixmap 100 100)
+    (copy-to-pixmap stream 0 0 100 100 pixmap 100 100)
```
dkochmanski: What about image viewer?
Wouldn't it make sense  to add copy-to-pixmap specialization on pixmap, which calls CLIM-INTERNALS::PIXMAP-MEDIUM on it?

dkochmanski:
for discussion look into IRC Log: https://irclog.tymoon.eu/freenode/%23clim?from=1522505293, merging as is now.

PR #458, opened 2018-03-31
Remove inappropriate method on bounding-rectangle

### LINKED ISSUE #426: Regression: draw-arrow broke between ql-dist 2017-10-19 and 2018-1 ql-dist
Draw-connector no longer functions. I had to revert to the October ql-dist to make it work.

karlosz:
Fixes #426.

### DIFF
--- a/Core/clim-basic/regions.lisp
+++ b/Core/clim-basic/regions.lisp
@@ -2516,10 +2516,6 @@
 (defmethod region-contains-region-p ((a region) (b region))
   (or (eq a b)
       (region-equal +nowhere+ (region-difference b a))))
-
-(defmethod region-contains-region-p ((a standard-rectangle) (b bounding-rectangle))
-  (or (eq a b)
-      (region-equal +nowhere+ (region-difference (bounding-rectangle b) a))))
 ;;;; ===========================================================================
 
 (defmethod bounding-rectangle* ((a standard-line))


karlosz:
Fixes #426.

dkochmanski:
This function solves some cases which are unsolvable otherwise. I'm pretty sure this method is correct, so I'd suspect problem in your code.

```
(defmethod region-contains-region-p ((a standard-rectangle) (b bounding-rectangle))
  (or (eq a b)
      (region-equal +nowhere+ (region-difference (bounding-rectangle b) a))))
```
Exact meaning is as follows: if a (region being a rectangle) is bigger than the rectangle which bounds b (where b may be ellipse or other shape), then a is certainly bigger than b (because bounding rectangle is bigger or equal than the shape itself). In this case region-difference is nowhere, hence A (as a rectangle) contains b.

karlosz:
With this extra method, (region-contains-region-p #<STANDARD-BOUNDING-RECTANGLE X -120:120 Y 0:384 {1006DA1B73}> #<STANDARD-LINE 0 384 0 408>) returns NIL. However, CLIM specifies that rectangles are closed, i.e. contain their boundary, so this method should return T instead.

dkochmanski:
You are right, this method causes CLIM return inappropriate answers, hence it is wrong.

```

dkochmanski:
It will be nice to have graphical test which shows region operations on regions which are possible to pick from the list. Like: rectangle vs ellipse, region-difference, this and this parameters and it draws both regions (with composable ink) and the resulting region border.

Issue #467, opened 2018-04-09
stream-pointer-position wrong coordinates


admich:
The stream-pointer-position function returns the coordinates of the pointer in the pointer's sheet coordinate system not in the coordinates of the stream argument coordinate systems.

admich:
I try to solve the problem with this:
https://github.com/admich/McCLIM/commit/2a7af9843cce64ba480a85210f838b2974a9ea6d
It seems to work and it solves also issue #465 because with this code `stream-position-code` does not call `pointer-position` that I think is buggy (issue #466).

nsrahmad:
Hello @admich , Your changes looks reasonable (I haven't tried them yet). In order to test and merge it in McCLIM, I need you to open a Pull Request. You had contributed before with this patch-included style (e.g #293) , but anything non-trivial is much easier with [GIthub's work flow](https://guides.github.com/introduction/flow/).

Issue #464, opened 2018-04-09
Scigraph: fixnum bug


admich:
In scigraph (make-demo-frame) gives the following error:
The value
  5521/16
is not of type
  FIXNUM
   [Condition of type TYPE-ERROR]

Restarts:
 0: [CLEAR-PANE-TRY-AGAIN] Clear the output history of the pane and reattempt forceful redisplay
 1: [CLEAR-PANE] Clear the output history of the pane, but don't redisplay
 2: [SKIP-REDISPLAY] Skip this redisplay
 3: [SCIGRAPH-TOP-LEVEL] Abort to SCIGRAPH Top Level
 4: [ABORT] abort thread (#<THREAD "Frame Top Level" RUNNING {10075D1943}>)

The solution is very easy in :Apps/Scigraph/scigraph/draw.lisp is necessary change stream-height function.

(in-package :graph)

(defun stream-height (stream)
  "Height of the viewport."
  (multiple-value-bind (ignore height)
      (stream-viewport-size stream)
    (declare (ignore ignore))
    (truncate height)))   		; added truncate

admich:
I solved this in my repo with this commit:
https://github.com/admich/McCLIM/commit/d3f940c97b80ca3456eb33e6606b07e16c7c81be

Issue #465, opened 2018-04-09
Scigraph: pointer problem


admich:
Many scigraph command that are execute with menu gives the following error:

The value
  NIL
is not of type
  XLIB:WINDOW
when binding XLIB:WINDOW
   [Condition of type TYPE-ERROR]

Restarts:
 0: [TRY-AGAIN] Try executing the command COM-SLIDER-CROSSHAIRS again
 1: [ABORT] Return to application command loop
 2: [SCIGRAPH-TOP-LEVEL] Abort to SCIGRAPH Top Level
 3: [ABORT] abort thread (#<THREAD "Frame Top Level" RUNNING {10075C1CD3}>)

Backtrace:
  0: (XLIB:QUERY-POINTER NIL) [external]
      Locals:
        NIL = NIL
  1: ((:METHOD POINTER-POSITION (CLIM-CLX::CLX-BASIC-POINTER)) #<CLIM-CLX::CLX-POINTER {1001ACCB03}>) [fast-method]
      Locals:
        POINTER = #<CLIM-CLX::CLX-POINTER {1001ACCB03}>
        CLIM:SHEET = #<CLIM-CLX::CLX-MENU-PANE-DUMMY "(Unnamed Pane)" {10085D1B93}>
  2: ((:METHOD STREAM-POINTER-POSITION (STANDARD-EXTENDED-INPUT-STREAM)) #<CLIM-CLX::CLX-APPLICATION-PANE-DUMMY DISPLAY {1008B23DA3}> :POINTER NIL) [fast-method]
  3: ((:METHOD STREAM-POINTER-POSITION* (T)) #<CLIM-CLX::CLX-APPLICATION-PANE-DUMMY DISPLAY {1008B23DA3}>) [fast-method]
  4: (UV-UNDER-MOUSE #<CLIM-CLX::CLX-APPLICATION-PANE-DUMMY DISPLAY {1008B23DA3}>)
  5: ((FLET "DESTINATION-CONTINUATION0" :IN COM-SLIDER-CROSSHAIRS))
  6: (COM-SLIDER-CROSSHAIRS #<ANNOTATED-GRAPH ANNOTATED-GRAPH-1> #<CLIM-CLX::CLX-APPLICATION-PANE-DUMMY DISPLAY {1008B23DA3}> :OUTPUT-DESTINATION NIL)
[...]

PR #468, opened 2018-04-11
scigraph bug fix and new `stream-pointer-position`

admich:
The first commit fix issue #464.
The other two fix issue #467 and incidentally also #465


nsrahmad:
Thanks for your contribution, specially to Scigraph as it needed some love.

PR #469, opened 2018-04-18
Improve image -> pattern code by calling specific opticl functions.

knusbaum:
opticl isn't great at guessing file types. It appears to use the file extension in order to guess, which isn't always correct.

In cases where the user is providing a `:format` argument, we shouldn't make opticl try to guess, since the user has told us what format the file is. The current code ignores the ':format' argument for common formats that opticl is used to decode. This fixes that.

Previously, these didn't work - opticl couldn't guess because of the lack of an extension - but now they do:
```
CL-USER> (clim:make-pattern-from-bitmap-file "/home/kyle/Documents/12pbm" :format :pbm)
#<CLIM-EXTENSIONS:RGB-PATTERN {10309BD053}>                                                                                                                                                          
CL-USER> (clim:make-pattern-from-bitmap-file "/home/kyle/Documents/12pgm" :format :pgm)
#<CLIM-EXTENSIONS:RGB-PATTERN {1030A18113}>                                                                                                                                                          
CL-USER> (clim:make-pattern-from-bitmap-file "/home/kyle/Documents/12tiff" :format :tiff)
#<CLIM-EXTENSIONS:RGB-PATTERN {1030FFB3B3}> 
```

Of course, it still falls back on previous behavior for any formats not explicitly specified.
The formats I've identified are:
```
gif
jpg
pbm
pgm
png
pnm
ppm
tiff
```
Which are all the ones opticl can decode, as far as I can tell.

### DIFF
--- a/Extensions/bitmap-formats/bitmaps.lisp
+++ b/Extensions/bitmap-formats/bitmaps.lisp
@@ -11,12 +11,26 @@
   (:documentation "This condition is signaled when trying to read a
   bitmap file whose format is not supported." ))
 
-(defun opticl-read-bitmap-file (image-pathname)
-  (let* ((img (handler-case
-                  (opticl:read-image-file image-pathname)
-                (error ()
-                  (error 'unsupported-bitmap-format))))
-         (height (array-dimension img 0))
+(defmacro define-opticl-reader (name opticl-reader)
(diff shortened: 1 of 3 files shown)


dkochmanski:
lgtm, thanks!

PR #471, opened 2018-04-19
ux: text-selection: be more specific with the mixin

### LINKED ISSUE #470: Text Selection Regression
PR #456 breaks text selection drawing.
Now, instead of highlighting text, selection (with shift + click and drag) causes selected text to disappear. I've included before-and-after shots.

Before at `221acc2b`:
screenshot from 2018-04-18 23-58-26

After at `095a91af`:
screenshot from 2018-04-19 00-00-56

dkochmanski:
- previously mixin was specialized on a pane, but methods were specialized on
  extended streams, what lead to erros when trying to select on (for instance) a
  gadget - now we mix it in onlhy for standard-input-stream and
  standard-output-stream what resolves this (unreported) issue;

- in repaint protocol we have used pane-background instead of medium-background
  what lead to regression in drawing text-selection background - we fix
  that (closes #470);

- previously selection background was blue and text was white – I've changed to
  to while-yellow/red what imho looks much better - very opinionated tweak.

### DIFF
--- a/Core/clim-basic/repaint.lisp
+++ b/Core/clim-basic/repaint.lisp
@@ -203,8 +203,8 @@ want to do the same.")
            (native-sheet-region (effective-repaint-region parent sheet region)))
       (with-sheet-medium (medium parent)
 	(with-drawing-options (medium :clipping-region native-sheet-region
-				      :ink
-                                      #-jd-test(pane-background sheet)
+				      :ink (medium-background medium)
+                                      #+jd-test(pane-background sheet)
                                       #+jd-test(alexandria:random-elt
                                                 (list +darkmagenta+
                                                       +darkblue+
--- a/Core/clim-basic/stream-input.lisp
+++ b/Core/clim-basic/stream-input.lisp
@@ -76,9 +76,10 @@
 ;;; what we've got after it returns.
 
 (defclass standard-input-stream (fundamental-character-input-stream
-				 standard-sheet-input-mixin)
+                                 standard-sheet-input-mixin
+                                 cut-and-paste-mixin)
   ((unread-chars :initform nil
-		 :accessor stream-unread-chars)))
+                 :accessor stream-unread-chars)))
 
 (defmethod stream-read-char ((pane standard-input-stream))
   (if (stream-unread-chars pane)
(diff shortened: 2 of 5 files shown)


nsrahmad:
looks good. except one edge-case.

You choose light-yellow/dark-red as BG/FG. What if a user decides to use these colors as there application's BG/FG?

nsrahmad:
A possible solution would be to flip BG and FG for selection. You could use flipping ink or more easier solution below:
1) Get rid of `*marked-foreground*` and `*marked-background*`.
2) Replace [this](https://github.com/robert-strandh/McCLIM/blob/391b1cdf2a0ef632f76b48ccd87393b1c6fc5aca/Core/clim-basic/text-selection.lisp#L166-L167) with:
```common-lisp
...
((letf (((medium-foreground medium) (medium-background medium))
        ((medium-background medium) (medium-foreground medium)))
...
```
not the prettiest, but guaranteed to work.

dkochmanski:
I think that the best general approach would be having a pattern drawn over the text, so the colors are preserved (for instance code highlight), but that has to wait for rethink how copy&paste should work in McCLIM (right now it is a pile of hacks put on top of dispatch-event). I've implemented intermediate solution (light-yellow vs dark-red for default colors and color swap for other pairs).

Pattern-idea illustration for reference

obraz

nsrahmad:
Thats even better. But couldn't get too fancy on CLX backend without hardware acceleration and anti-aliasing.

nsrahmad:
Didn't work too well (CLIM-USER> is also highlighted)!
screenshot_20180419_212425
Below is the version when color is always swapped.
screenshot_20180419_213032
if you insist on that colorset, at least make the background darker. Sorry if I am being too nitpicky.

dkochmanski:
I think I do not understand the problem here: in both cases clim-user is highlighted *and* in both cases if there is color to the text it is preserved. Issue whenever clim-user should be possible to select for copy-paste is a matter of specialization of the corresponding `eos/*` methods (this pull request solves only the problem with disappearing text, it doesn't attempt to fix design of text-selection behavior).

nsrahmad:
Sorry, my mistake. Didn't realize color temperature  program is making the background impossible to see.

PR #473, opened 2018-04-20
Fittestbits/demo fixes2 - three small bug fixes

fittestbits:
Bugs found running demodemo: add new page to tab-layout caused error; truetype font cache failed and caused error; and, text width calculation didn't match text drawing code.

More details in commit log messages.


dkochmanski:
thanks!

PR #479, opened 2018-05-02
invoke-w-r-f-g : Fix position of output record when :height is given

### LINKED ISSUE #435: [$100] with-room-for-graphics does not obey height for graphics not starting at 0,0
See this example from the listener. If we don't draw point at 0,0, then room-for-graphics start at height 100.

obraz

nsrahmad:
Fixes #435.

### DIFF
--- a/package.lisp
+++ b/package.lisp
@@ -1874,7 +1874,12 @@
 
 (defpackage :clim-extensions
   (:use)
+  (:nicknames :clime)
   (:export
+   ;; sheets
+   #:always-repaint-background-mixin
+   #:never-repaint-background-mixin
+
    #:line-style-effective-thickness
    #:medium-miter-limit
    #:raised-pane #:raising


Core/clim-basic/recording.lisp
```diff
@@ -2141,7 +2141,7 @@ according to the flags RECORD and DRAW."
                 (setf (output-record-position record)
                       (values (max cx (+ cx x1))
                               (if height
-                                  (max cy (+ cy (- height (- y2 y1))))
+                                  (max cy (+ cy (- height (+ (abs y1)))))
```
dkochmanski: works for me (nice catch). Is there a reason why there is `(+ (abs y1))` and not simply `(abs y1)`? If not, please remove the unnecessary function call. After that you may merge it right away and claim a bounty. Thanks!

PR #483, opened 2018-05-15
Fix typo in manual.

jl2:


dkochmanski:
Thanks! (sorry for late reply)

PR #485, opened 2018-05-29
A few minor fixes

froggey:
The inspector was printing some inspected values to *standard-output* instead of the pane.
The method-browser demo has been made more robust.
Additional type declarations have been added to the opticl image function to improve performance.


dkochmanski:
Looks good, thank you!

PR #488, opened 2018-06-10
Remove the inconsistency between :FIX and :FIXED

lokedhs:
According to the spec, the keyword :FIX should be used to indicate a
fixed-width font. In various places in the code, the additional
keyword :FIXED was also allowed, and spacial cases were implemented to
transform that to :FIX.

However, this was not consistent, and specifically the implementation
of TEXT-STYLE-FIXED-WIDTH-P returned true only for the case where the
font was set to :FIXED.

This fix removes any reference to :FIXED, in order to make the API
consistent. One of the examples in clim-examples was also changed to
use :FIX instead of :FIXED.

### DIFF
--- a/Backends/CLX/fonts.lisp
+++ b/Backends/CLX/fonts.lisp
@@ -99,9 +99,6 @@
                      size   (or size :normal)
                      size   (round (getf clim-clx::*clx-text-sizes* size size)))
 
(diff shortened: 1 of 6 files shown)


dkochmanski:
thanks!

PR #492, opened 2018-06-19
Update functions.lisp

chuchana:
Added OS X library name.


dkochmanski:
thanks. next time you may put both changes in one commit / one merge request.

chuchana:
Sure, I'll try. I thought of that when I did the this one, but I didn't know how to add it to the first PR.

dkochmanski:
if you push to the same branch as the active PR is made from, then PR is updated (either with a new commit or with amended one)

PR #497, opened 2018-06-28
Support pasting UTF-8 strings

lokedhs:
When shift-middle-clicking, request a UTF-8 string instead of a plain string from the owner of the selection.

### DIFF
--- a/Backends/CLX/text-selection.lisp
+++ b/Backends/CLX/text-selection.lisp
@@ -97,20 +97,23 @@
   (setf (selection-timestamp port) nil))
 
 (defmethod request-selection ((port clx-text-selection-port-mixin) requestor time)
-  (xlib:convert-selection :primary :STRING requestor :bounce time))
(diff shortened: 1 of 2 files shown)


lokedhs:
This fix simply changes the type read from the SelectionNotify event from STRING to UTF8_STRING.

There is an outstanding question as to whether this will prevent copying from an old application which does not support Unicode. To properly handle such case, one would need to send a TARGETS event first, wait for the response that indicates which types are available to verify that UTF8_STRING is supported.

However, I have tested with one of the oldest X applications available, xedit, which is so old it uses the Athena widget set. It also does not support Unicode, as trying to paste any Unicode content into it will result in garbage. Even this application can properly paste into McCLIM with this fix, so I believe that UTF8_STRING is automatically converted to STRING, if it is not directly supported by the application.

nsrahmad:
This is also part of #498. So no need to merge it separately.

Issue #500, opened 2018-07-12
Ellipse regions broken for non-hairy transforms


deterministic-arts:
I am getting the following error in a little toy program

```
  0: ((:METHOD SLOT-MISSING (T T T T)) #<unused argument> #<CLIM-INTERNALS::STANDARD-TRANSLATION :DX 3 :DY 3> CLIM-INTERNALS::MXX SLOT-VALUE NIL) [fast-method]
      Locals:
        SB-KERNEL:INSTANCE = #<CLIM-INTERNALS::STANDARD-TRANSLATION :DX 3 :DY 3>
        SB-PCL::NEW-VALUE = NIL
        SB-PCL::OPERATION = SLOT-VALUE
        SB-PCL::SLOT-NAME = CLIM-INTERNALS::MXX
  1: ((LAMBDA (SB-PCL::OBJECT) :IN SB-PCL::SLOT-MISSING-INFO) #<CLIM-INTERNALS::STANDARD-TRANSLATION :DX 3 :DY 3>)
  2: (CLIM-INTERNALS::%ELLIPSE-SIMPLIFIED-REPRESENTATION/RADIUS #<unavailable argument> #<unavailable argument>)
  3: (CLIM-INTERNALS::ELLIPSE-SIMPLIFIED-REPRESENTATION #<STANDARD-ELLIPSE [0.0d0 360.0d0] #<STANDARD-TRANSLATION :DX 3 :DY 3>>)
  4: (CLIM-INTERNALS::ELLIPSE-BOUNDING-RECTANGLE #<STANDARD-ELLIPSE [0.0d0 360.0d0] #<STANDARD-TRANSLATION :DX 3 :DY 3>>)
  5: ((:METHOD INITIALIZE-INSTANCE :AFTER (CLIM-INTERNALS::DRAW-ELLIPSE-OUTPUT-RECORD)) #<CLIM-INTERNALS::DRAW-ELLIPSE-OUTPUT-RECORD X 0.0d0:0.0d0 Y 0.0d0:0.0d0 {10065FFF73}>) [fast-method]
  6: ((LAMBDA (SB-PCL::|.P0.| SB-PCL::|.P1.| SB-PCL::|.P2.| SB-PCL::|.P3.| SB-PCL::|.P4.| SB-PCL::|.P5.| SB-PCL::|.P6.| SB-PCL::|.P7.| SB-PCL::|.P8.| SB-PCL::|.P9.|)) #<unavailable argument> #<unavailable ..
[...]

PR #503, opened 2018-07-12
#500: Don't directly access TRANSFORMATION slots

deterministic-arts:
Not all subclasses of TRANSFORMATION offer the MXX, MXY, etc. slots.
Use the GET-TRANSFORMATION function to obtain these values instead in
%ELLIPSE-SIMPLIFIED-REPRESENTATION/RADIUS.


nsrahmad:
Good catch, `get-transformation` is a generic function. when called with `translation` or `identity` transformations it fills the remainig components with appropriate values.

PR #507, opened 2018-07-16
add image-ops mentioned in document

zqy767:
I am not sure about the meaning of translate-image funtion so I don't wirte that.
For I use the library opticl a lot ,i write those function in opticl module.

### DIFF
--- a/Extensions/render/mcclim-render.asd
+++ b/Extensions/render/mcclim-render.asd
@@ -52,7 +52,8 @@
   :components ((:module "opticl"
                         :serial t
                         :components
-                        ((:file "opticl-image")))))
+                        ((:file "opticl-image")
+                         (:file "opticl-image-utilities")))))
 
 (defsystem #:mcclim-render/backend
     :depends-on (#:mcclim-render/cl-vectors)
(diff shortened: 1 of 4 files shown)


nsrahmad:
There are reasons why these image-ops are not needed!

1) mcclim-render images have accessor `image-pixels` which can be used to manipulate them directly with opticl (for example blur or flip etc).

2) For scale, rotate and translate, CLIM already defines the transformation API (though it's not implemented for images yet). There is no need for another API e.g. instead of using `(draw-image* (rotate-image image) ...)` you will use `(clim:with-rotation (angle) (draw-image* ...))` or provide the `:transformation` argument to `draw-image*` function. That also explains why you got confused about `translate-image`. `translate-image` didn't modifies the image pixels. It modifies the location where the image is drawn, which only makes sense in `draw-image*` function. 

So, I need you to modify the PR. 

`git reset --hard` these two commits, Remove the line 36 and line 187-198 from the file https://github.com/McCLIM/McCLIM/blob/master/Documentation/Manual/Texinfo/extensions.txi in manual and commit that.

PR #508, opened 2018-07-23
Render scroll

lokedhs:
This pull request implements several improvements.

### Fix end-of-page behaviour when creating output records

The issue here is that whenever `TERPRI` is a called, the enclosing pane is extended to accommodate the next line. However, previously this was done regardless of whether or not the output record was actually drawing or not. The result was that if you created a display function that creates an output record that prints several lines of text, but never actually add this output record to the stream, the pane would still expand.

This fix only expands the size of the pane when the text is output to the stream.

### Output pixmaps using xrender

This improves the performance of pixmap output. It also prepares for supporting image transforms.

### Optimise scrolling

Previously, whenever a pane was scrolled, the entire pane was repainted. For complex output records (in particular, very large graphs consisting of many small line segments) took long enough to draw that it created lots of flickering. Another case was when scrolling panes containing a lot of text while using the Freetype renderer (which is significantly slower than the default Truetype renderer).

This fix uses xrender to copy the overlapping area, and then request a repaint of only the part that was scrolled in. This completely eliminates all flicker and choppyness, since repainting only a few lines of graphics is much faster than repainting the entire screen.

On slower Lisp implementations, this takes scrolling performance from being almost useless to perfectly usable.

### DIFF
--- a/Backends/CLX/frame-manager.lisp
+++ b/Backends/CLX/frame-manager.lisp
@@ -125,14 +125,12 @@
   concrete-pane-class)
 
 (defmethod make-pane-1 ((fm clx-frame-manager) (frame application-frame) type &rest args)
-  (let* ((cp (find-concrete-pane-class type))
-         (p (maybe-mirroring fm cp)))
-    (apply #'make-instance
-	   p
-	   :frame frame
-	   :manager fm
-	   :port (port frame)
-	   args)))
+  (apply #'make-instance
+	 (maybe-mirroring fm (find-concrete-pane-class type))
+	 :frame frame
+	 :manager fm
+	 :port (port frame)
+	 args))
 
 
 (defmethod adopt-frame :before ((fm clx-frame-manager) (frame menu-frame))
--- a/Backends/CLX/frame-manager.lisp
+++ b/Backends/CLX/frame-manager.lisp
@@ -98,16 +98,11 @@
       (find-first-defined-class (find-symbols (generate-clx-pane-specs type)))
       type))
 
-(defclass clx-pane-mixin ()
-  ()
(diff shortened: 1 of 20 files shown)


lokedhs:
I have already experimented with image transforms using xrender, and it works pretty well. I believe that the additional code needed to add high performance transformations will be quite small.

nsrahmad:
Hello @lokedhs ,
I have not looked at the code yet, found two issues while testing:
- [ ] coordinate-swizzling is broken (demodemo->Coordinate swizzling).
- [ ] Horizontal scrolling still curropts the output (demodemo->Scroll Test 2).

lokedhs:
@nsrahmad thank you for having a look. When analysing the scrolling issue, I realised that I have bad a mistake when designing this feature in the first place. When scrolling a pane that is partially covered by another window, you end up copying that content instead of the actual content of the window.

For this to work, I need to either identify the situations where the window is partially covered, or do all drawing to a pixmap which is subsequently copied to the window. This work will have to be deferred until I get started on double buffering (which is something I have planned for a while).

Until then, I have disabled the scrolling optimisation a of 70904c9b36db0d58a88f400b4684615e8e0e4bdf

lokedhs:
The coordinate swizzling problem turned out to be a problem in CLX. It's fixed with the follwoing pull request: https://github.com/sharplispers/clx/pull/103

lokedhs:
The coordinate swizzling still have a problem? I am unable to reproduce.

nsrahmad:
No problems anymore, Merging it.

nsrahmad:
There are some style issues like using `let*` when it's not needed, Using `cond` when there are only 2 clauses etc. but I guess it doesn't matter much if code works.

lokedhs:
I'll fix that shortly. Thanks.

PR #509, opened 2018-07-30
CLX implementation of image transforms

lokedhs:
This fix implements support for transformations on calls to `DRAW-PATTERN*` for CLX mediums.

Using the groundwork provided by this fix, it should be reasonably easy to implement support for this for other mediums as well, by calling out to an external library to do image transforms.


lokedhs:
I also implemented a new demo in the Examples directory that demonstrates image transforms.

lokedhs:
After rotating the image and subsequently scrolling the pane, there is a visual problem. What seems to happen is that a second copy of the image is displayed in the wrong place.

I'm looking into the reason for why this is the case (likely caused by the medium transformation being applied twice (or not at all)). However, the fact that the image is drawn twice points to a greater problem, one that is not directly related to this patch.

nsrahmad:
Nice work!
I was about to report the scrolling behaviour. Otherwise it seems working quite well.

Examples/image-transform-demo.lisp
```diff
@@ -0,0 +1,101 @@
...
+(in-package :image-transform-demo)
+
+(clim:define-application-frame image-transform-demo ()
+  ((image :initarg :image
+          :accessor image-transform-demo/image))
+  (:panes (image-demo :application
+                      :display-function 'display-image-demo)
+          (rot-slider :slider
+                      :value 0
+                      :show-value t
```
nsrahmad: `show-value` is a predicate so no wonder its actually `:show-value-p`.

nsrahmad: I catched it when it didn't show values!

Examples/image-transform-demo.lisp
```diff
@@ -0,0 +1,101 @@
...
+  (:use :cl)
+  (:export :image-transform-demo))
+
+(in-package :image-transform-demo)
+
+(clim:define-application-frame image-transform-demo ()
+  ((image :initarg :image
+          :accessor image-transform-demo/image))
+  (:panes (image-demo :application
+                      :display-function 'display-image-demo)
```
nsrahmad: Adding `:scroll-bars :both` will improve navigating image a bit.

Examples/image-transform-demo.lisp
```diff
@@ -0,0 +1,101 @@
...
+                         :max-value 1
+                         :value-changed-callback #'slider-updated-callback)
+          (y-skew-slider :slider
+                         :value 0
+                         :show-value t
+                         :orientation :horizontal
+                         :min-value -1
+                         :max-value 1
+                         :value-changed-callback #'slider-updated-callback))
+  (:layouts (default (clim:vertically (:width 1000 :height 900)
```
nsrahmad: Please avoid using explicit pixel values for `:height` and `:width`. (my laptops display height is 768). You can try my modified layout:
```common-lisp
(:layouts (default (clim:vertically ()
		       (14/20 (clim:labelling (:label "Test Image"
					       :align-x :center
					       :label-alignment :top)
				image-demo))
		       (clim:horizontally ()
			 (clim:vertically ()
			   (1/20 (clim:labelling (:label "Rotate")
				   rot-slider))
			   (1/20 (clim:labelling (:label "Translate X")
				   x-slider)))
			 (clim:vertically ()
			   (1/20 (clim:labelling (:label "Translate Y")
				   y-slider))
			   (1/20 (clim:labelling (:label "Scale")
				   scale-slider)))
			 (clim:vertically ()
			   (1/20 (clim:labelling (:label "Skew X")
				   x-skew-slider))
			   (1/20 (clim:labelling (:label "Skew Y")
				   y-skew-slider)))))))
```

Core/clim-basic/recording.lisp
```diff
@@ -1515,37 +1515,44 @@ were added."
...
 
-;;; The Spec says that "transformation only affects the position at
-;;; which the pattern is drawn, not the pattern itself"
-(def-grecording draw-pattern (() pattern x y) ()
-  (let ((width (pattern-width pattern))
-        (height (pattern-height pattern))
-        (transform (medium-transformation medium)))
-    (setf (values x y) (transform-position transform x y))
-    (values x y (+ x width) (+ y height))))
+(defclass drawing-transform-mixin ()
```
nsrahmad: since it has transformed X and Y values, better name would be `position-transform-mixin`.

Core/clim-basic/recording.lisp
```diff
@@ -1515,37 +1515,44 @@ were added."
...
+  ((transformed-dx :initform 0
+                   :accessor drawing-transform-mixin-transformed-dx)
```
nsrahmad: accessor name `transformed-dx` and `transformed-dy` should be enough.

lokedhs:
I have figured out what was going on. There are two different output records generated by a call to `DRAW-PATTERN*`. The real one, which is defined in `clim-basic/recording.lisp`, and a second one in in `image/recording.lisp`. This happens because the output record is generated by an `:AROUND` around method, before the actual method is called in `Extensions/image/medium.lisp`. This method ends up calling `MEDIUM-DRAW-IMAGE-DESIGN*` which creates another output record.

Because there was a bug in my handling of the transformations in the latter function, this resulted in the two images being drawn differently, revealing what was happening.

I have worked around the issue by explicitly disabling recording when calling `MEDIUM-DRAW-IMAGE-DESIGN*` from `MEDIUM-DRAW-PATTERN*`. I don't know if this is the best solution, and would appreciate any suggestions.

This fix also significantly improved the performance of image drawing. Thanks to this, the transform demo can now repaint the pane as you are dragging..

lokedhs:
I have renamed the mixin as @nsrahmad suggested.

lokedhs:
I merged this myself after fixing the issues @nsrahmad pointed out. Please let me know if there are any other issues, in which case I'll fix them immediately since it's already in master.

nsrahmad:
Don't worry about it. No other problems I can find. only thing remaining was to merge it.

PR #510, opened 2018-07-31
Fix repaint order in incremental-redisplay

lokedhs:
Incremental-redisplay works by calling compute-difference-set which
collects a list of drawing operations which are to be performed. It
does this by pushing the output records to a list. This results in the
order of the output records being reversed, which causes them to be
drawn in the wrong order. This behaviour can be seen when adding
two graphics operations to an updating-output form which draws on top
of each other. After redisplay, the output records will have been
drawn in reverse order.

This fix simply calls REVERSE on the list before returning it.

### DIFF
--- a/Core/clim-core/incremental-redisplay.lisp
+++ b/Core/clim-core/incremental-redisplay.lisp
@@ -837,7 +837,7 @@ in an equalp hash table"))
          ;; moves
          nil
          ;; draws
-         come
+         (reverse come)
          ;; erase overlapping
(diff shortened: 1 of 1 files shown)


lokedhs:
The easiest way to see the effect of this bug, add an extra draw operation to the image transform demo that paints something on top of the graphics. When rotating the view, the graphics will disappear.

dkochmanski:
lgtm, thanks!

PR #512, opened 2018-08-01
Transform fix

lokedhs:
This pull request contains three fixes:

**Fix 1**

When applying the CLIM transformation matrix when loading characters using Freetype, the rotation direction is wrong. The previous solution was to transpose the matrix. This works for rotation, but ends up applying skew transformations along the wrong axis.

This fix solves the problem in the correct way by negating the xy and yx values.

**Fix 2**

The skew transformation matrix in the image-transform demo was computed as the product of the x-skew and y-skew vectors. This fix uses a single x and y skew matrix.

**Fix 3**

Some text was added to the image-transform demo in order to be able to verify that text transform works.

### DIFF
--- a/Examples/image-transform-demo.lisp
+++ b/Examples/image-transform-demo.lisp
@@ -116,9 +116,7 @@
                   (make-skew-transformation x-skew y-skew))))))
       (clim:with-drawing-options (stream :transformation tr)
         (clim:draw-pattern* stream image 0 0)
-        (clim:draw-rectangle* stream 0 0 (clim:pattern-width image) (clim:pattern-height image) :filled nil :ink clim:+blue+)
-        (clim:with-text-size (stream 60)
-          (clim:draw-text* stream "Foo abcdefgh" 100 100))))))
+        (clim:draw-rectangle* stream 0 0 (clim:pattern-width image) (clim:pattern-height image) :filled nil :ink clim:+blue+)))))
 
 (defun image-transform-demo ()
   (let ((frame (clim:make-application-frame 'image-transform-demo)))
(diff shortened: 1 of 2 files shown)


nsrahmad:
Text messed up the output in Truetype renderer. I suggest adding `#+MCCLIM-FFI-FREETYPE before `draw-text` so it's displayed only when freetype is used.

lokedhs:
What do you mean by messed up? If you're talking about the blue lines that gets drawn all over when skewing, this is unrelated to both image transforms and the text output, but rather is caused by a bug in incremental-redisplay where it doesn't properly clear the previous output records.

The fact that the text is not transformed in with the Truetype renderer is expected. Only Freetype and Postscript can do transformed text at the moment. If that is what you meant, sure, we can remove the text altogether (perhaps create a separate text transformation demo?).

nsrahmad:
I didn't mean to remove the text entirely. if you do something like:
```lisp
#+MCCLIM-FFI-FREETYPE
(clim:with-text-size (stream 60)
          (clim:draw-text* stream "Foo abcdefgh" 100 100))
```
Text will be there when Freetype is used, But not when Truetype is used.

lokedhs:
I figured that removing it altogether would be more consistent. Otherwise people running the demo may be confused as to why there is text sometimes, but not always.

If you prefer to keep it, then I'm happy to put it back.

dkochmanski:
I've asked loke to re-add text in case of a truetype renderer - we want to support transformations in it too, so this demo reveals TTFR defect.

nsrahmad:
>reveals TTFR defect.

Yes, it does, but it ruins the demo (See gif above). Until text transformations are supported, There will be users checking out `Image Transform` demo, why they need to know Truetype does not support text transformations?

lokedhs:
How about adding checkboxes to allow the users to choose what should be drawn in the pane? Then the demo can be renamed to Transformation Demo instead.

PR #515, opened 2018-08-06
Fix the :CONTENTS initialiser for SHEET-MULTIPLE-CHILD-MIXIN

lokedhs:
Previously, the CHILDREN slot has an initarg called :CHILDREN. This
initarg was not usable since none of the logic implemented by
SHEET-ADOPT-CHILD was performed.

This fix changes this so that SHEET-ADOPT-CHILD is called once for
each child. It also renames the keyword argument to :CONTENTS in order
to be consistent with other panes.

### DIFF
--- a/Core/clim-basic/sheets.lisp
+++ b/Core/clim-basic/sheets.lisp
@@ -619,7 +619,11 @@
 ;;; sheet multiple child mixin
 
 (defclass sheet-multiple-child-mixin ()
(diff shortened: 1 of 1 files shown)


dkochmanski:
lgtm, thanks!

dkochmanski:
For discussion around these changes see https://irclog.tymoon.eu/freenode/%23clim?around=1533541680#1533541680

PR #517, opened 2018-08-06
filtered button-release for pointer-wheel events

ormf:
X generates a button-press and a button-release event on each scroll-wheel change. The code assumes that only one scroll-wheel event should get generated on scroll wheel movement and only generates the event on button-press.

### DIFF
--- a/Backends/CLX/input.lisp
+++ b/Backends/CLX/input.lisp
@@ -143,8 +143,10 @@
 	((:button-press :button-release)
 	 (let ((modifier-state (clim-xcommon:x-event-state-modifiers *clx-port* state))
                (button (decode-x-button-code code)))
-           (if (member button '(#.+pointer-wheel-up+ #.+pointer-wheel-down+
-                                #.+pointer-wheel-left+ #.+pointer-wheel-right+))
+           (if (and
+                (member button '(#.+pointer-wheel-up+ #.+pointer-wheel-down+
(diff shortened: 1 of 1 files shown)


Backends/CLX/input.lisp
```diff
@@ -143,8 +143,10 @@
 	((:button-press :button-release)
 	 (let ((modifier-state (clim-xcommon:x-event-state-modifiers *clx-port* state))
                (button (decode-x-button-code code)))
-           (if (member button '(#.+pointer-wheel-up+ #.+pointer-wheel-down+
-                                #.+pointer-wheel-left+ #.+pointer-wheel-right+))
+           (if (and
+                (member button '(#.+pointer-wheel-up+ #.+pointer-wheel-down+
+                                 #.+pointer-wheel-left+ #.+pointer-wheel-right+))
+                (eq event-key :button-press))
                (make-instance 'climi::pointer-scroll-event
```
dkochmanski: Nice catch! thanks. Could you please change order of test and indent? `eq` should go first and it should be in the same line as `IF`. Also please put `#.+…` in separate lines.

```
(if (and (eq event-key …)
         (member '(#.+a+
                   #.+b+
                   #.+c+ …))
```

dkochmanski: thanks. no problem. Regarding PR's, if you force-push (or even simply push) to the branch PR is opened from, it will update itself automatically.

PR #519, opened 2018-08-10
Remove old code that was part of the optimised scroll

lokedhs:
The optimised scroll implementation implemented scrolling by copying
the content of the window instead of repaining it. This worked
sometimes, but was problematic when a window was partially visible.

As it turns out, the proper way to do this is to always draw the
window content to a backing pixmap so that the previous content is
always available. If this is done, not much of the old code will be
useful anyway, so to clean things up this commit removes all of the
unused code.

### DIFF
--- a/Core/clim-core/panes.lisp
+++ b/Core/clim-core/panes.lisp
@@ -2666,21 +2666,12 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
   (when (stream-drawing-p stream)
     (change-stream-space-requirements stream :height new-height)
     (unless (eq :allow (stream-end-of-page-action stream))
-      ;; At this point, we might be in the middle of drawing and
-      ;; recording an output record. We can't call SCROLL-EXTENT
-      ;; directly here since it might want to repaint a portion of the
-      ;; pane that has been revealed. However, since the output
-      ;; records won't be added to the stream until later, the repaint
-      ;; will not draw the correct thing.
-      ;;
-      ;; The solution is to queue an event that will trigger the scoll
-      ;; later, at which point the output records are properly setup.
       (scroll-extent stream
-                   0
-                   (max 0 (- new-height
-                             (bounding-rectangle-height
-                              (or (pane-viewport stream)
-                                  stream))))))))
+                     0
+                     (max 0 (- new-height
+                               (bounding-rectangle-height
+                                (or (pane-viewport stream)
+                                    stream))))))))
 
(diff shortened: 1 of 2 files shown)


dkochmanski:
- did you test it with sheets which are larger than maximum x11 mirror?
- how does it behave with clx-fb?

Core/clim-core/panes.lisp
```diff
@@ -2690,11 +2675,12 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
...
-                   (make-instance 'scroll-extent-event
-                                  :sheet stream
-                                  :stream stream
-                                  :new-height new-height)))))
+      (scroll-extent stream
+                   0
+                   (max 0 (- new-height
+                             (bounding-rectangle-height
+                              (or (pane-viewport stream)
+                                  stream))))))))
```
dkochmanski: according to comment above this was queued to prevent race when drawing. how does making this operation immediate prevent the problem? also, if it does, then the comment above is not adequate anymore, so in such situation it should be removed too.

lokedhs: I made that change (and wrote that comment) as part of that work. Rolling this back means that we're back where we started.

The issue that that particular code fixes is real. It ensures that any repainting that happens as part of the scroll will draw the correct content. However, without the optimised scroll, this never happens. 

Why remove it then? Well, the current implementation is incomplete when there are more than one scroll between repaints. If the second scroll operation looks at the position, it will see the old coordinate (since the actual scroll hasn't happened yet) and the resulting scroll position will be incorrect. It's rare, but it can happen in an interactor pane.

This can be solved by adding more complexity, but I figures since nothing needs this as the moment its better to roll back to what worked before I started working on the scroll optimisation.

lokedhs:
With regards to the first comment, with the exception of the code that I commented on above, this change removes code that was never run. None of these functions were actually called and were part of the failed attempt to implement the optimised scrolling. They were included as part of the merge of the support for image transforms, since it shared some code. The optimised scroll function shouldn't really have been included in the first place.

Core/clim-core/panes.lisp
dkochmanski: so this comment should be modified, because it describe a measurement which is no longer there

lokedhs: Indeed. It should have been removed along with the code change. That was a mistake. Thanks for pointing it out. I have removed it as of a04e9c2ab708f53971b97cbecb3da2617ef05e34

PR #521, opened 2018-08-13
Small changes

dkochmanski:
These changes were extracted from my current medium work and are unrelated to it (still needed and being improvement).

- compose-x-with-y transformation functions had arguments reversed
- small cleanup for scroll methods and initargs (no distracting unused code)
- specialization for mediums to be able draw on non-stream sheet inside with-room-for-graphics

I hope to make a follow-up pull request wrt invoke-with-room-for-graphics, but this will be bigger task and I don't want to get distracted from what I'm doing now.

### DIFF
--- a/Core/clim-basic/recording.lisp
+++ b/Core/clim-basic/recording.lisp
@@ -2119,8 +2119,12 @@ according to the flags RECORD and DRAW."
   (when (stream-drawing-p stream)
     (call-next-method)))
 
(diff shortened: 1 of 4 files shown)


nsrahmad:
for an example of Genera's `with-room-for-graphics` in Dynamic Windows, Look at page 11 and 12 of http://www.bitsavers.org/pdf/symbolics/software/genera_8/Program_Development_Tutorial.pdf 

Also, I agree with your interpretation of the `height` argument.

PR #524, opened 2018-08-20
fix PDF text transformation

slyrus:
Improves transformed text drawing in the PDF backend.


nsrahmad:
in `07) Text Scale`, I can see a lonely `Chao` at the top. which probably means you are getting transformation wrong and output is rendered outside the page.

PR #527, opened 2018-08-22
transform text in PDF medium-draw-text*

slyrus:
properly (hopefully) transform text in medium-draw-text* in PDF backend.


nsrahmad:
works as expected, thanks.

PR #528, opened 2018-08-22
scigraph: use draw-string instead of draw-vertical-text

slyrus:
* this allows for properly rotated text

 * adjust text angle and position for proper vertical
   positioning/orientation.

### DIFF
--- a/Apps/Scigraph/scigraph/annotated-graph.lisp
+++ b/Apps/Scigraph/scigraph/annotated-graph.lisp
@@ -394,14 +394,14 @@ older classes in favor of cleaned up newer ones.  Someday, ...
 	(let* ((annotation (make-border-annotation
 			     self STREAM y-label :left
 			     ull vll
-			     'y-label (/ pi 2) nil))
+			     'y-label #.(/ pi -2) nil))
 	       (height nil))
 	  (setf (style annotation) (y-label-text-style self stream))
 	  (setq height (* (length y-label) (stream-line-height stream)))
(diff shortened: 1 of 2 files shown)


Apps/Scigraph/scigraph/annotated-graph.lisp
```diff
@@ -394,14 +394,14 @@ older classes in favor of cleaned up newer ones.  Someday, ...
...
 			     self STREAM y-label :left
 			     ull vll
-			     'y-label (/ pi 2) nil))
+			     'y-label #.(/ pi -2) nil))
 	       (height nil))
 	  (setf (style annotation) (y-label-text-style self stream))
 	  (setq height (* (length y-label) (stream-line-height stream)))
 	  (set-uv-position annotation
 			   (- ull (* (stream-character-width stream)
 				     (+ y-digits 2)))
```
nsrahmad: I think position becomes much more pleasant if above line is `(- y-digits 2)))`

slyrus: Hmm.. This is interesting. In the case of the PDF, you're right, but I'm not sure about for, e.g., CLX backend. Could be a PDF backend bug. But, either way this seems to be independent of this change. Can we commit this and open a new one to resolve the issue of the uv-position?

nsrahmad: CLX doesn't support rotated text yet, so it's understandable it gets uv-position wrong.

PR #533, opened 2018-08-30
Traffic light demo fixes

slyrus:


Examples/traffic-lights.lisp
```diff
@@ -49,41 +49,44 @@
...
+    (cond ((string= label "Y")
            (traffic-pause 2)
-           (simulate-action (find-pane-named *application-frame* 'red)))
+           (simulate-action (find-pane-named *application-frame* 'red))
+           (repaint-sheet pane +everywhere+))
 	  ((string= label "G")
            (traffic-pause 3)
-           (simulate-action (find-pane-named *application-frame* 'yellow)))
+           (simulate-action (find-pane-named *application-frame* 'yellow))
+           (repaint-sheet pane +everywhere+))
```
dkochmanski: repaint-sheet should be called in simulate action not here

dkochmanski: since repaint sheet is called in gadget callback, do we need it here anyway?

slyrus: 1) it seems like more trouble teaching simulate-action about the pane it needs to redraw then it does just to have the light-pane's handle event method do the repainting.

2) Unfortunately, yes, we do. Perhaps the callback isn't called after setf gadget-value?

dkochmanski:
lgtm, thanks!

PR #534, opened 2018-08-30
add light-color slot to light-pane

slyrus:
* and use it instead of the climi::gadget-current-color etc... stuff.

### DIFF
--- a/Examples/traffic-lights.lisp
+++ b/Examples/traffic-lights.lisp
@@ -30,13 +30,13 @@
 (in-package :clim-demo)
 
 ;;; example gadget definition
-(defclass light-pane (basic-gadget) ())
+(defclass light-pane (basic-gadget)
+  ((light-color :initform +green+ :accessor light-color :initarg :light-color)))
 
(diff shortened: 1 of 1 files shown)


Examples/traffic-lights.lisp
```diff
@@ -72,24 +72,21 @@
...
   (repaint-all-sheets))
 
 (defun callback-yellow (gadget value)
-  (declare (ignore gadget))
-  (when value
-    (setf (clim-internals::gadget-current-color (slot-value *application-frame* 'light))
-	  (clim-internals::gadget-highlighted-color (slot-value *application-frame* 'light))))
+  (let ((pane (find-pane-named *application-frame* 'light)))
+    (when value
+      (setf (light-color pane) +gold1+)))
```
dkochmanski: this let when should be reversed
(when value (let ((pane …))))

dkochmanski:
this example could be improved if we had used a timer event instead of a "pause" function (see graph-toy-example) and specialize handle-event on that.

PR #535, opened 2018-08-31
Font Selector: Change 'one-armed' IFs to WHEN or AND

PuercoPop:
While reading through the font-selector I saw some some minor style improvements and went for them.

In general I've preferred WHEN, except when the IF form being replaced
was all in one line, in that case I've preferred to use AND.

### DIFF
--- a/Examples/font-selector.lisp
+++ b/Examples/font-selector.lisp
@@ -107,7 +107,7 @@
 
 (defun face-changed (pane value)
   (declare (ignore pane))
-  (let ((sizes (if value (font-face-all-sizes value) nil)))
+  (let ((sizes (and value (font-face-all-sizes value))))
     (reset-list-pane (find-pane-named *application-frame* 'size)
 		     sizes
 		     (or (position-if (lambda (x) (>= x 20)) sizes) 0))))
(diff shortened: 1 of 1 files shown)


dkochmanski:
I disagree with these changes. While I'm for replacing one-armed ifs with while/and etc, it only applies to the situation, where we do not care about the result. When block is used as argument to setf or as a binding (imho) it should stay as if and NIL should be clearly written.

Examples/font-selector.lisp
```diff
@@ -107,7 +107,7 @@
 
 (defun face-changed (pane value)
   (declare (ignore pane))
-  (let ((sizes (if value (font-face-all-sizes value) nil)))
+  (let ((sizes (and value (font-face-all-sizes value))))
```
dkochmanski: this is also semantically incorrect:
`(if value sizes nil)` means if there is value, return all sizes, otherwise nil
`(and value sizes)` means return whenever both value and sizes are non-NIL
while it doesn't make a difference to the compiler in case of Common Lisp, it does for a reader

PuercoPop: > this is also semantically incorrect:

Given that the forms are semantically equivalent, it is incorrect to say it is semantically incorrect. The truth table for the value sizes is bound to are identical in both cases. `AND` is an idiomatic way to write one line `when`s. I agree it does make a different to the reader, it maintains a consistent way to separate clauses inside of `if`s and `when`s, using new lines.

PuercoPop:
Thank you for being straightforward. I also want to _sincerely_ commend you on saying no. Sometimes as a maintainer it is hard to say no to changes one disagrees, but it is important to do so imho.

That said I disagree with your assessment. Lisp is an expression oriented language. The `(when ...)` form returns `nil`. There is no value in being redundant (either way the code is explicitly returning nil). The result is noiser, non-idiomatic, code.

I'll close the PR as the final call is yours to make. Keep up the good work with McCLIM!

PR #541, opened 2018-09-14
define-command: be more conforming with evaluating argument descriptions

### LINKED ISSUE #540: Command argument presentation types not evaluated
Hi,

I'm an old time CLIM user going back to my days at Symbolics.  I've been using the Allegro version for years and years.  Just started to see if McClim is up to the task of supporting my body of code.

An issue that bit me is that in the release I have command arguments to define-command aren't evaluated (except for a hack that doesn't work for my code)

The following comment is in the code for define-command:

  ;; Argument types shouldn't be evaluated. Unfortunately in McCLIM
  ;; they are, and in all McCLIM applications up to day types are
  ;; quoted. To preserve backward compatibility we honor this, but we
  ;; don't evaluate types which are not quoted, but rather add the
  ;; quotation ourself.
  ;;
  ;; Thanks to that we should achieve compatibility with other CLIMs
  ;; without breaking already existing applications (unless they did
  ;; some fancy computation in `define-command' macro which isn't
  ;; conforming anyway).

But that isn't what the CLIM 2 Spec says:

"name-and-options and body are not evaluated. In the argument descriptions, the parameter variable name is not evaluated, and everything else is evaluated at run-time when argument parsing reaches that argument, except that the value for :when is evaluated when parsing reaches the keyword arguments, and :gesture isn’t evaluated at all."

dkochmanski:
Differences:
- if presentation type is an atom, we autoquote it
  (if we really want to put evaluated variable here, do (identity ptype))
- we requiere macroexpand-constants for keyword specifiers

Improvements:
- we enforce a set of allowed keywords in argument description
- extra allowed keywords are PROMPT-MODE and INSERT-DEFAULT to prevent breaking
  ESA. Do they work? ESA should be fixed in this regard.

Fixes #540.

### DIFF
--- a/Core/clim-core/commands.lisp
+++ b/Core/clim-core/commands.lisp
@@ -1130,24 +1130,22 @@ examine the type of the command menu item to see if it is
   ;; the spec annotation we require keywords being macroexpand-time
   ;; constant). We allow two custom key arguments, but we should fix ESA
   ;; instead. -- jd 2018-09-14
-  (block nil
-    (mapc (lambda (argument-description)
-            ;; Keyword arguments are normally processed.
-            (when (eq argument-description '&key) (return))
+  (mapc (lambda (argument-description)
+          (unless (eq argument-description '&key)
             ;; Ensure correct structure and valid keywords.
             (destructuring-bind (parameter type &key
-                                 default default-type display-default mentioned-default
-                                 prompt documentation when gesture
-                                 ;; These two are not standard, but ESA uses them.
-                                 prompt-mode insert-default)
+                                default default-type display-default mentioned-default
+                                prompt documentation when gesture
+                                ;; These two are not standard, but ESA uses them.
+                                prompt-mode insert-default)
                 argument-description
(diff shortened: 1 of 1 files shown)


nsrahmad:
It looks like McCLIM behaved like this from ~2006. In `dragndrop-translator` demo, `named-color` has to be quoted now:
https://github.com/McCLIM/McCLIM/blob/b89d9616e6a304a288211f18d5e7ea9dc9ba87f0/Examples/dragndrop-translator.lisp#L73-L75
Please update above example.

dkochmanski:
hm, I think we should autoquote all key arguments to reassemble functions / methods. they are not argument descriptions.

dkochmanski:
hm, it seems I've been wrong. all arguments are argument-descriptions. key arguments should be processed exactly like required ones.

dkochmanski:
I've pushed commit for that a moment ago.

nsrahmad:
It took me some time to grok it. Current implementation is the best possible interpretation of the spec, merging it now.

PR #551, opened 2018-09-20
make-space-requirement: clamp optimal width/height to min/max bounds

dkochmanski:
We keep the assertion (<= min max), but min/optimal/max values are clamped
before that (min >=0, max <=+fill+, opt clamped to min;max).

That will make combining requirements more intuitive (previously, when we
specified only min-width for the component width defaulted to 0, which is
smaller, and dropped us in the debugger). Similarily, max-width/height could end
up being bigger than +fill+ when we called space-requirement+.

We also add min-width and min-height for text-pane to fit at least one
character. New modify macro is introduced in climi called CLAMPF (clamps and
setf's the place).


Core/clim-core/panes.lisp
```diff
@@ -243,17 +243,20 @@ order to produce a double-click")
               max-height))))
 
 (defun make-space-requirement (&key (min-width 0) (min-height 0)
-                                    (width min-width) (height min-height)
-				    (max-width +fill+) (max-height +fill+))
-  (assert (<= 0 min-width width max-width) (min-width width max-width))
-  (assert (<= 0 min-height height max-height) (min-height height max-height))
+                                 (width min-width) (height min-height)
+                                 (max-width +fill+) (max-height +fill+))
+  ;; Defensive programming. space-requierment-+ may double max-width=+fill+.
```
nsrahmad: Typo in "requirement"; I didn't understand the rest of the sentence either.

dkochmanski: when we call `space-requirement+` on two space requirements which has `max-width` `+fill+`, then we end up with `max-width` being `(* +fill+ +fill+)`, hence *bigger* than our maximum size.

PR #556, opened 2018-09-28
Fix breakage from "medium-transformation-fixes" merge

scymtym:
I had to fix these things in order to load McCLIM without warnings or errors.

One is due to unusual `*read-default-float-format*`, but that used to not be a problem.


dkochmanski:
Merged, thanks!

PR #557, opened 2018-09-29
ellipsis: fix worst rotated-elliipsis offender

dkochmanski:
Note that something is still broken on one demo with ellipsis rotations when
used with CCL. I *think* that the underlying problem is with different float
rounding what leads to wrong region-intersection between ellipse and lien on CCL.


slyrus:
This is, essentially, what I tried at first and leaves things still broken.

PR #560, opened 2018-10-01
provide :initform for rgb-image-pixmap

slyrus:
* without this the various multiple-inheritance of rgb-image-pixmap
   results in an initform of :nil whereas the slot is designed to be a
   region, not (or null region).

 * The error only shows up (I think) when running with safety 3, but
   seems like a legitimate error.


PuercoPop:
Could this be handled by an `:initform (error ....)`? Either to require an argument or require a subclass to provide an init-form?

dkochmanski:
@slyrus bump

slyrus:
@dkochmanski should be ready to go now.

PR #564, opened 2018-10-05
Pdf ellipse drawing fixes

slyrus:
There's still some work to do, like fixing ovals, but, in general, the ellipse drawing with the PDF backend is much better with this code.


slyrus:
Superseded.

PR #566, opened 2018-10-07
Now it only uses exported symbols

josrr:
I changed the example so that it no longer uses not exported symbols.


nsrahmad:
Good work! But the commit message is problemetic. "it" in the message is anaphoric and must refer to something. Change the message to something like `Examples/color-slider: Now it only uses exported symbols"`.
You can run `git commit --amend` to edit the message, then push to github with `--force` option. Once it is done, I will merge this.

josrr:
It's done

PR #567, opened 2018-10-09
Examples/puzzle.lisp: no longer uses non exported symbols

josrr:
I got rid of climi::convert-from-relative-to-absolute-coordinates

### DIFF
--- a/Apps/Scigraph/dwim/present.lisp
+++ b/Apps/Scigraph/dwim/present.lisp
@@ -60,16 +60,9 @@ advised of the possiblity of such damages.
 
 (defun bounding-rectangle* (presentation)
   "Get the bounding edges of the presentation."
-  (let ((stream *standard-output*))
-    ;; Seem to need to know the stream under the presentation.
-    ;; Take a wild guess.
-    (multiple-value-bind (xoff yoff)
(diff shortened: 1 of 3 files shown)


josrr:
Yes I agree; In fact, what you propose is what I did first, but the call to `climi::convert-from-relative-to-absolute-coordinates` confused me a bit.

josrr:
./Apps/Scigraph/dwim/present.lisp uses climi::convert-from-relative-to-absolute-coordinates; so I think we cannot get rid of it.

nsrahmad:
That file uses it exactly in this fashion, I will say go ahead and update that file too. If you do not feel like it then I will do that.

josrr:
OK, I will update that file.

josrr:
The specification mentions the function.

http://bauhh.dyndns.org:8000/clim-spec/edit/apropos?q=redisplay-output-record

nsrahmad:
That's under the `implementation-note` and not actually part of the spec. In McCLIM that function always returns two zeroes and not really useful for anything.

For now, leave that one out.

nsrahmad:
Looks good, Thank you for your work. It made McCLIM better.

PR #571, opened 2018-10-14
Pdf ellipse drawing fixes 3

slyrus:


slyrus:
Nisar, the point of test-runner.lisp is that one can easily make new drawing-test-style applications without having to duplicate all of the test display infrastructure code. I suppose we could just add these test cases to drawing-tests but it seemed it was already getting a bit bloated.

nsrahmad:
@slyrus Okay, I see now, But in its current state it adds lot of duplication. If you want to keep `test-runner`, then you also need to divide drawing-tests into `ellipse-test`, `line-tests`, `polygon-tests`, `text-tests` etc. and get rid of drawing-tests.

dkochmanski:
imho it should be added to drawing-tests and they should be worked upon, so we have two lists: one selects primitive (test category), and second selects a test. then it will be less bloated, still we'll have all drawing tests in one place.

slyrus:
@dkochmanski @nsrahmad thanks for the comments. I'm going to leave the suggested improvements to drawing-tests for another day. For now, the ellipse tests have been added there, test-runner and ellipse-tests removed and the mcclim-ellipse extension has been removed.

package.lisp
```diff
@@ -1952,7 +1952,18 @@
...
+   #:reparameterize-ellipse
+   #:ellipse-point
+   #:ellipse-point*
+   #:ellipse-derivative
+   #:ellipse-cubic-bezier-control-points
+   #:ellipse-cubic-bezier-control-points*
+   #:ellipse-cubic-bezier-points
+   #:ellipse-cubic-bezier-points*
+   #:transformation-angle
+   #:transform-angle))
```
dkochmanski: do you really think these methods should be part of official McCLIM API advertised outside? If so, please provide documentation what would I potentially need ellipse-derivative function for (etc). I think that these methods should be part of climi and not exported at all unless there is indeed a good usecase for normal application writing to know these. Programmer (imho) is only interested about the fact, that there are ellipsis and how he can draw it, not how it is done.

dkochmanski: nb: we should probably write some developer guideline with a consistent set of rules for that

dkochmanski:
also, does these changes fix clx backend issues with polyline and bad starting angle taken?

package.lisp
slyrus: I don't feel strongly about where these live. They can live in climi if you would like. The PDF backend needs reparameterize-ellipse, ellipse-cubic-bezier-points, and transform-angle. transformation-angle is a useful utility that I could see others wanting. drawing-tests want parameterize-ellipse as well. I plan on using some of these functions in my ellipse drawing code to help make a nice UI to the control points. personally, I'm a fan of lots of little packages but that isn't McCLIM style. Happy to move it all to climi if you would like.

dkochmanski: then please do (there is no need for exporting these symbols, our backends are part of McCLIM internals anyway, so they can directly `import-from` from climi).

package.lisp
```diff
@@ -2051,7 +2040,18 @@
...
```
dkochmanski: this is still wrong. clim-backend package exports symbols, which are meant to be implemented by a backend to be used from the core (so climb client is core module and provider is backend, not the other way around). Backend should import directly from climi.

slyrus:
@nsrahmad let me know if you have any other comments/requests.

PR #572, opened 2018-10-14
Define getenv for clasp

kpoeck:
There are other errors loading mcclim in clasp, but these are errors in clasp, so I don't want to workaround in mcclim

### DIFF
--- a/Core/clim-basic/utils.lisp
+++ b/Core/clim-basic/utils.lisp
@@ -15,7 +15,8 @@
   #+openmcl (ccl::getenv string)
   #+lispworks (lw:environment-variable string)
   #+ecl (ext:getenv string)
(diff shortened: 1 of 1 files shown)


nsrahmad:
@kpoeck Does clasp have MOP support? because If it does not yet, then I will wait for that first.

kpoeck:
Clasp has good mop-support, it is even already defined in closer-mop. 

See https://gist.github.com/kpoeck/191bb618d8ab029488f4289d14a814e3 for proof that I have mcclim more or less running in clasp.

dkochmanski:
This looks OK, I confirm that clasp has mop support in closer mop (https://github.com/pcostanza/closer-mop/blob/master/closer-clasp.lisp) and bt support too. I'm merging this as it is (especially that it is a minimal change). Thanks.

PR #574, opened 2018-10-16
Print object enhancements 5

slyrus:
These commits allow for printing some CLIM objects readably when *print-readably* is t.

### DIFF
--- a/Core/clim-basic/regions.lisp
+++ b/Core/clim-basic/regions.lisp
@@ -159,6 +159,17 @@
 (defclass standard-polygon (polygon)
   ((points :initarg :points)))
 
+(defmethod slots-for-pprint-object append ((object standard-polyline))
+  '(points closed))
+
+(defmethod print-object ((self standard-polyline) sink)
+  (cond
+    ((and *print-readably* (not *read-eval*))
+     (error "cannot readably print standard-polyline when not *read-eval*."))
+    ((and *print-pretty* *print-readably*)
+     (simple-pprint-object sink self))
+    (t (print-unreadable-object (self sink :identity nil :type t)))))
+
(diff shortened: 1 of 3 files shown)


nsrahmad:
@slyrus I've seen `*print-readably*` used for serialization, but never for debugging. Do you mind telling how you used it?

slyrus:
@nsrahmad as an example of where I would like to use this, I'm currently trying to track down a problem with arc drawing in the PDF backend. I see this frame the output of a trace of my transform-angle function: #<CLIM-INTERNALS::STANDARD-HAIRY-TRANSFORMATION 1 0 0 -1 0 842>. It would be great if I were to set *print-readably* to t and have the output be something like #.(make-transformation 1 0 0 -1 0 842). Yes, I can do this by hand, but why bother? Particularly for the immutable classes, this seems like a win.

Core/clim-basic/regions.lisp
```diff
@@ -551,12 +551,13 @@
...
-  (with-slots (start-angle end-angle tr) ell
-    (format stream "#<~A [~A ~A] ~A>"
-            (type-of ell)
-            (and start-angle (* (/ 180 pi) start-angle))
-            (and end-angle (* (/ 180 pi) end-angle))
-            tr)))
+  (print-unreadable-object (ell stream :type t :identity t)
+    (with-slots (start-angle end-angle tr) ell
+      (format stream "~A [~A ~A] ~A"
+              (type-of ell)
```
scymtym: I don't know if this has been mentioned before, but `print-unreadable-object … :type t` already prints the type so this shouldn't be needed.

slyrus:
I have addressed the requested changes and rebased to the new master branch. Closing this PR and opening a new one. Not sure if there's a better way to deal with conflicts/rebasing than that, but here goes.

PR #576, opened 2018-10-17
Pdf arc angle fix

slyrus:
This fixes some problems drawing arcs and ovals in the PDF backend, along with adding some new tests, and fixing a backend-independent problem with losing medium options when drawing certain ovals.


Backends/PDF/graphics.lisp
```diff
@@ -124,50 +138,67 @@ end-angle are NIL and filled is T, then a solid ellipse is drawn. If
...
-                       (segment-count (ceiling (abs (/ sweep (/ pi 4)))))
+    (flet ((draw-ellipse-segment (lambda1 lambda2)
+             (multiple-value-bind (p1x p1y q1x q1y q2x q2y p2x p2y)
+                 (ellipse-cubic-bezier-points lambda1 lambda2
+                                              center-x center-y
+                                              a b theta)
+               (with-transformed-position (tr p1x p1y)
+                 (with-transformed-position (tr q1x q1y)
+                   (with-transformed-position (tr q2x q2y)
+                     (with-transformed-position (tr p2x p2y)
```
dkochmanski: maybe `with-transformed-positions` macro will be more elegant here?

Backends/PDF/graphics.lisp
```diff
@@ -124,50 +138,67 @@ end-angle are NIL and filled is T, then a solid ellipse is drawn. If
...
+          ;; angles to go CCW in a :first-quadrant nil sense so we
+          ;; need to:
+          ;;
+          ;; 1. swap end-angle and start-angle
+          ;;
+          ;; 2. subtract minus theta from both start-angle and end-angle.
+          ;;
+          ;; 3. reverse the signs
+          ;;
+          ;; that reduces to:
```
dkochmanski: no need for empty newlines in this enumeration

Examples/drawing-tests.lisp
```diff
@@ -1321,6 +1342,43 @@ outside the clipping area should be grey.")
                         (draw-point* stream x1 y1 :ink +red+ :line-thickness 5)
                         (draw-point* stream x2 y2 :ink +green+ :line-thickness 5)))))))))
 
+
+(define-drawing-test "09) Filled Ovals" (stream)
+    ""
```
dkochmanski: please add verbal description what should be observed

Examples/drawing-tests.lisp
```diff
@@ -1321,6 +1342,43 @@ outside the clipping area should be grey.")
...
+                                    :line-thickness 3
+                                    :filled t
+                                    :ink +pink+)
+                        (draw-point* stream (/ (+ x1 x2) 2) (/ (+ y1 y2) 2)
+                                     :ink +blue+ :line-thickness 5)
+                        (draw-point* stream x1 y1 :ink +red+ :line-thickness 5)
+                        (draw-point* stream x2 y2 :ink +green+ :line-thickness 5)))))))))
+
+(define-drawing-test "09) Simple Oval 1" (stream)
+    ""
```
dkochmanski: ditto

Examples/drawing-tests.lisp
```diff
@@ -1321,6 +1342,43 @@ outside the clipping area should be grey.")
...
+  (draw-oval* stream 200 200 25 25 :ink +blue+ :filled nil :line-thickness 4))
+
+(define-drawing-test "09) Simple Oval 2" (stream)
+    ""
```
dkochmanski: ditto

Examples/drawing-tests.lisp
```diff
@@ -1321,6 +1342,43 @@ outside the clipping area should be grey.")
...
+  (draw-oval* stream 200 200 50 25 :ink +blue+ :filled nil :line-thickness 4))
+
+(define-drawing-test "09) Simple Oval 3" (stream)
+    ""
```
dkochmanski: ditto

slyrus:
I'd rather leave the polyline thing for a separate PR. If no one else does, I can take another look after this lands, but this is a separate (but important, at least for the PDF backend) issue.

Backends/PDF/graphics.lisp
slyrus: At the expense of consing up a coord-seq? I don't think that's worth it.

Backends/PDF/graphics.lisp
slyrus: Fixed.

Examples/drawing-tests.lisp
```diff
@@ -1018,6 +1026,20 @@ outside the clipping area should be grey.")
   (draw-circle* stream *center-x* *center-y* 150 :ink +orange+ :filled nil
                 :start-angle 0 :end-angle (/ pi 4)))
 
+(define-drawing-test "06) More Arcs" (stream)
+    ""
```
dkochmanski: this still lacks description.

Backends/PDF/graphics.lisp
```diff
@@ -124,50 +138,63 @@ end-angle are NIL and filled is T, then a solid ellipse is drawn. If
...
+      (if (or (and start-angle end-angle
+                   (zerop start-angle) (= (* pi 2) end-angle))
+              (and start-angle end-angle
+                   (zerop end-angle) (= (* pi 2) start-angle)))
+          (let* ((sweep (* pi 2))
+                 (steps 16)
+                 (step (/ sweep steps)))
+            (let ((seg-start-angle 0)
+                  (seg-end-angle (+ start-angle step)))
+              (loop for i below steps
```
dkochmanski: you may ignore this remark, but wouldn't it look better if you'd put all these let variables in loop as `with` clause?

PR #583, opened 2018-10-27
Small optimizations

scymtym:
The optimizations are all small but in important places (Case in point, profiling a real application lead me to these functions).


Core/clim-basic/coordinates.lisp
```diff
@@ -68,4 +68,6 @@
 (defun coordinate/= (x y)
   (/= x y))
 
+(deftype rectangle-coordinate-vector () '(simple-array coordinate (4)))
```
dkochmanski: do you mean a rectangle or a standard-rectangle? the difference is that the rectangle is not required to be aligned with XY (and it is a protocol class) while standard-rectangle is guaranteed to be aligned that way.

Core/clim-basic/recording.lisp
```diff
@@ -593,28 +594,28 @@ the associated sheet can be determined."
...
+      (with-slots (x y) record
+        (values x y x y))
+      (let ((new-x1 0)
+            (new-y1 0)
+            (new-x2 0)
+            (new-y2 0)
+            (first-time t))
+        (flet ((do-child (child)
+                 (cond ((null-bounding-rectangle-p child))
+                       (first-time
```
dkochmanski: I would find more readable `(unless (null …) (if first-time foo bar))`, but it is opinionated so feel free to ignore this remark.

PR #584, opened 2018-10-29
Initarg fixes and protocol class performance improvement

scymtym:


Backends/CLX/port.lisp
```diff
@@ -106,9 +106,12 @@
...
 (defmethod initialize-instance :after ((port clx-port) &rest args)
   (declare (ignore args))
-  (let ((options (cdr (port-server-path port))))
+  (let ((frame-manager-options
+          (loop for (key value) on (cdr (port-server-path port)) by #'cddr
+                when (member key '(:mirroring :class-gensym))
+                collect key and collect value)))
     (push (apply #'make-instance 'clx-frame-manager
-                 :port port options)
+                 :port port frame-manager-options)
```
dkochmanski: what happens with other options? maybe we should clean up what is passed in port-server-path? or, eventually, add allow-other-keys to clx-frame-manager.

scymtym: > add allow-other-keys to clx-frame-manager.
I did that.

I suggest leaving the cleanup for another time.

PR #599, opened 2018-10-31
Small backend improvements

dkochmanski:


Core/clim-basic/stream-output.lisp
```diff
@@ -290,32 +290,25 @@
 
 (defun seos-write-newline (stream)
   (let ((medium       (sheet-medium stream))
-        (view-height  (bounding-rectangle-height stream)))
+        ;; Stream region may be +everywhere+.
+        (view-height  (ignore-errors (bounding-rectangle-height stream))))
```
scymtym: Couldn't this be tested with `(region-equal (sheet-region stream) +everywhere+)` instead of `ignore-errors`?

Backends/CLX/medium.lisp
```diff
@@ -92,16 +92,37 @@ region and its clipping pixmap. This is looked up for optimization with region-e
...
+;;; XXX: this should be refactored into a reusable protocol in clim-backend
+;;; with specialization on medium. -- jd 2018-10-31
+(defun line-style-scale (line-style medium)
+  (let ((unit (line-style-unit line-style)))
+    (ecase unit
+      (:normal 1)
+      (:point (/ (graft-width (graft medium))
+                 (graft-width (graft medium) :units :inches)
+                 72))
+      (:coordinate (multiple-value-call #'(lambda (x y) (sqrt (+ (expt x 2) (expt y 2))))
```
scymtym: Why not `multiple-value-bind`?

Backends/CLX/medium.lisp
```diff
@@ -92,16 +92,37 @@ region and its clipping pixmap. This is looked up for optimization with region-e
...
-                     (line-style-thickness old-line-style))
-          ;; this is kind of false, since the :unit should be taken
-          ;; into account -RS 2001-08-24
+        (unless (and (eql (line-style-thickness line-style)
+                          (line-style-thickness old-line-style))
+                     (eql (line-style-unit line-style)
+                          (line-style-unit old-line-style)))
           (setf (xlib:gcontext-line-width gc)
-                (round (line-style-thickness line-style))))
+                (round (line-style-effective-thickness line-style medium))))
```
scymtym: Could `line-style-effective-thickness` also round, for consistency with `line-style-effective-dashes`?

package.lisp
```diff
@@ -1877,8 +1877,10 @@
    ;; sheets
    #:always-repaint-background-mixin
    #:never-repaint-background-mixin
-
-   #:line-style-effective-thickness
+   #:background
+   #:foreground
+   ;; #:line-style-effective-thickness
```
scymtym: Is this intentional? The commit message doesn't mention it.

Core/clim-basic/stream-output.lisp
dkochmanski: it could, I guess it was me being lazy (or overly general) here. Fix pushed.

Backends/CLX/medium.lisp
dkochmanski: at that point of time it seemed more natural (I didn't even think about m-v-b), but when you mention it I agree that it m-v-b is more idiomatic here. Fixed.

Backends/CLX/medium.lisp
dkochmanski: it does now. note, that it is clx-specific requirement to round these (renderer nor output-only backends doesn't seem to require that), what is a good indicator, that line-style-effective-* should be part of climb package (and not line-style-scale with these functions working on top of that). Fixed.

package.lisp
dkochmanski: Yes, I don't think this belongs in extensions (especially that it was not defined in all backends until now). We may think about readding it there /after/ we have working protocol in climb package.

PR #603, opened 2018-11-06
Accepting values regression

### LINKED ISSUE #602: Regression in accepting-values
## A
1) In `Address Book` demo, Run `Describe` command without any argument. `accepting-values` dialog will appear.
2) Select a Person's name and click `OK`.
3) You will get an error `Please supply all arguments`. This is a regression. You can try selecting the name again and add a newline to it. This time it will work.
## B
1) In `German Towns` demo, Select the menu item `Get Distance`. You will get an `accepting-values` dialog asking for two towns. Selecting a town from the map inserts the town in text-field but immediately puts a newline, so town is selected but you can't see it.

nsrahmad:
Fixes #602.

### DIFF
--- a/Core/clim-basic/recording.lisp
+++ b/Core/clim-basic/recording.lisp
@@ -1603,8 +1603,6 @@ were added."
     (string start end point-x point-y align-x align-y toward-x toward-y transform-glyphs)
   (and (if-supplied (string)
          (string= (slot-value record 'string) string))
-       (transformation-equal (graphics-state-transformation record)
-                             (graphics-state-transformation record))
        (if-supplied (start)
          (eql (slot-value record 'start) start))
        (if-supplied (end)
--- a/Core/clim-basic/recording.lisp
+++ b/Core/clim-basic/recording.lisp
@@ -1032,15 +1032,6 @@ were added."
   (if-supplied (text-style)
     (text-style-equalp (slot-value record 'text-style) text-style)))
 
-(defmethod replay-output-record :around
-    ((record gs-transformation-mixin) stream &optional region x-offset y-offset)
-  (declare (ignore region x-offset y-offset))
-  (with-drawing-options (stream :transformation (graphics-state-transformation record))
-    (call-next-method)))
-
-(defrecord-predicate gs-transformation-mixin (transformation)
-  (transformation-equal (graphics-state-transformation record) transformation))
-
 (defclass standard-graphics-displayed-output-record
     (standard-displayed-output-record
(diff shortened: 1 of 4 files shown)


Core/clim-basic/recording.lisp
```diff
@@ -1604,13 +1595,16 @@ were added."
...
+      (with-drawing-options (medium :transformation transformation)
+        (medium-draw-text* medium string point-x point-y start end align-x
+                           align-y toward-x toward-y transform-glyphs)))))
 
 (defrecord-predicate draw-text-output-record
     (string start end point-x point-y align-x align-y toward-x toward-y transform-glyphs)
   (and (if-supplied (string)
          (string= (slot-value record 'string) string))
+       (transformation-equal (graphics-state-transformation record)
+                             (graphics-state-transformation record))
```
dkochmanski: is there even a remote possibility, that this won't be T? we are calling transformation-equal on the same object after all.

Core/clim-basic/recording.lisp
```diff
@@ -1032,15 +1032,6 @@ were added."
   (if-supplied (text-style)
     (text-style-equalp (slot-value record 'text-style) text-style)))
 
-(defmethod replay-output-record :around
-    ((record gs-transformation-mixin) stream &optional region x-offset y-offset)
-  (declare (ignore region x-offset y-offset))
-  (with-drawing-options (stream :transformation (graphics-state-transformation record))
-    (call-next-method)))
```
dkochmanski: how is this breaking the dialog? I'm still convinced, it is the right thing to do. if some other record type doesn't want with-drawing-options :transformation, then it shouldn't inherit from gs-transformation-mixin.

And this is around replay-output-record on draw-text-output-record, which inherits from gs-transformation-mixin, so for this kind of records it should make a difference (so this should be functionally equivalent to the reverted version).

Core/clim-basic/recording.lisp
```diff
@@ -1604,13 +1595,16 @@ were added."
                toward-y transform-glyphs transformation)
       record
     (let ((medium (sheet-medium stream)))
-      (medium-draw-text* medium string point-x point-y start end align-x
-                         align-y toward-x toward-y transform-glyphs))))
+      (with-drawing-options (medium :transformation transformation)
+        (medium-draw-text* medium string point-x point-y start end align-x
+                           align-y toward-x toward-y transform-glyphs)))))
```
dkochmanski: ditto

PR #606, opened 2018-11-12
Mouse wheel improvements

scymtym:
* `slider-pane` and `scroll-bar-pane` react to `pointer-scroll-event`

* Scroll events (with at least one modifier) can be turned into gestures and used for commands

* Unrelated refactoring of `slider-pane`


dkochmanski:
superseded by #1057

PR #608, opened 2018-11-15
Bug fixes

scymtym:
I feel pretty confident about everything except maybe "remove "adapter" for DRAG-N-DROP-TRANSLATOR".

### DIFF
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -1,10 +1,10 @@
 ;;; -*- Mode: Lisp; Package: CLIM-INTERNALS -*-
 
 ;;;  (c) copyright 1998,1999,2000 by Michael McDonald (mikemac@mikemac.com)
-;;;  (c) copyright 2000 by 
+;;;  (c) copyright 2000 by
 ;;;           Iban Hatchondo (hatchond@emi.u-bordeaux.fr)
 ;;;           Julien Boninfante (boninfan@emi.u-bordeaux.fr)
-;;;  (c) copyright 2000, 2014 by 
+;;;  (c) copyright 2000, 2014 by
 ;;;           Robert Strandh (robert.strandh@gmail.com)
 ;;;  (c) copyright 2004 by
 ;;;           Gilbert Baumann <unk6@rz.uni-karlsruhe.de>
@@ -20,8 +20,8 @@
 ;;; Library General Public License for more details.
 ;;;
 ;;; You should have received a copy of the GNU Library General Public
-;;; License along with this library; if not, write to the 
-;;; Free Software Foundation, Inc., 59 Temple Place - Suite 330, 
+;;; License along with this library; if not, write to the
+;;; Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 ;;; Boston, MA  02111-1307  USA.
 
 (in-package :clim-internals)
@@ -70,7 +70,7 @@
 (defgeneric note-input-focus-changed (pane state)
   (:documentation "Called when a pane receives or loses the keyboard
 input focus. This is a McCLIM extension."))
-    
+
 (defclass standard-application-frame (application-frame
 				      presentation-history-mixin)
   ((port :initform nil
@@ -130,7 +130,7 @@ input focus. This is a McCLIM extension."))
 		     :reader frame-top-level-lambda)
    (hilited-presentation :initform nil
 			 :initarg :hilited-presentation
-			 :accessor frame-hilited-presentation)   
+			 :accessor frame-hilited-presentation)
    (process :accessor frame-process :initform nil)
    (client-settings :accessor client-settings :initform nil)
    (event-queue :initarg :frame-event-queue
@@ -195,7 +195,7 @@ documentation produced by presentations.")
 			(space-requirement-width (compose-space pane))))
 	     (height (or geometry-height
 			 (and geometry-top geometry-bottom (- geometry-bottom geometry-top))
-			 (space-requirement-height (compose-space pane))))	   
+			 (space-requirement-height (compose-space pane))))
 	     ;; See if a position is wanted and return left, top.
 	     (left (or geometry-left
 		       (and geometry-right (- geometry-right geometry-width))))
@@ -711,7 +711,7 @@ documentation produced by presentations.")
   (setf (slot-value pane 'name) name)
   pane)
 
-(defun do-pane-creation-form (name form)  
+(defun do-pane-creation-form (name form)
   (cond
     ;; Single form which is a function call
     ((and (= (length form) 1)
@@ -828,7 +828,7 @@ documentation produced by presentations.")
 	(disabled-commands nil)
 	(command-definer t)
 	(top-level '(default-frame-top-level))
-	(others nil)	
+	(others nil)
 	(pointer-documentation nil)
 	(geometry nil)
 	(user-default-initargs nil)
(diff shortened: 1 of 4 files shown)


Core/clim-core/frames.lisp
```diff
@@ -1668,13 +1670,14 @@ have a `pointer-documentation-pane' as pointer documentation,
...
-		    feedback-fn (feedback dest-translator)
-		    hilite-fn (highlighting dest-translator))
+		    feedback-fn       (when dest-translator (feedback dest-translator))
+		    hilite-fn         (when dest-translator (highlighting dest-translator)))
 	      (do-hilite presentation window :highlight)
 	      (do-feedback window x y :highlight)
-	      (document-drag-n-drop dest-translator presentation
-				    context-type frame event window
-				    x y)))
+	      (when dest-translator
```
dkochmanski: imho this should be handled in document-drag-n-drop, something in a spirit of diff below. rationale: this is a local function to frame-drag-and-drop (I guess it was put standalone due to length), such handling would be consistent with what you did with `do-highlight` and `do-feedback`.

```diff
 (defun document-drag-n-drop
     (translator presentation context-type frame event window x y)
-  (when *pointer-documentation-output*
-    (let ((s *pointer-documentation-output*))
-      (window-clear s)
-      (with-end-of-page-action (s :allow)
-	(with-end-of-line-action (s :allow)
-	  (funcall (pointer-documentation translator)
-	   *dragged-object*
-	   :presentation *dragged-presentation*
-	   :destination-object (and presentation
-				    (presentation-object presentation))
-	   :destination-presentation presentation
-	   :context-type context-type
-	   :frame frame
-	   :event event
-	   :window window
-	   :x x
-	   :y y
-	   :stream s))))))
+  (alexandria:when-let* ((s *pointer-documentation-output*)
+                         (pd (and translator (pointer-documentation translator))))
+    (window-clear s)
+    (with-end-of-page-action (s :allow)
+      (with-end-of-line-action (s :allow)
+        (funcall pd *dragged-object*
+                 :presentation *dragged-presentation*
+                 :destination-object (and presentation
+                                          (presentation-object presentation))
+                 :destination-presentation presentation
+                 :context-type context-type
+                 :frame frame
+                 :event event
+                 :window window
+                 :x x
+                 :y y
+                 :stream s)))))
```

Core/clim-core/frames.lisp
```diff
@@ -1684,9 +1687,10 @@ have a `pointer-documentation-pane' as pointer documentation,
 	      (do-hilite last-presentation (last-window) :unhighlight))
 	    (setq last-presentation nil)
 	    (do-feedback window x y :highlight)
-	    (document-drag-n-drop translator nil
-				  context-type frame event window
-				  x y))
+	    (when translator
+              (document-drag-n-drop translator nil
```
dkochmanski: Is this possible that translator is NIL?

```
;; Try to run the feedback and highlight functions of the translator
	 ;; that got us here.
	 (translator (or (find translator-name translators :key #'name)
			 (car translators)))
```

Core/clim-core/frames.lisp
```diff
@@ -1668,13 +1670,14 @@ have a `pointer-documentation-pane' as pointer documentation,
 	      (when last-presentation
 		(do-hilite last-presentation (last-window) :unhighlight))
 	      (setq last-presentation presentation
-		    feedback-fn (feedback dest-translator)
-		    hilite-fn (highlighting dest-translator))
+		    feedback-fn       (when dest-translator (feedback dest-translator))
+		    hilite-fn         (when dest-translator (highlighting dest-translator)))
```
dkochmanski: I wonder if we should set it to NIL or we should use default functions as specified for `define-drag-and-drop-translator` (namely `frame-drag-and-drop-highlighting` and `frame-drag-and-drop-feedback`). If we do the latter, then this should be if and we don't have to check whenever feedback-fn is NIL.

If you happen to hit slot of the translator being NIL, I think this is a problem which should be handled during translator definition (so it defaults to what it is meant to, and if NIL is passed explicitly that should error).

```
 		(do-hilite last-presentation (last-window) :unhighlight))
-	      (setq last-presentation presentation
-		    feedback-fn (feedback dest-translator)
-		    hilite-fn (highlighting dest-translator))
+	      (if dest-translatorn
+                  (setq last-presentation presentation
+                        feedback-fn (feedback dest-translator)
+                        hilite-fn (highlighting dest-translator))
+                  (setq last-presentation presentation
+                        feedback-fn #'frame-drag-and-drop-feedback
+                        hilite-fn #'frame-drag-and-drop-highlighting))
```

Core/clim-core/presentation-defs.lisp
```diff
@@ -1518,6 +1518,7 @@ protocol retrieving gestures from a provided string."))
   (frob real)
   (frob rational)
   (frob ratio)
+  (frob integer)
```
dkochmanski: I think you made a type in the commit message.

Core/clim-core/presentation-defs.lisp
```diff
@@ -2303,21 +2302,9 @@ protocol retrieving gestures from a provided string."))
...
-                        :destination-presentation presentation
-                        args)
-                 (apply func object args)))))
-    (setf (slot-value obj 'documentation) (make-adapter documentation))
-    (when pointer-documentation
-      (setf (slot-value obj 'pointer-documentation)
-            (make-adapter pointer-documentation)))))
+  (setf (slot-value obj 'documentation) documentation)
+  (when pointer-documentation
+    (setf (slot-value obj 'pointer-documentation) pointer-documentation)))
```
dkochmanski: I agree with this interpretation.

Core/clim-core/presentation-defs.lisp
scymtym: Fixed.

Core/clim-core/presentation-defs.lisp
scymtym: OK.

Core/clim-core/frames.lisp
scymtym: Likely. With the your suggested change to `document-drag-n-drop` it will be safe either way.

Core/clim-core/frames.lisp
scymtym: Good idea. I'm changing the commit accordingly.

Core/clim-core/frames.lisp
scymtym: If there is no `dest-translator`, highlighting should happen, as far as I can tell. I adapted the suggestion accordingly.

Core/clim-core/frames.lisp
```diff
@@ -1727,24 +1729,21 @@ have a `pointer-documentation-pane' as pointer documentation,
...
-	   :event event
-	   :window window
-	   :x x
-	   :y y
-	   :stream s))))))
-
-
-
+  (let ((stream *pointer-documentation-output*)
+        (function (pointer-documentation translator)))
```
dkochmanski: I think it wasn't looking like that in the previous version. Are you sure that `*pointer-documentation-output*` is always bound to a stream and this `when` was unnecessary? Is it certain, that `(pointer-documentation translator)` will return a function?

Everything else looks fine to me, although I'm a bit skeptical about using multiple-value calls.

scymtym: You are right with respect to `*pointer-documentation-output*`, it can be `nil`. I put the guard back in and tested with programs with and without pointer documentation pane. So that should be fine.

`(pointer-documentation translator)` should never be `nil`. I read the spec that way, the `define-drag-and-drop-translator` macro inserts a default value if necessary and the other call to `(pointer-documentation translator)` in the code base also assumes that it is always a function.

mv calls avoid having three helper functions that all check whether `last-event` has already been set (I found that problem while testing the other changes). I it's fine with you, I would keep them.

PR #623, opened 2018-11-29
Scigraph: bug fix

admich:
Simple bug fix for Scigraph. 

Invalid initialization argument:
  :INITIAL-CURSOR-VISIBILITY

:initial-cursor-visibility initarg is not accepted in application-pane  initialization and it is not in the specification.

Before commit 
9434255f * bad protocol-classes: remove runtime overhead, restore initarg checking

this didn't produce an error, now it does.

### DIFF
--- a/Apps/Scigraph/scigraph/frame.lisp
+++ b/Apps/Scigraph/scigraph/frame.lisp
@@ -72,9 +72,7 @@ class.  So you should do the following (in the ws package, 'natch):
    (display :application
 	    :display-function 'redisplay-graphs
 	    :display-time t
-	    :text-style 
-	    (parse-text-style '(:fix :roman :normal))
-	    ;; :initial-cursor-visibility nil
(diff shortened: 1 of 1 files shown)


Apps/Scigraph/scigraph/frame.lisp
```diff
@@ -74,7 +74,7 @@ class.  So you should do the following (in the ws package, 'natch):
 	    :display-time t
 	    :text-style 
 	    (parse-text-style '(:fix :roman :normal))
-	    :initial-cursor-visibility nil
+	    ;; :initial-cursor-visibility nil
```
dkochmanski: please simply remove the line. since you are at this part, please also move parse-text-style to the same line as the keyword. thanks!

admich:
I made the changes requested.

dkochmanski:
Thank you, merged

PR #625, opened 2018-12-03
remove :borders arg from make-pane call in ESA

slyrus:


slyrus:
@scymtym The :borders argument triggers an error when one tries to create an esa::info-pane (as Gsharp does). I don't know if this ever worked (presumably it must have) but in the current McCLIM, borders is not an initarg to info-pane, or its superclass application-pane. Removing the :borders arg from the :default-initargs allows Gsharp to run. Without doing so, we get a (IMO correct) error.

slyrus:
Closing and reopening a new PR with an updated commit message.

PR #626, opened 2018-12-04
remove :borders arg from make-pane call in ESA

slyrus:
* The :borders argument that used to be here triggers an error when
   one tries to create an esa::info-pane (as Gsharp does). I don't
   know if this ever worked (presumably it must have) but in the
   current McCLIM, :borders is not an initarg to info-pane, or its
   superclass application-pane. Removing the :borders arg from the
   :default-initargs allows Gsharp to run. Without doing so, we get an
   error.

### DIFF
--- a/Libraries/ESA/esa.lisp
+++ b/Libraries/ESA/esa.lisp
@@ -82,8 +82,7 @@ will probably have the same value as `*application-frame*'.")
   ((master-pane :initarg :master-pane :reader master-pane))
   (:default-initargs
       :background +gray85+
-      :scroll-bars nil
-      :borders nil))
+      :scroll-bars nil))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; 


dkochmanski:
`:borders` is a key-arg to `make-clim-stream-pane` just as `:scroll-bars` is.

Accidently (or in purpose) `clim-stream-pane` accepts `:scroll-bars` initarg and has a slot which preserves scroll value (but it doesn't have a slot for borders, that's why we have an error).

For consistency this fix should be more elaborate. I have two proposals:

1. remove scroll-bars initarg from clim-stream-pane and initargs associated with this (easier, but will probably introduce some regressions).

2. (preferred by me) always call clim-make-stream-pane in `do-pane-creation-form` from frames.lisp. Looking at its description[1] it is not said that pane creates *must* be a stream. That way these initargs if supplied in `:panes` will be recognized alright. That still means, that we should remove `:borders` and `:scroll-bars` from default initargs if they happen to be defined anywhere like that. Also `:scroll-bars` slot should be removed from clim-stream-pane.

What do you think?

[1] http://bauhh.dyndns.org:8000/clim-spec/edit/apropos?q=make-clim-stream-pane

slyrus:
Closing and opening new PR with updated commit.

PR #627, opened 2018-12-06
Xkeyboard

quasus:
I’m adding some support for XKB using a patched version of the clx-xkeyboard library available at
https://github.com/quasus/clx-xkeyboard
(with a rudimentary support for XKB events).  The modifications address the bug "clx: input: english layout" https://www.bountysource.com/issues/35049403-450-clx-input-english-layout

### DIFF
--- a/Backends/CLX/mcclim-clx.asd
+++ b/Backends/CLX/mcclim-clx.asd
@@ -5,10 +5,12 @@
                #:mcclim-clx/basic
 	       #:mcclim-clx/input
 	       #:mcclim-clx/output
-	       #:mcclim-clx/text-selection)
+	       #:mcclim-clx/text-selection
+               #:xkeyboard)
   :components
   ((:file "port")
-   (:file "frame-manager" :depends-on ("port"))))
+   (:file "frame-manager" :depends-on ("port"))
+   (:file "xkb")))
 
 (defsystem #:mcclim-clx/basic
   :depends-on (#:clx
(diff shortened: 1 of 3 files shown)


dkochmanski:
Hey! How is this change relevant to https://github.com/sharplispers/clx/pull/100? Can't it be implemented on CLX side as an extension and then used from McCLIM?

quasus:
Yes, that was my idea, too: to rely on an implementation of XKB rather than to partly reimplement it within McCLIM.  The McCLIM code I propose is basically glue.  In terms of XKB support, we only have Filonenko’s clx-xkeyboard at the moment, which is very incomplete and not a part of CLX.  Surprisingly, in this particular case it almost worked but for the lack of events.  I made a provisional fix, but I thought I’d have to figure out how to implement XKB events within clx-xkeyboard and make a pull request there as well.

clx#100 seems complementary to my pull request.  It’ll be great if XKB is better supported and integrated into CLX, and hopefully, I could contribute at least concerning the events.

dkochmanski:
so what do we do? are we waiting for clx#100 (do you plan to work on this PR too?) or you think it is a better idea to merge it as is (after conflict being resolved and a peer review)?

quasus:
I think it makes sense to put it off for the time being.  Meanwhile, I’ll have a go at XKB events.  When implemented, they could be incorporated into CLX or clx-xkeyboard, and after that we could decide about McCLIM.

dkochmanski:
OK, I will close it then for now. If you have a separate repository for xkb events please let me know.

PR #636, opened 2018-12-17
Box adjuster improvements

scymtym:
This re-implementation uses a different approach: instead of tweaking the space requirements of the left and right client, the box client strategies (fixed size, proportional size, fill) are adjusted in a way that reserves the general strategy (e.g. a client using fill above a client using proportional) but changes the parameters.

Furthermore, instead of tracking an offset from the start point of the drag, the pointer position in the coordinate system of the left child is tracked. This is necessary for the implementation strategy described above.

### DIFF
--- a/Core/clim-core/gadgets.lisp
+++ b/Core/clim-core/gadgets.lisp
@@ -2866,14 +2866,23 @@ if INVOKE-CALLBACK is given."))
     ((or hbox-pane hrack-pane) :vertical)
     ((or vbox-pane vrack-pane) :horizontal)))
 
+(cl:defconstant +box-adjuster-gadget-major-size+ 6)
+(cl:defconstant +box-adjuster-gadget-minor-size+ 1)
+
+(defclass box-adjuster-gadget-dragging-state ()
+  ((%left         :initarg :left
+                  :reader  dragging-state-left)
+   (%right        :initarg :right
+                  :reader  dragging-state-right)
+   (%clients-size :initarg :clients-size
+                  :reader  dragging-state-clients-size)
+   (%total-size   :initarg :total-size
+                  :reader  dragging-state-total-size)))
 
 (defclass clim-extensions::box-adjuster-gadget
-               (basic-gadget 3d-border-mixin orientation-from-parent-mixin)
-  ((drag-from :initform nil)
-   (left-sr)
-   (left-peer)
-   (right-sr)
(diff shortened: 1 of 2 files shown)


Core/clim-core/gadgets.lisp
```diff
@@ -2886,62 +2895,94 @@ it in a layout between two panes that are to be resizeable.  E.g.:
...
                           &key width height)
   (declare (ignore width height))
-  (if (eq (orientation gadget) :vertical)
-      (make-space-requirement :min-width 6 :width 6 :max-width 6
-                              :min-height 1 :height 1)
-      (make-space-requirement :min-height 6 :height 6 :max-height 6
-                              :min-width 1 :width 1)))
+  (let ((major-size +box-adjuster-gadget-major-size+)
+        (minor-size +box-adjuster-gadget-minor-size+))
+    (if (eq (orientation gadget) :vertical)
```
dkochmanski: if it is either vertical or horizontal ecase would be more specific

scymtym: I thought so as well when reading the original code. Should have acted on the thought. I will update the commit accordingly.

In general I would like to make dealing with orientations less annoying. My current thinking is something like this:

```lisp
(with-orientation (orientation gadget)
  …
  (bounding-rectangle-major-size …)
  (bounding-rectangle-max-major …)
  …)
```

where `with-orientation` would bind a dynamic variable and "major" and "minor" accessors would read it.

dkochmanski: I think that "dada" evil macro did what you have suggested in a hacky way, please see "panes.lisp" and vbox/hbox panes.

Core/clim-core/gadgets.lisp
```diff
@@ -2866,14 +2866,23 @@ if INVOKE-CALLBACK is given."))
     ((or hbox-pane hrack-pane) :vertical)
     ((or vbox-pane vrack-pane) :horizontal)))
 
+(cl:defconstant +box-adjuster-gadget-major-size+ 6)
+(cl:defconstant +box-adjuster-gadget-minor-size+ 1)
```
dkochmanski: this is just a note (not change request) - we should find where we depend on non-conforming defconstant, fix these places and do not patch defconstant in climi.

dkochmanski:
merging, thanks!

PR #639, opened 2018-12-23
README.md improvements

slyrus:
* Replaced the inline HTML with actual markdown. We lose the
   centering, but gain the ability to have Markdown parsers actually
   successfully parse the README.md file, which is particularly nice
   for eval'ing the lisp code directly out of README.md.

 * fix some long line/line ending spaces

 * Use markdown numbered lists (1., 2., , etc...) instead of 1), 2), etc...


dkochmanski:
lgtm, thanks!

PR #652, opened 2019-01-01
Fix borders

dkochmanski:


lokedhs:
I've been testing it, and it seems to work for every test I done.

Issue #624, opened 2018-11-30
Label panes and other labelled gadgets ignore the size component of *default-text-style*


john-a-carroll:
The label-pane class definition :default-initargs include `:text-style (make-text-style :sans-serif nil nil)`, and \*default-text-style\* is not merged in before drawing the label. This means that the label text may be a different size to the size in \*default-text-style\*. List and option panes have similar behaviour in this respect: their text (and that of the option-pane popup) is always drawn in size :normal.

There's a related problem with push-buttons and other labelled gadgets: the label is drawn in the face and size specified by \*default-text-style\*, but the overall gadget may end up with an incorrect height (possibly clipping off the top of the label).

I think it's reasonable for label panes always to be :sans-serif and :roman, but in that case push-buttons and other labelled gadgets should also be in :roman face, for consistency. But more importantly, I don't think any of them should ignore the size component of \*default-text-style\*. Here's a relevant sentence from the specification:

>\*default-text-style\* [Constant]
>The default text style used on a CLIM medium if no text style it [is] explicitly specified for the medium when it it [is] created.
http://bauhh.dyndns.org:8000/clim-spec/11-1.html#_570

The following shows the problems:

```
(in-package :clim-user)

(setq *default-text-style* (make-text-style :serif :italic 24))

(define-application-frame labeltest ()
  ()
  (:menu-bar t)
  (:panes
    (p1 (make-pane 'label-pane :label "P1 label"))
    (p2 (make-pane 'push-button :label "P2 button"))
    (p3 (make-pane 'toggle-button :label "P3 toggle"))
    (p4
      (make-pane 'slider
        :orientation :horizontal
        :min-value 0 :max-value 100 :value 50
        :show-value-p t))
    (p5 
      (make-pane 'list-pane
        :value "P5 item 1"
        :items '("P5 item 1" "P5 item 2")))
    (p6 
      (make-pane 'option-pane
        :value "P6 option 1"
        :items '("P6 option 1" "P6 option 2"))))
 (:layouts
    (default
      (spacing (:thickness 20)
        (vertically (:y-spacing 20) p1 p2 p3 p4 p5 p6)))))
               
(defun labeltest ()
  (run-frame-top-level (make-application-frame 'labeltest :width 250)))

(define-labeltest-command (quit :name t :menu t) ()
  (with-application-frame (frame) (frame-exit frame)))

(labeltest)
```

[...]

PR #647, opened 2018-12-30
Handle pointer scroll events in a few gadgets

scymtym:

PR #655, opened 2019-01-02
Label panes and other gadgets take face/size from *default-text-style*

john-a-carroll:
I decided against defaulting gadgets to :roman face - why force this choice on the user? Instead they pick up the face from `*default-text-style*`. The compose-space method on slider-pane needed a bit more care with the minor dimension size to properly take into account the width/height of the value. I also fixed typos in popup-init (`:suggest-` -> `:suggested-`) while I was editing that function.

### DIFF
--- a/Core/clim-core/gadgets.lisp
+++ b/Core/clim-core/gadgets.lisp
@@ -1106,6 +1106,7 @@ and must never be nil.")
     :y-spacing 4))
 
 (defmethod compose-space ((gadget push-button-pane) &key width height)
+  (declare (ignore width height))
   (let ((2*x-spacing (* 2 (pane-x-spacing gadget)))
         (2*y-spacing (* 2 (pane-y-spacing gadget)))
         (2*border-thickness (* 2 *3d-border-thickness*)))
@@ -1661,22 +1662,24 @@ and must never be nil.")
 
 (defmethod compose-space ((pane slider-pane) &key width height)
   (declare (ignore width height))
-  (multiple-value-bind (vw vh)
-      (text-size pane (format-value (gadget-min-value pane) (slider-decimal-places pane)))
-    (setq vw
-      (max vw
-           (text-size pane
-                      (format-value (gadget-max-value pane) (slider-decimal-places pane)))))
-    (let ((minor
-            (if (gadget-show-value-p pane)
-                (* 2 (+ 10.0 (if (eq (gadget-orientation pane) :vertical) vw vh))) ; the value
-                (* 2 (1+ 8.0)))) ; the thingy
-          (major 128))
-      (if (eq (gadget-orientation pane) :vertical)
-          (make-space-requirement :min-width  minor :width  minor
-                                  :min-height major :height major)
-          (make-space-requirement :min-width  major :width  major
-                                  :min-height minor :height minor)))))
+  (let* ((vw
+           (max ; no slider value can be wider than widest of min-value or max-value
+             (text-size pane
+                        (format-value (gadget-min-value pane) (slider-decimal-places pane)))
+             (text-size pane
+                        (format-value (gadget-max-value pane) (slider-decimal-places pane)))))
+         (vh
+           (text-style-ascent (pane-text-style pane) pane))
+         (minor
+           (if (gadget-show-value-p pane)
+               (* 2 (+ 10.0 (if (eq (gadget-orientation pane) :vertical) vw vh))) ; the value
+               (* 2 (1+ 8.0)))) ; the knob
+         (major 128))
+    (if (eq (gadget-orientation pane) :vertical)
+        (make-space-requirement :min-width  minor :width  minor
+                                :min-height major :height major)
+        (make-space-requirement :min-width  major :width  major
+                                :min-height minor :height minor))))
 
 (defmethod handle-event ((pane slider-pane) (event pointer-enter-event))
   (with-slots (armed) pane
--- a/Core/clim-core/gadgets.lisp
+++ b/Core/clim-core/gadgets.lisp
@@ -1662,19 +1662,20 @@ and must never be nil.")
 
 (defmethod compose-space ((pane slider-pane) &key width height)
   (declare (ignore width height))
-  (let* ((vw
-           (max ; no slider value can be wider than widest of min-value or max-value
-             (text-size pane
-                        (format-value (gadget-min-value pane) (slider-decimal-places pane)))
-             (text-size pane
-                        (format-value (gadget-max-value pane) (slider-decimal-places pane)))))
-         (vh
-           (text-style-ascent (pane-text-style pane) pane))
-         (minor
-           (if (gadget-show-value-p pane)
-               (* 2 (+ 10.0 (if (eq (gadget-orientation pane) :vertical) vw vh))) ; the value
-               (* 2 (1+ 8.0)))) ; the knob
-         (major 128))
+  (let* ((value-size (ecase (gadget-orientation pane)
+                       (:horizontal (text-style-ascent (pane-text-style pane) pane))
+                       (:vertical (let* ((dp (slider-decimal-places pane))
+                                         (s1 (format-value (gadget-min-value pane) dp))
+                                         (s2 (format-value (gadget-max-value pane) dp)))
+                                     ;; (over-)estimate, otherwise would have to check all
+                                     ;; possible strings in order to allow for kerning etc
+                                     (* (max (length s1) (length s2))
+                                        (text-size pane #\0))))))
+          (minor (if (gadget-show-value-p pane)
+                     ;; values reflect sizes of controls and spacing in handle-repaint
+                     (* 2 (+ 10.0 value-size)) ; the numeric value
+                     (* 2 (1+ 8.0)))) ; the knob
+          (major 128))
     (if (eq (gadget-orientation pane) :vertical)
         (make-space-requirement :min-width  minor :width  minor
                                 :min-height major :height major)
--- a/Core/clim-core/gadgets.lisp
+++ b/Core/clim-core/gadgets.lisp
@@ -520,7 +520,9 @@
                      :reader slider-number-of-quanta))
   (:documentation "The value is a real number, and default value for orientation is :vertical,
 and must never be nil.")
-  (:default-initargs :orientation :vertical))
+  (:default-initargs
+    :text-style (merge-text-styles (make-text-style :sans-serif nil nil) *default-text-style*)
+    :orientation :vertical))
 
 (defmethod drag-callback ((pane slider) client gadget-id value)
   (declare (ignore client gadget-id))
@@ -1096,7 +1098,7 @@ and must never be nil.")
                       :initarg :show-as-default-p
                       :accessor push-button-show-as-default-p))
   (:default-initargs
-    :text-style (make-text-style :sans-serif nil nil)
+    :text-style (merge-text-styles (make-text-style :sans-serif nil nil) *default-text-style*)
     :background *3d-normal-color*
     :align-x :center
     :align-y :center
@@ -1173,7 +1175,7 @@ and must never be nil.")
                    :initform :some-of) )
   (:default-initargs
     :value nil
-    :text-style (make-text-style :sans-serif nil nil)
+    :text-style (merge-text-styles (make-text-style :sans-serif nil nil) *default-text-style*)
     :align-x :left
     :align-y :center
     :x-spacing 2
@@ -1259,7 +1261,7 @@ and must never be nil.")
                             sheet-leaf-mixin)
   ()
   (:default-initargs
-    :text-style (make-text-style :sans-serif nil nil)
+    :text-style (merge-text-styles (make-text-style :sans-serif nil nil) *default-text-style*)
     :background *3d-normal-color*
     :x-spacing 3
     :y-spacing 2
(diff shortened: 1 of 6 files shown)


Core/clim-core/gadgets.lisp
```diff
@@ -520,7 +520,9 @@
                      :reader slider-number-of-quanta))
   (:documentation "The value is a real number, and default value for orientation is :vertical,
 and must never be nil.")
-  (:default-initargs :orientation :vertical))
+  (:default-initargs
+    :text-style (merge-text-styles (make-text-style :sans-serif nil nil) *default-text-style*)
```
scymtym: Please add a function like `default-gadget-text-style' so this initform doesn't have to repeated as many times.

Core/clim-core/gadgets.lisp
```diff
@@ -1659,13 +1661,22 @@ and must never be nil.")
...
-  (let ((minor (+ 8 (* 2 4) (if (gadget-show-value-p pane) 30 0)))
-        (major 128))
-    (if (eq (gadget-orientation pane) :vertical)
-        (make-space-requirement :min-width  minor :width  minor
-                                :min-height major :height major)
-        (make-space-requirement :min-width  major :width  major
-                                :min-height minor :height minor))))
+  (multiple-value-bind (vw vh)
+      (text-size pane (format-value (gadget-min-value pane) (slider-decimal-places pane)))
+    (setq vw
```
scymtym: `alexandria:maxf` can be used here.

Why is `vh` not updated in the same way?

Core/clim-core/gadgets.lisp
```diff
@@ -1659,13 +1661,22 @@ and must never be nil.")
...
+      (max vw
+           (text-size pane
+                      (format-value (gadget-max-value pane) (slider-decimal-places pane)))))
+    (let ((minor
+            (if (gadget-show-value-p pane)
+                (* 2 (+ 10.0 (if (eq (gadget-orientation pane) :vertical) vw vh))) ; the value
```
scymtym: Please avoid (by defining constants) or explain the new magic numbers.

Core/clim-core/gadgets.lisp
```diff
@@ -2624,10 +2639,9 @@ if INVOKE-CALLBACK is given."))
           ;;       which I could have easily worked around without adding kludges
           ;;       to the scroller-pane..
           (let* ((topmost-pane (if scroll-p
-                                  ;list-pane
                                   (scrolling (:scroll-bar :vertical
-                                              :suggest-height height   ;; Doesn't appear to be working..
-                                              :suggest-width (if scroll-p (+ 30 (bounding-rectangle-width list-pane))))
+                                              :suggested-height height
```
scymtym: Please explain this change in the commit message or even make a separate commit.

Core/clim-core/gadgets.lisp
dkochmanski: 1. `merge-text-styles` is unnecessary here. `make-text-style` will merge with `*default-text-style*` all NIL arguments.
2. shouldn't `(:default-initargs :text-style (make-text-style :sans-serif nil nil))` be supplied as a `basic-gadget` class option? I don't see a need for `default-gadget-text-style`, but maybe I miss something (?)

Core/clim-core/menu.lisp
```diff
@@ -282,8 +282,8 @@ account, and create a list of menu buttons."
 
 ;;; Menu creation from command tables
 
-(defparameter *enabled-text-style*  (make-text-style :sans-serif :roman :normal))
-(defparameter *disabled-text-style* (make-text-style :sans-serif :roman :normal))
+(defparameter *enabled-text-style*  (make-text-style :sans-serif nil nil))
+(defparameter *disabled-text-style* (make-text-style :sans-serif nil nil))
```
dkochmanski: since we have to create a text-style on each invocation *anyway* I think that we should inline them (and remove these special variables). Each of them is used only twice in the same function. We won't need merging text style then as well, because (make-text-style …) fills NIL's with the default one.

john-a-carroll: My reasoning might be wrong, but I think the underlying issue is that while labels and other text associated with gadgets is output in the fully specified style eventually produced by medium-merged-text-style, the internal layout of gadgets (and compose-space etc) is guided by pane-text-style, which might have unspecified components. When the size is unspecified, methods such as text-style-ascent seem to take it as 12. To avoid this mismatch, my commits were intended to fully specify the pane text style from the outset. Another solution could be to simulate the outcome of medium-merged-text-style throughout gadgets.lisp - but that would require very many changes.

Core/clim-core/gadgets.lisp
john-a-carroll: The change is referred to in the commit message: fixed typos in popup-init (`:suggest-` -> :`suggested-`). The effect is so minor I didn't think it deserved a separate commit.

Core/clim-core/gadgets.lisp
john-a-carroll: Maybe this is the wrong solution. See my other reply about the underlying issue.

Core/clim-core/gadgets.lisp
john-a-carroll: Yes, `maxf` is better. Updating `vh` seems like overkill, since there's only a difference for fonts that have digits of varying heights, it only affects the bounding box for horizontal sliders, and in any case the effect is minimal.

Core/clim-core/gadgets.lisp
scymtym: I only saw that in the description of the pull request, not the commit message itself. It is generally useful to include all relevant information in the commit message.

Core/clim-core/gadgets.lisp
john-a-carroll: These values (2, 10.0, 8.0 and 1+) tie in with the slider-pane handle-repaint method, so yes, it would improve readability to make some of them constants and change handle-repaint accordingly.

scymtym: Given that #647 contains some refactoring of the slider gadget, maybe it is fine to leave the magic numbers for now.

Core/clim-core/gadgets.lisp
dkochmanski: is it costful though? if not, then having correct behavior is better.

Core/clim-core/gadgets.lisp
dkochmanski: #647 has been merged.

Core/clim-core/menu.lisp
dkochmanski: Thank you for the explanation, I'll look into the code to have a better understanding of that.

Core/clim-core/gadgets.lisp
john-a-carroll: You're right. Commit https://github.com/McCLIM/McCLIM/pull/655/commits/a695dea5dc52727f1f6099f8fd622a3503981627 includes an improved `vw` and `vh` computation that is correct in all cases.

Core/clim-core/gadgets.lisp
```diff
@@ -1659,8 +1662,19 @@ and must never be nil.")
...
+                        (format-value (gadget-min-value pane) (slider-decimal-places pane)))
+             (text-size pane
+                        (format-value (gadget-max-value pane) (slider-decimal-places pane)))))
+         (vh
+           (text-style-ascent (pane-text-style pane) pane))
+         (minor
+           (if (gadget-show-value-p pane)
+               (* 2 (+ 10.0 (if (eq (gadget-orientation pane) :vertical) vw vh))) ; the value
+               (* 2 (1+ 8.0)))) ; the knob
+         (major 128))
```
dkochmanski: I think that this change (in general) operates on top of invalid assumption that min/max value is the narrowest/widest (especially if we agree, that text-style may be something else than `:fix`).

For instance, let's take range from 0 to 111 - in this case 100 is wider than both (assuming `0` is wider than `1`).
As of vertical panes ascent suggests, that font will never have numbers having pixels /below/ the baseline. I think it is a reasonable assumption, but point still stands.

I think that instead of having vw and vh values this should be rewritten to take into account the worst case scenario using provided clim primitives for that. I think that something in this spirit should be OK:

```
(let* ((value-size (etypecase (gadget-orientation pane)
                     (:vertical (text-style-height (pane-text-style pane) pane))
                     (:horizontal (let* ((dp (slider-decimal-places pane))
                                         (s1 (format-value (gadget-min-value pane) dp))
                                         (s2 (format-value (gadget-max-value pane) dp)))
                                    (* (max (length s1) (length s2))
                                       (text-style-width (pane-text-style pane) pane))))))
       (minor (if (gadget-show-value-p pane)
                  (* 2 (+ 10.0 value-size))
                  (* 2 (1+ 8.0))))
       (major 128))
...)
```

note that in this case compose-space will demand more space, because height is factually `(+ ascent descent)` and width is `(* number-of-characters size-of-letter-m)`. But IMHO this approach is more general than probing for text-size of min/max values.

dkochmanski: also I'd add better comments where these numeric constants come from. 10.0, 8.0, why 1+? why *2? all this is not obvious. I'll look into main part with text-styles for merging tomorrow.

john-a-carroll: Thanks for your thoughtful suggestions. In my testing I hadn't noticed intermediate values being wider than min/max because DejaVu Sans (and it turns out many other fonts) have decimal digits that are all the same width. However, I believe the following deals correctly with variable width digits. It probes each digit individually since in some fonts even the widest digit is significantly narrower than the letter "m". For the other dimension, I used the font ascent not the height because draw-text* in handle-repaint draws with y alignment `:baseline`, not `:bottom`.
```
(defmethod compose-space ((pane slider-pane) &key width height)
  (let* ((value-size (ecase (gadget-orientation pane)
                       (:horizontal (text-style-ascent (pane-text-style pane) pane))
                       (:vertical (let* ((dp (slider-decimal-places pane))
                                         (s1 (format-value (gadget-min-value pane) dp))
                                         (s2 (format-value (gadget-max-value pane) dp)))
                                    (* (max (length s1) (length s2))
                                       (loop for c across "0123456789.-"
                                             maximize (text-size pane c)))))))
          (minor (if (gadget-show-value-p pane)
                     (* 2 (+ 10.0 value-size)) ; the value
                     (* 2 (1+ 8.0)))) ; the knob
          (major 128))
    (if (eq (gadget-orientation pane) :vertical)
        (make-space-requirement :min-width  minor :width  minor
                                :min-height major :height major)
        (make-space-requirement :min-width  major :width  major
                                :min-height minor :height minor))))
```
The 10.0 and 8.0 are used in the same way as in handle-repaint. The 1+ corresponds to the draw-circle* call that includes 1+ and 8.0. The 2 is because sliders are drawn with the same amount of space on one side of the line as the other. These values are more obvious when looked at in the context of the complete source code file.

john-a-carroll: On further reflection, a McCLIM programmer would be ill-advised to use sliders with `:show-value-p t` in conjunction with a proportional font that has non-monospaced digits. If they do, they will get the unpleasant visual effect described in the article [Proportional vs. Monospaced Numbers: When to use which one in order to avoid "Wiggling Labels"](https://blog.usejournal.com/proportional-vs-monospaced-numbers-when-to-use-which-one-in-order-to-avoid-wiggling-labels-e31b1c83e4d0)

dkochmanski: thanks for the interesting link!  regardless, your approach with maximize and text-size looks good to me. ill advised or not, this is not our problem.

regarding magic values: my intention was to suggest, that you should put comment describing where they come from (i.e `magic values taken from handler-repaint`).

Core/clim-core/gadgets.lisp
dkochmanski: this change is superseded. resolving this conversation.

Core/clim-core/gadgets.lisp
dkochmanski: same concern is voiced against new code, resolving this conversation.

Core/clim-core/gadgets.lisp
dkochmanski: please amend commit description, there is no need for a separate commit, but noting it in the commit message makes sense

dkochmanski:
also it is worth noting, that pull request is related to this issue: https://github.com/McCLIM/McCLIM/issues/624

Core/clim-core/gadgets.lisp
dkochmanski: OK, so the problem originated in extensions/fonts module. Please see branch default-text-style-fixes for a commit which fixes issue #624. It also modifies text-style initarg to be put directly in basic-gadget (only once) -- second commit. Please rebase your remaining changes (`compose-space` fixes and `suggested-height` against that branch).

I'm not doing that myself because that would be kind of hijack of the pull request.

Core/clim-core/menu.lisp
dkochmanski: OK, so the problem originated in extensions/fonts module. Please see branch https://github.com/McCLIM/McCLIM/tree/default-text-style-fixes for a commit which fixes issue #624. It also modifies text-style initarg to be put directly in basic-gadget (only once) -- second commit. Please rebase your remaining changes (compose-space fixes and suggested-height against that branch).

I'm not doing that myself because that would be kind of hijack of the pull request. This fix is a direct result of peer reviewing your work, so all credit belongs to you.

dkochmanski:
nb (regarding last push) - please rebase changes so there are no spuriuos "Merge branch" commits.

Core/clim-core/menu.lisp
john-a-carroll: Thanks for finding and fixing the problem! I don't need any credit, and I'm getting myself thoroughly confused in trying to rebase the remaining changes - probably due to the fact that I'm going down with flu. So please go ahead and make those changes yourself (as in commit https://github.com/McCLIM/McCLIM/pull/655/commits/1ace64b6facef6fe2ec50ed9f1412a2ed3820aba and changing :suggest- -> :suggested- x2)

dkochmanski:
Alright, I'll go ahead then and make a new PR. Get healthy soon!

PR #659, opened 2019-01-03
text-style and compose-space fixes

dkochmanski:
This supersedes the previous PR https://github.com/McCLIM/McCLIM/pull/655.

### DIFF
--- a/Backends/common/fonts.lisp
+++ b/Backends/common/fonts.lisp
@@ -187,10 +187,13 @@ of letters specified in a separate kerning-table."))
 
 (declaim (inline climb:normalize-font-size))
 (defun climb:normalize-font-size (size)
-  (if (numberp size)
-      (round (max size 2))
-      (or (getf +font-sizes+ size size)
-          (getf +font-sizes+ :normal))))
+  (cond ((numberp size)
+         (round (max size 2)))
+        ((null size)
+         (getf +font-sizes+ (text-style-size *default-text-style*)))
+        (T
+         (or (getf +font-sizes+ size nil)
+             (error "~s is not a valid text style size!" size)))))
 
 (defgeneric climb:text-style-to-font (port text-style))
 
--- a/Core/clim-core/gadgets.lisp
+++ b/Core/clim-core/gadgets.lisp
@@ -1104,6 +1104,7 @@ and must never be nil.")
     :y-spacing 4))
 
 (defmethod compose-space ((gadget push-button-pane) &key width height)
+  (declare (ignore width height))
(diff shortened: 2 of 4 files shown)


Backends/common/fonts.lisp
```diff
@@ -187,10 +187,13 @@ of letters specified in a separate kerning-table."))
...
-      (round (max size 2))
-      (or (getf +font-sizes+ size size)
-          (getf +font-sizes+ :normal))))
+  (cond ((numberp size)
+         (round (max size 2)))
+        ((null size)
+         (getf +font-sizes+ (text-style-size *default-text-style*)))
+        (T
+         (or (getf +font-sizes+ size size)
+             (error "~s is not a valid text style size!" size)))))
```
scymtym: How does this work? In this leg of the `cond`, `size` is not null, thus `(getf +font-sizes+ size size)` should not return null.

Core/clim-core/gadgets.lisp
```diff
@@ -2635,8 +2635,8 @@ if INVOKE-CALLBACK is given."))
           (let* ((topmost-pane (if scroll-p
                                   ;list-pane
                                   (scrolling (:scroll-bar :vertical
-                                              :suggest-height height   ;; Doesn't appear to be working..
-                                              :suggest-width (if scroll-p (+ 30 (bounding-rectangle-width list-pane))))
+                                              :suggested-height height   ;; Doesn't appear to be working..
```
scymtym: Is the comment still true after fixing the typo?

dkochmanski: I don't know how to verify that (I didn't investigate what the actual code does – I've just picked the change from the original commit). Do you want me to remove the comment without testing?

Backends/common/fonts.lisp
dkochmanski: you are right, fixed in the force-pushed version.

scymtym: `getf` defaults to `nil` so the third argument could be omitted. Not super important.

Core/clim-core/gadgets.lisp
scymtym: Maybe replace with "TODO investigate whether these arguments actually have an effect"

Backends/common/fonts.lisp
dkochmanski: I want to put emphasis on returning NIL if not found

Core/clim-core/gadgets.lisp
dkochmanski: added comment as suggested

Issue #641, opened 2018-12-25
Images rendered as part of an output record on exactly a half-pixel position gets visual artefacts


lokedhs:
When creating an output record that draws an image (either using `DRAW-DESIGN*` or `DRAW-RECTANGLE*` with an ink set to the image, if the x or y coordinates of the output record is exactly on a half-pixel (for example, 10.5 but not 10.49999 or 10.50001) the rendered image displays some visual artefacts.

An example of the issue can be found here: https://github.com/lokedhs/clim-test/blob/master/image.lisp

The example code renders the same image three times, on the following Y coordinates: 10, 400.5 and 800.49999. It can be seen that only the second image is drawn incorrectly.

Here is a screenshot of the issue:

![screenshot from 2018-12-25 15-16-34](https://user-images.githubusercontent.com/1028755/50415816-34f31180-0858-11e9-8780-bad0335bf1c1.png)

PR #661, opened 2019-01-05
In DESIGN-INK, use consistent rounding for coordinates

lokedhs:
Previously, ROUND was used to ensure that the arguments to DESIGN-INK
were integers. This caused a problem in some situations when the
coordinates have a fractional part of 0.5. In such cases, ROUND will
either round up or down depending on whether the integer part is odd
or even.

Fixes https://github.com/McCLIM/McCLIM/issues/641

### DIFF
--- a/Core/clim-basic/pattern.lisp
+++ b/Core/clim-basic/pattern.lisp
@@ -388,7 +388,9 @@ Returns a pattern representing this file."
          (transformation (transformed-design-transformation effective-pattern))
          (inv-tr (invert-transformation transformation)))
     (multiple-value-bind (x y) (transform-position inv-tr x y)
-      (design-ink source-pattern (round x) (round y)))))
+      ;; It is important to not use ROUND here, since when the
+      ;; fractional part is exactly 0.5, we get wrong dimensions -- loke 2019-01-06
+      (design-ink source-pattern (floor (+ x 0.5)) (floor (+ y 0.5))))))
 
 (defmethod %collapse-pattern ((pattern transformed-pattern))
   (with-bounding-rectangle* (x1 y1 x2 y2) pattern


Core/clim-basic/pattern.lisp
```diff
@@ -388,7 +388,7 @@ Returns a pattern representing this file."
          (transformation (transformed-design-transformation effective-pattern))
          (inv-tr (invert-transformation transformation)))
     (multiple-value-bind (x y) (transform-position inv-tr x y)
-      (design-ink source-pattern (round x) (round y)))))
+      (design-ink source-pattern (floor (+ x 0.5)) (floor (+ y 0.5))))))
```
scymtym: Maybe add a short comment here (despite the commit messages containing the detailed explanation).

lokedhs: Committed an update.

dkochmanski: This is too lenghy comment for such message. Let it something like this:
```
;; We don't use round because it doesn't round consistently and makes images
;; to have one extra pixel for coordinates having 0.5 fractions. -- loke 2018-01-05
```

also squash comment with the commit, they are one change. adding nick and date to comment is useful too.

lokedhs:
I can rewrite this tomorrow, but if someone could explain to me how to squash commits, I'd be grateful.

On my own repositories I usually just use -f but that's not recommended when more than one person uses it.

scymtym:
> but if someone could explain to me how to squash commits

In this case (assuming you are not using [magit](https://magit.vc/)):
* Check out the branch (should be two commits ahead of master)
* ``git rebase --interactive HEAD~2``
* This will open an editor containing the "rebase plan"
* Change the action in front of the second commit to "fixup"
* Save and exit the editor
* Inspect the result with ``git log``

lokedhs:
@scymtym thank you, I believe I have done what was asked, although I'm not sure how to read the git log that resulted from the rebase operation.

dkochmanski:
this is not right, you have no created 10 commits instead of one.
try this:
```
git reset --hard image-coordinate
# change the file again
git commit -m "your commit message"
git push --force
```

regarding ordinary way to rebase (I think this train already left, this looks like a mess):
```
git fetch
git rebase -i origin/master
# you may reorder commits here and squash them (file should be commented)
git push --force
```

Core/clim-basic/pattern.lisp
```diff
@@ -388,7 +388,10 @@ Returns a pattern representing this file."
          (transformation (transformed-design-transformation effective-pattern))
          (inv-tr (invert-transformation transformation)))
     (multiple-value-bind (x y) (transform-position inv-tr x y)
-      (design-ink source-pattern (round x) (round y)))))
+      ;; It is important to not use ROUND here, since when the
+      ;; fractional part is exactly 0.5, we get wrong dimensions.
+      ;; loke 2019-01-06
```
dkochmanski: make it `-- loke 2019-01-06` for consistency with other comments. it may be in the same line as the comment.

PR #662, opened 2019-01-05
Feedback for edges in draggable graph example

scymtym:
Update edges in addition to the node while dragging.

### DIFF
--- a/Core/clim-core/graph-formatting.lisp
+++ b/Core/clim-core/graph-formatting.lisp
@@ -491,8 +491,8 @@ Assumes that GENERATE-GRAPH-NODES has generated only nodes up to the cutoff-dept
   ((stream)
    (arc-drawer)
    (arc-drawing-options)
-   (from-node :initarg :from-node)
-   (to-node :initarg :to-node)))
+   (from-node :initarg :from-node :reader from-node)
+   (to-node :initarg :to-node :reader to-node)))
 
 (defun layout-edges (graph node stream arc-drawer arc-drawing-options)
   (dolist (k (graph-node-children node))
--- a/Examples/draggable-graph.lisp
+++ b/Examples/draggable-graph.lisp
@@ -24,49 +24,37 @@
 
 (define-application-frame draggable-graph-demo () ()
   (:pane (make-pane 'application-pane
-		    :width :compute
-		    :height :compute
-		    :display-function 'generate-graph
-		    :display-time t)))
+                    :width :compute
+                    :height :compute
+                    :display-function 'generate-graph
+                    :display-time t)))
 
 (defun generate-graph (frame pane)
   (declare (ignore frame))
   (format-graph-from-roots
    (list (find-class 'number))
    (lambda (object stream)
-     (present (class-name object)
-	      (presentation-type-of object)
-	      :stream stream))
+     (present (class-name object) (presentation-type-of object) :stream stream))
    #'c2mop:class-direct-subclasses
    :stream pane))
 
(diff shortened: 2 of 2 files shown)


lokedhs:
Extra indentation happens automatically in Emacs if you have `slime-indentation` enabled.

As for the loop keywords, I don't think most of the McCLIM code uses the colon prefix before the keywords.

scymtym:
* Removed keywords in `loop` form
* Indented `do` clause

Examples/draggable-graph.lisp
```diff
@@ -24,49 +24,39 @@
...
 (defun generate-graph (frame pane)
   (declare (ignore frame))
   (format-graph-from-roots
    (list (find-class 'number))
    (lambda (object stream)
-     (present (class-name object)
-	      (presentation-type-of object)
-	      :stream stream))
+     (present (class-name object) (presentation-type-of object)
+              :stream stream))
```
dkochmanski: I don't understand why did you change one parameter per line style to "two parameters in the first and one remaining in a separate line", I'm fine with having all in one line or putting everything in separate lines but such mixed style doesn't convince me (especially here where we don't have that many arguments,  I could understand two obligatory arguments in first line and keyword arguments in next (one by one) if there were more of them.

Examples/draggable-graph.lisp
```diff
@@ -24,49 +24,39 @@
...
-       (declare (ignore parent))       
-       (push edge edges))
-     (slot-value node 'climi::edges-to))
-    edges))
+  (append (alexandria:hash-table-values (slot-value node 'climi::edges-from))
+          (alexandria:hash-table-values (slot-value node 'climi::edges-to))))
+
+(defun node-and-edges-region (node edges)
+  (stupid-copy-rectangle
+   (reduce #'region-union edges :initial-value node)))
```
dkochmanski: please put `(reduce …` in the same line as `stupid-copy-rectangle`

Examples/draggable-graph.lisp
```diff
@@ -85,42 +75,39 @@
...
+  (let* ((stream *standard-output*)
+         (node-record (find-graph-node record))
+         (edge-records (node-edges node-record))
+         (graph-record (output-record-parent node-record))
+         (erase-region))
+    (assert (typep graph-record 'graph-output-record))
+    (erase-output-record node-record stream)
+    (drag-output-record
+     stream node-record
+     :feedback (lambda (record stream old-x old-y x y mode)
```
dkochmanski: please put `stream` in the same line as the function. In general please avoid style, where first function argument gets its own line. I don't recall this style being used anywhere in the source code and I'm not a fan of it myself. If lambda function is nested too deep for feedback use flet.

Core/clim-core/graph-formatting.lisp
```diff
@@ -491,8 +491,10 @@ Assumes that GENERATE-GRAPH-NODES has generated only nodes up to the cutoff-dept
   ((stream)
    (arc-drawer)
    (arc-drawing-options)
-   (from-node :initarg :from-node)
-   (to-node :initarg :to-node)))
+   (from-node :initarg :from-node
+              :reader from-node)
+   (to-node :initarg :to-node
+            :reader to-node)))
```
dkochmanski: please put reader arguments in the same line (this style is mixed across the codebase, but this line is short enough to make it reasonable).

scymtym:
Reformatted.

dkochmanski:
looks good to me, thank you.

PR #664, opened 2019-01-05
Fix recent regressions

dkochmanski:


Backends/CLX/fonts.lisp
```diff
@@ -72,13 +72,10 @@
...
                (text-style-components text-style)
-
-             (setf face   (or face :roman)
-                   family (or family :fix)
-                   size   (max 2 (climb:normalize-font-size size)))
-
+             (setf size (max 2 (climb:normalize-font-size size)))
              (let ((display (clim-clx::clx-port-display port)))
                (find-and-make-xlib-face display family face size)))))
+      (setf text-style (merge-text-styles text-style *default-text-style*))
```
scymtym: How expensive is this? Should we only do it if one of the components is unspecific?

dkochmanski: the answer is: not very expensive. If we want to change that (with looking if the component is not specified), I'd go with

```
           (setf face   (or face (text-style-face *default-text-style*))
                   family (or family (text-style-family *default-text-style*))
                   size   (climb:normalize-font-size size))
```

instead of merge style.

dkochmanski: as a side note: cleaner solution would be having text-style with wildcard components having a separate class. but this seems to be a broad change and I'm not very interested in making it now.

dkochmanski: I've fixed it (also I've fixed unreported issue when *default-text-style* size was numeric value.

nsrahmad:
Works fine on CCL.

But I can't load it with current master and SBCL 1.4.15, it drops to ldb. quickloading with `:verbose t` reveals the last compiling forms are:
```
; file: /home/nsrahmad/common-lisp/McCLIM/Backends/common/fonts.lisp
; in: DECLAIM (INLINE NORMALIZE-FONT-SIZE)
;     (DECLAIM (INLINE CLIM-BACKEND:NORMALIZE-FONT-SIZE))
; --> EVAL-WHEN 
; ==>
;   (SB-C::%PROCLAIM '(INLINE CLIM-BACKEND:NORMALIZE-FONT-SIZE)
;                    (SB-C:SOURCE-LOCATION))
; 
; caught STYLE-WARNING:
;   Proclaiming CLIM-BACKEND:NORMALIZE-FONT-SIZE to be INLINE, but 1 call to it was
;   previously compiled. A declaration of NOTINLINE at the call site will eliminate
;   this warning, as will proclaiming and defining the function before its first
;   potential use.

; compiling (DEFUN NORMALIZE-FONT-SIZE ...)
; compiling (DECLAIM (INLINE PARSE-TEXT-STYLE*))
; compiling (DEFUN PARSE-TEXT-STYLE* ...)......

; file: /home/nsrahmad/common-lisp/McCLIM/Backends/common/fonts.lisp
; in: DEFUN PARSE-TEXT-STYLE*
;     (CLIM-BACKEND:PARSE-TEXT-STYLE*
;      (APPLY #'CLIM:MAKE-TEXT-STYLE CLIM-INTERNALS::STYLE))
; 
; note: *INLINE-EXPANSION-LIMIT* (200) was exceeded, probably trying to
;   inline a recursive function.
```
while in inferior-lisp buffer:
```lisp
Heap exhausted during garbage collection: 0 bytes available, 64 requested.
Gen  Boxed Unboxed   LgBox LgUnbox  Pin       Alloc     Waste        Trig      WP GCs Mem-age
fatal error encountered in SBCL pid 3024(tid 0x7f7e6e80f700):
GC invariant lost, file "gencgc.c", line 474


Error opening /dev/tty: No such device or address
Welcome to LDB, a low-level debugger for the Lisp runtime environment.
ldb> 
```

dkochmanski:
@nsrahmad I've pushed a fix. parse-text-style* can't be declared inline because it calls itself.

PR #666, opened 2019-01-06
cleanup: don't muffle style-warnings

dkochmanski:
While I get the sentiment that &key and &optional are something what will stay
in codebase (because that's how CLIM is specified) so we don't need
style-warnings about that, these local declarations muffle *all* style warnings
what is plain wrong.

These are legitimate style warnings anyway and I see nothing wrong with the fact
it is signalled. Commit doesn't introduce any functional changes, just removes
(locally (declare (sb-ext:muffle-conditions style-warning)) …) declarations.

### DIFF
--- a/Core/clim-basic/stream-output.lisp
+++ b/Core/clim-basic/stream-output.lisp
@@ -397,22 +397,20 @@ STREAM-STRING-WIDTH will be called."))
     (declare (ignore y))
     (zerop x)))
 
-(locally
-    (declare #+sbcl (sb-ext:muffle-conditions style-warning))
-  (defmacro with-room-for-graphics ((&optional (stream t)
-                                               &rest arguments
-                                               &key (first-quadrant t)
-                                               height
-                                               (move-cursor t)
-                                               (record-type ''standard-sequence-output-record))
-                                     &body body)
-    (declare (ignore first-quadrant height move-cursor record-type))
-    (let ((cont (gensym "CONT."))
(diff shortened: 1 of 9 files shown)


dkochmanski:
This, on the other hand, is clutter in code (which in my opinion is worse).

slyrus:
To each their own, but I consider the "clutter" in the code an annotation that, yeah, this is lame. But not something that needs to pop up in my SLIME REPL every time I rebuild McCLIM. Admittedly, there are so many warnings/notes now that this additional output hardly seems noticeable but, like I said, it would be great if we had a warning-free (other than those we explicitly acknowledge and muffle) output one day.

dkochmanski:
rebased on current master

lokedhs:
Perhaps controversial, but it wouldn't be impossible to define a macro that is basically a defun with the appropriate declarations.

dkochmanski:
I think that the right developers to ask about having a separate condition for this style warning are sbcl developers , then we could muffle it from asd (as around method over compile file). Otherwise then that I find these parts perfectly normal and fact that sbcl doesn't like them doesn't mean we want to add sbcl-specific muffling in there.

namely this could be put in mcclim.asd file (if only this style warning had its own class).
```
#+sbcl
(defmethod perform :around ((op compile-op) (s (eql (find-system :mcclim))))
  (handler-bind ((sb-ext:optional-key-style-warning #'muffle-warning))
    (call-next-method)))
```

PR #670, opened 2019-01-07
input-refactor: another batch of changes

dkochmanski:


dkochmanski:
rebased on current master

Core/clim-basic/recording.lisp
```diff
@@ -1937,57 +1934,57 @@ add output recording facilities. It is not instantiable."))
...
-(defmethod stream-add-character-output
-    ((stream standard-output-recording-stream)
-     character text-style width height baseline)
-  (add-character-output-to-text-record
-   (stream-text-output-record stream text-style)
-   character text-style width height baseline))
+(defmethod stream-close-text-output-record ((stream standard-output-recording-stream))
+  (when-let ((record (stream-current-text-output-record stream)))
+    (setf (stream-current-text-output-record stream) nil)
+    #|record stream-current-cursor-position to (end-x record) - already done|#
```
nsrahmad: @dkochmanski is it a leftover note to self?

dkochmanski: no, this comment was already in this method. since I don't understand it I didn't remove it.

dkochmanski:
pushed change removes comment I've made because change (that we don't have :around method but a proper specialization) and not eager drawing makes it unnecessary. We don't need to draw directly on medium anynome from standard output.

Issue #678, opened 2019-01-08
tabify-buffer-region in Libraries/Drei/base.lisp is slightly wrong


wbooze:
I tried M-x Tabify Region many times over in Climacs, and some words looked almost like written together.
But upon closer inspection one could observe the Cursor disappear in between the words.
I thought maybe it's a font-issue, or some issue with presentation types. I looked at the file via emacs in a terminal and there was no hint of something being wrong. The almost together written words had a real space in there. 
The Bug also made the Climacs window split upon completion.

It's some off by one error. No resolution yet!

nsrahmad:
Could you try replacing function definition of `maybe-update-recordings` in `McCLIM/Libraries/Drei/views.lisp` with below version:
```lisp
(defun maybe-update-recordings (stream tabify)
  (with-accessors ((space-width recorded-space-width)
                   (tab-width recorded-tab-width)
                   (recorded-stream recorded-stream)) tabify
    (unless (eq stream recorded-stream)
      ;; Update the recorded values.
      ;; Width of space is usually zero, what is really needed here is advance-width
      ;; --NSR 2019-01-21
      (setf space-width (climb:font-glyph-dx (climb:text-style-to-font
					      (find-port)
					      (medium-text-style stream))
					     (char-code #\Space))
            tab-width (stream-character-width stream #\Tab)
            recorded-stream stream))))
```

Issue #672, opened 2019-01-07
method tab-width and cohorts in Libraries/Drei/views.lisp 


wbooze:
### the above methods use tab characters without checking for if use-tabs is actually set or not, so tabify gives me mixed files with tab-chars and space-chars, i don't think that's the way it is intended really.
so my solution is below:

```
(defgeneric space-width (stream tabify)
  (:documentation "Return the width of a space character on <stream> in device units (most likely pixels).")
  (:method ((stream extended-output-stream) (tabify tabify-mixin))
           (maybe-update-recordings stream tabify)
           (with-accessors ((recorded-space-width recorded-space-width)) tabify
                           recorded-space-width)))
(defgeneric tab-width (stream tabify)
  (:documentation "Return the width of a tab character on <stream> in device units (most likely pixels).")
  (:method ((stream extended-output-stream) (tabify tabify-mixin))
           (with-accessors ((tab-space-count tab-space-count) 
                                               (recorded-tab-width recorded-tab-width) (use-tabs use-tabs)) tabify
                           (if use-tabs
                               recorded-tab-width
                               (* (if tab-space-count tab-space-count 8) (space-width stream tabify))))))

(defgeneric next-tab-stop (stream tabify x)
  (:documentation "Return the distance to the next tab-stop after 'x' on <stream> in device units (most likely pixels).")
  (:method ((stream extended-output-stream) (tabify tabify-mixin) x)
           (with-accessors ((tab-space-count tab-space-count) 
                                               (tab-stops tab-stops) (use-tabs use-tabs)) tabify
                           (flet ((round-up (x width)
                                            (if (= 0 width)
                                                (* tab-space-count (space-width stream tabify))
                                              (- width (mod x width)))))
                                 (if use-tabs
                                     (when tab-stops
                                       (let ((next (find-if (lambda (pos) (> pos x)) tab-stops)))
                                         (or (and next (- next x)) (round-up x (tab-width stream tabify)))))
                                   (round-up x
                                                        (* (if tab-space-count tab-space-count 8) (space-width stream tabify))))))))

(defgeneric (setf tab-stop-columns) (column-list tabify)
  (:documentation "Set the TAB-STOPS of view at the character column offsets in <column-list>.")
  (:method (column-list (tabify tabify-mixin))
           (with-accessors ((tab-stops tab-stops) (recorded-stream recorded-stream)) tabify
    (setf tab-stops
          (and column-list
[...]

PR #679, opened 2019-01-08
Issues #678 and #672 related changes in Libraries/Drei

wbooze:
Issues #678 and #672 related changes in Libraries/Drei/base.lisp, Libraries/Drei/views.lisp

Apps/Listener/listener.lisp small adjustements for the listener appearance and word wrap behaviour:
interactor pane in listener.lisp needs a :end-of-line :wrap as action but it also depends on the 
fields :text-margin and/or :margin

I also have some file fix-sbcl.lisp which kinda is missing from master!


Lisp-Dep/fix-sbcl.lisp
```diff
@@ -0,0 +1,52 @@
+(in-package :cl-user)
+
```
dkochmanski: what do you need this file for? we use closer-mop for mop functionality. please remove this file from the PR.

Apps/Listener/listener.lisp
```diff
@@ -164,12 +164,14 @@
...
 (defun print-listener-prompt (stream frame)
   (declare (ignore frame))
-  (with-output-as-presentation (stream *package* 'package :single-box t)
-    (print-package-name stream))
-  (princ "> " stream)
-  (let ((h (- (bounding-rectangle-height (stream-output-history stream))
-              (bounding-rectangle-height (or (pane-viewport stream) stream)))))
-    (scroll-extent stream 0 (max 0 h))))
+  (with-output-as-presentation 
+   (stream *package* 'package :single-box t)
```
dkochmanski: please keep optinos argument in the same line as with-output-as-presentation macro (like it was before)

Apps/Listener/listener.lisp
```diff
@@ -164,12 +164,14 @@
...
+   (with-drawing-options (stream :text-face :roman text-size :normal)
+			 (print-package-name stream))
```
dkochmanski: this is not well indented since `(print-package-name stream)` is macro's body not functions argument

dkochmanski: please explain why this change is necessary

Libraries/Drei/base.lisp
```diff
@@ -275,7 +275,7 @@ constituent character of the line."
            until (end-of-buffer-p mark2)
            while (or (eql object #\Space) (eql object #\Tab))
            do (incf indentation
-                    (if (eql (object-after mark2) #\Tab) tab-width 1))
+                    (if (eql (object-after mark2) #\Tab) tab-width (/ tab-width 8)))
```
dkochmanski: what does it mean? if `tab-width` is 4 and character is not a tab we indent by 0.5? I don't understand how this could be right, if character is not a tab then indent is 1 like it was before.

Libraries/Drei/base.lisp
```diff
@@ -300,7 +300,7 @@ characters in the region between offset1 and offset2."
        for i from line-start-offset below offset
        do (incf column (if (eql (buffer-object buffer i) #\Tab)
                            (- tab-width (mod column tab-width))
-                           1))
+                           (/ tab-width 8)))
```
dkochmanski: ditto

Libraries/Drei/base.lisp
```diff
@@ -786,7 +786,7 @@ in the region delimited by mark1 and mark2."))
                                               offset
                                               tab-width))
                (count (- tab-width (mod column tab-width))))
-          (delete-buffer-range buffer offset 1)
+          (delete-buffer-range buffer offset count)
           (loop repeat count
```
dkochmanski: if this change is necessary please make it a separate commit and explain why it is necessary.

Libraries/Drei/views.lisp
```diff
@@ -81,37 +81,46 @@ If empty, tabs every TAB-WIDTH are assumed.")))
...
   (:documentation "Return the width of a tab character on
 `stream' in device units (most likely pixels).")
   (:method ((stream extended-output-stream) (tabify tabify-mixin))
-    (if (tab-space-count tabify)
-        (* (tab-space-count tabify) (space-width stream tabify))
-        (recorded-tab-width tabify))))
+	   (with-accessors ((tab-space-count tab-space-count) (recorded-tab-width recorded-tab-width) (use-tabs use-tabs)) tabify
+			   (if use-tabs
+			       recorded-tab-width
+				 (* (if tab-space-count tab-space-count 8) (space-width stream tabify))))))
```
dkochmanski: iden't here and above are awful, please fix that and explain in a commit why this change is necessary.

Libraries/Drei/views.lisp
```diff
@@ -81,37 +81,46 @@ If empty, tabs every TAB-WIDTH are assumed.")))
...
+					    (if (= 0 width)
+						(* tab-space-count (space-width stream tabify))
+					      (- width (mod x width)))))
+				 
+				 (if use-tabs
+				     (when tab-stops
+				       (let ((next (find-if (lambda (pos) (> pos x)) tab-stops)))
+					 (or (and next (- next x)) (round-up x (tab-width stream tabify)))))
+				   (round-up x
+					     (* (if tab-space-count tab-space-count 8) (space-width stream tabify))))))))
```
dkochmanski: ditto

Libraries/Drei/views.lisp
```diff
@@ -81,37 +81,46 @@ If empty, tabs every TAB-WIDTH are assumed.")))
...
-    (setf (tab-stops tabify) 
+	   (with-accessors ((tab-stops tab-stops) (recorded-stream recorded-stream)) tabify
+    (setf tab-stops
 	  (and column-list
-	       (sort (mapcar (lambda (col) (* col (space-width (recorded-stream tabify) tabify)))
-			     column-list) 
-		     #'<)))))
+	       (sort (mapcar (lambda (col) (* col (space-width recorded-stream tabify)))
+			     column-list)
+		     #'<))))))
```
dkochmanski: ditto

Libraries/Drei/base.lisp
wbooze: tab-space-count is 8 for a tab, other than that i don't know how or which other metrics go into the computation (like fonts or whatever), since there are many levels of abstraction like presentations/views/fonts etc..
the only thing we are sure is the tab-space-count. So what is 1 ?  is it 1 character or is it 1 in device units etc...
I thought that (/ tab-width 8) would give a more portable expression or even (/ tab-width tab-space-count) for that matter that's all, if you think it's really not needed or wrong then delete it please

Libraries/Drei/base.lisp
wbooze: same as above :) sorry

Libraries/Drei/base.lisp
wbooze: please have a look into the issue #678 (probably  #672 too) , there I already described it.

PR #682, opened 2019-01-09
Gadget views

admich:
I start to implement the gadget views to be used with accepting-values as described in Franz Manual section 8.6.13 (CLIM 2.2 specification).

For this:

1. Specification conformity: some slots defined in abstract class gadget specification in McCLIM was implemented only in the concrete gadget class. With this PR  this slots are defined inthe abstract class.

2. The gadget views are defined in such a way that they have the same slots of the corresponding abstract gadget class. (This is obtain with a macro that use MOP).

3. For each gadget view is implemeted the method accept-present-default

4. In the Accepting Value Test demo is inserted a test for the gadget views.

### DIFF
--- a/Core/clim-basic/views.lisp
+++ b/Core/clim-basic/views.lisp
@@ -44,34 +44,6 @@
 (defclass pointer-documentation-view (textual-view)
   ())
 
-;;; Views described in the Franz User manual (CLIM 2.2)...
-
-(defclass toggle-button-view (gadget-view)
-  ())
-
-(defclass push-button-view (gadget-view)
-  ())
-
-(defclass radio-box-view (gadget-view)
-  ())
-
-(defclass check-box-view (gadget-view)
-  ())
-
-(defclass slider-view (gadget-view)
-  ())
-
-(defclass text-field-view (gadget-dialog-view)
-  ((width :accessor width :initarg :width :initform nil)))
-
-(defclass text-editor-view (gadget-view)
-  ())
-
-(defclass list-pane-view (gadget-view)
-  ((visible-items :accessor visible-items :initarg :visible-items)))
-
-(defclass option-pane-view (gadget-view)
-  ())
 
 (defparameter +textual-view+ (make-instance 'textual-view))
 
@@ -87,21 +59,6 @@
 
 (defparameter +pointer-documentation-view+ (make-instance 'pointer-documentation-view))
 
-(defparameter +toggle-button-view+ (make-instance 'toggle-button-view))
-
-(defparameter +push-button-view+ (make-instance 'push-button-view))
-
-(defparameter +radio-box-view+ (make-instance 'radio-box-view))
-
-(defparameter +slider-view+ (make-instance 'slider-view))
-
-(defparameter +text-field-view+ (make-instance 'text-field-view))
-
-(defparameter +text-editor-view+ (make-instance 'text-editor-view))
-
-(defparameter +list-pane-view+ (make-instance 'list-pane-view))
-
-(defparameter +option-pane-view+ (make-instance 'option-pane-view))
 
 (defmethod stream-default-view (stream)
   (declare (ignore stream))
--- a/Core/clim-core/dialog-views.lisp
+++ b/Core/clim-core/dialog-views.lisp
@@ -251,6 +251,31 @@
 						 present-p
 						 query-identifier)))))
 
+;;;; text-editor-view
+
+(define-presentation-method accept-present-default
+    ((type string) stream (view text-editor-view) default default-supplied-p
+     present-p query-identifier)
+  (unless default-supplied-p
+    (setf default ""))
+  (let ((gadget (make-gadget-pane-from-view view stream
+                                            :value default
+                                            :value-changed-callback
+                                            (%standard-value-changed-callback query-identifier))))
+    (updating-output (stream
+                      :cache-value t    ; don't redisplay
+                      :unique-id query-identifier
+                      :record-type 'accepting-values-record)
+
+                     (with-output-as-presentation (stream query-identifier 'selectable-query
+                                                          :single-box t)
+                       (surrounding-output-with-border (stream :shape :rounded
+                                                               :radius 3 :background clim:+background-ink+
+                                                               :foreground clim:+foreground-ink+
+                                                               :move-cursor t)
+                                                       (with-output-as-gadget (stream)
+                                                         gadget))))))
+
 ;;; A gadget that's not in the spec but which would  be useful.
 (defclass pop-up-menu-view (gadget-dialog-view)
   ()
--- a/Core/clim-core/dialog-views.lisp
+++ b/Core/clim-core/dialog-views.lisp
@@ -205,6 +205,34 @@
                                 :value-changed-callback
                                 (%standard-value-changed-callback query-identifier)))
 
+;;;; slider-view
+
+(define-presentation-method accept-present-default
+    ((type real) stream (view slider-view) default default-supplied-p
+     present-p query-identifier)
+  (make-output-record-from-view view stream query-identifier
+                                :value default
+                                :max-value high ; high: presentation parameter
+                                :min-value low ; low: presentation parameter
+                                :show-value-p t
+                                :value-changed-callback (%standard-value-changed-callback query-identifier)))
+
+(define-presentation-method accept-present-default
+    ((type integer) stream (view slider-view) default default-supplied-p
+     present-p query-identifier)
+  (make-output-record-from-view view stream query-identifier
+                                :value default
+                                :max-value high ; presentation parameter
+                                :min-value low ; presentation parameter
+                                :show-value-p t
+                                :number-of-quanta (- high low)
+                                :value-changed-callback
+                                (lambda (pane item)
+                                  (declare (ignore pane))
+                                  (let ((value (round item)))
+                                    (throw-object-ptype `(com-change-query ,query-identifier ,value)
+                                                        '(command :command-table accept-values))))))
+
 ;;; A gadget that's not in the spec but which would  be useful.
 (defclass pop-up-menu-view (gadget-dialog-view)
   ()
--- a/Core/clim-core/dialog-views.lisp
+++ b/Core/clim-core/dialog-views.lisp
@@ -115,6 +115,68 @@
                                 :value-changed-callback
                                 (%standard-value-changed-callback query-identifier)))
 
+;;; radio-box-view
+
+(define-presentation-method accept-present-default
+    ((type completion) stream (view radio-box-view) default default-supplied-p
+     present-p query-identifier)
+  (let ((buttons (loop for choice in sequence collect
+                      (make-pane 'toggle-button
+                                 :indicator-type :one-of
+                                 :id (funcall value-key choice)
+                                 :label (funcall name-key choice)))))
+    (make-output-record-from-view view stream query-identifier
+                                  :choices buttons
+                                  :current-selection (find default buttons :test test :key #'gadget-id)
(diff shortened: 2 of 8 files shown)


Core/clim-basic/views.lisp
```diff
@@ -44,34 +44,6 @@
...
-  ())
-
-(defclass check-box-view (gadget-view)
-  ())
-
-(defclass slider-view (gadget-view)
-  ())
-
-(defclass text-field-view (gadget-dialog-view)
-  ((width :accessor width :initarg :width :initform nil)))
```
scymtym: Please explain in the commit message:

* Why can the `width` slot be removed? (The `define-gadget-view` macro looks at `text-field` which doesn't have such a slot)
* Why is the superclass now `gadget-dialog-view` for all view classes when it previously was `gadget-view` for most view classes?

package.lisp
```diff
@@ -1626,6 +1626,8 @@
    #:+push-button-view+                 ;constant
    #:radio-box-view                     ;class
    #:+radio-box-view+                   ;class
+   #:check-box-view                     ;class
+   #:+check-box-view+                   ;class
```
scymtym: Please split this into two commits.

Core/clim-core/gadgets.lisp
```diff
@@ -713,7 +713,14 @@ and must never be nil.")
 ;;; 30.4.9 The abstract text-editor Gadget
 
 (defclass text-editor (text-field)
-  ()
+  ((ncolumns :reader text-editor-ncolumns
```
scymtym: I'm not sure this will work. The concrete class uses an editor substrate but doesn't have the slots. So this will create a detached copy of the information.

What is the goal here? Just having the initargs accepted?

Core/clim-core/dialog-views.lisp
```diff
@@ -70,6 +70,37 @@
                                          :default default
                                          :default-type default-type))
 
+(defun make-gadget-pane-from-view (view stream query-identifier &rest initargs)
+  (let* ((class-view (type-of view))
```
scymtym: Please use `class-of`

Core/clim-core/dialog-views.lisp
```diff
@@ -70,6 +70,37 @@
...
+         (class-gadget (intern (subseq (symbol-name class-view) 0 (position #\- (symbol-name class-view) :from-end t)) 'climi))
+         (slots (c2mop:class-slots (find-class class-view))))
+    (loop for slot in slots when (slot-boundp view (c2mop:slot-definition-name slot)) do
+         (push (slot-value view (c2mop:slot-definition-name slot)) initargs)
+         (push (car (c2mop:slot-definition-initargs slot)) initargs))
```
scymtym: I would like a different solution for this. How about the following: a new generic function `gadget-for-view` return two values: a gadget class and a plist of initargs and values. A method specialized on e.g. `gadget-view' can then do something similar to this code as a default.

To get rid of the `intern` and symbol name mangling at runtime: the macro that generates the view classes could put the name of the corresponding gadget class into a class-allocated slot of the view class.

Core/clim-core/dialog-views.lisp
```diff
@@ -70,6 +70,37 @@
...
+                      :cache-value t    ; don't redisplay
+                      :unique-id query-identifier
+                      :record-type 'accepting-values-record)
+        (with-look-and-feel-realization ((frame-manager *application-frame*) *application-frame*)
+          (with-output-as-gadget (stream)
+            (apply #'make-pane class-gadget initargs))))))
+
+;;; toggle-button-view
+
+(define-presentation-method accept-present-default
```
scymtym: I'm out of my depth here, but you seem to know what you are doing.

Core/clim-core/dialog-views.lisp
```diff
@@ -101,6 +101,85 @@
                                     :value default
                                     :value-changed-callback #'value-changed-callback))))
 
+;;; radio-box-view
```
scymtym: There is a typo in the first line of the commit message.

Core/clim-core/dialog-views.lisp
```diff
@@ -101,6 +101,85 @@
...
+                              :value-changed-callback
+                              (lambda (pane item)
+                                (let ((value (gadget-id item)))
+                                  (throw-object-ptype `(com-change-query ,query-identifier ,value) '(command :command-table accept-values))))
+                              initargs)))
+          gadget)))))
+
+;;; check-box-view
+
+(define-presentation-method accept-present-default
```
scymtym: Please extract the identical parts of this method and the previous method into a helper function.

Core/clim-core/dialog-views.lisp
```diff
@@ -101,6 +101,85 @@
...
+    ((type completion) stream (view option-pane-view) default default-supplied-p
+     present-p query-identifier)
+  (flet ((generate-callback (query-identifier)
+           (lambda (pane item)
+             (declare (ignore pane))
+             (when *accepting-values-stream*
+
+               (let ((query (find query-identifier (queries *accepting-values-stream*)
+                                  :key #'query-identifier :test #'equal)))
+                 (when query
```
scymtym: We typically use `when-let` nowadays.

Core/clim-core/dialog-views.lisp
```diff
@@ -180,6 +180,42 @@
                                 :items sequence
                                 :value-changed-callback (generate-callback query-identifier))))
 
+;;; list-pane-view
```
scymtym: "for for" in the commit message.

Please rephrase the final sentence of the commit message. I can't understand it.

Core/clim-core/dialog-views.lisp
```diff
@@ -216,6 +216,41 @@
...
+             (throw-object-ptype `(com-change-query ,query-identifier ,item)
+                                 command-ptype)))
+      (make-gadget-pane-from-view view stream query-identifier
+                                  :value default
+                                  :max-value high ; high: presentation parameter
+                                  :min-value low ; low: presentation parameter
+                                  :show-value-p t
+                                  :value-changed-callback #'value-changed-callback))))
+
+(define-presentation-method accept-present-default
```
scymtym: Please extract the identical code of the two methods into a helper function.

Core/clim-core/dialog.lisp
```diff
@@ -535,24 +535,6 @@ is called. Used to determine if any editing has been done by user")))
 
 (defparameter *no-default-cache-value* (cons nil nil))
 
-;;; Hack until more views / dialog gadgets are defined.
```
scymtym: Please explain the reasoning more clearly in the commit message.

Is the method not needed because of the fallback to `textual-dialog-view`? Then the bug is irrelevant.

Or is the fallback making things worse but unavoidable because of the bug?

Core/clim-core/dialog-views.lisp
```diff
@@ -251,6 +251,36 @@
                                   :value-changed-callback #'value-changed-callback))))
 
 
+;;;; text-editor-view
```
scymtym: "implemetation" typo in the commit message

Core/clim-core/dialog-views.lisp
```diff
@@ -251,6 +251,36 @@
...
+
+(define-presentation-method accept-present-default
+    ((type string) stream (view text-editor-view) default default-supplied-p
+     present-p query-identifier)
+  (unless default-supplied-p
+    (setf default ""))
+  (let* ((slots (c2mop:class-slots (find-class 'text-editor-view)))
```
scymtym: I think this should be done using a generic function as described in an earlier annotation.

In case that is not possible: 
* `(class-slots (find-class 'text-editor-view))' should not be called multiple times. 
* Doesn't `class-slots` require the class to be finalized?

Examples/accepting-values-test.lisp
```diff
@@ -61,7 +61,10 @@
     (present '(com-accept-popup) 'command :stream pane)
     (fresh-line pane)
     (present '(com-accepting-with-list-pane-view) 'command :stream pane)
-    (fresh-line pane)))
+    (fresh-line pane)
+    (present '(com-accepting-with-gadgets) 'command :stream pane)
+    (fresh-line pane)
+    ))
```
scymtym: Please put the closing parens on the previous line.

scymtym:
@admich Btw, thank you for the very cool changes.

admich:
@scymtym thank you for your review. I need some time to fix with all your suggestions.

scymtym:
Sure. Thank you for working on this.

admich:
At the end I'm ready: I pushed the requested changes. Sorry for the delay but I was busy with my work.
You will find some comments to your suggestions below each one.

Core/clim-core/gadgets.lisp
admich: I known, but I think that if the manual said that this gadget have these slots then the abstract class must have them. However they work. If you change `:ncolumns` or `:nline` in the view the gadget change the dimensions.

Core/clim-core/dialog-views.lisp
admich: Now I don't use `type-of` or `class-of` because the name of the gadget is stored in a class-allocated slot of the view instance.

Core/clim-core/dialog-views.lisp
admich: I try to follow your suggestion, I hope that now it is better.

Core/clim-core/dialog-views.lisp
admich: Ok, I've change it.

Core/clim-core/dialog-views.lisp
admich: Ok, I fixed it

Core/clim-core/dialog-views.lisp
admich: Ok, I fixed it.

Core/clim-core/dialog-views.lisp
admich: I think now it is better.

Core/clim-core/dialog.lisp
admich: I don't remove the method `text-field-view` anymore. I only moved it in `Core/clim-core/dialog-views.lisp`. Anyway I'm not sure it works, I tried with different `:width` but I didn't see any difference.

Core/clim-core/dialog-views.lisp
admich: Ok, I fixed it.

Core/clim-core/dialog-views.lisp
admich: I've changed it.

Examples/accepting-values-test.lisp
admich: Ok, fixed.

scymtym:
> At the end I'm ready: I pushed the requested changes. Sorry for the delay but I was busy with my work.

No worries. Thank you for putting in all that work.

> You will find some comments to your suggestions below each one.

Great.

I would like to merge this now. Are you OK with me rebasing and fixing a few more things like excess whitespace and then merging?

admich:
No problem for me you can rebasing and fixing.

admich:
It's ok your version you should merge it. 

It's similar to what Franz CLIM code do (https://gitlab.common-lisp.net/mcclim/gramps-clim2) but better because you check wich initarg the view can accept. (PS: I think that mcclim can learn a lot about dialog and gadget from Franz CLIM code).

I really didn't understand what is the fault  with my code but I trust you.

admich:
Ye I'm ok

scymtym:
@admich OK, merged. Thank you very much for your contribution.

PR #685, opened 2019-01-13
drawing-tests: improve and refactor; 

josrr:
Please review it, I think it cover all requested changes.

### DIFF
--- a/Examples/drawing-tests.lisp
+++ b/Examples/drawing-tests.lisp
@@ -36,8 +36,7 @@
    (signal-condition-p :initform nil)
    (current-selection :initform nil)
    (application-frame-backend :initform :clx-ttf)
-   (frames :initform (make-hash-table))
-   (application-frame-recording-p :initform nil))
+   (frames :initform (make-hash-table)))
   (:panes
    (category-test-pane
     (horizontally (:equalize-height t)
@@ -125,12 +124,7 @@
                   (2/3 (clim-extensions:lowering ()
                          (make-pane 'text-field-pane
                                     :name 'server-path
-                                    :value ":custom")))))))
-          (labelling (:label "Recording")
-            (clim:with-radio-box (:orientation :vertical
-                                  :value-changed-callback #'%update-application-frame-recording-option)
-              "yes"
-              (clim:radio-box-current-selection "no")))))
+                                    :value ":custom")))))))))
       (spacing (:thickness 6)
         (vertically ()
           (make-pane 'push-button
@@ -177,25 +171,27 @@
    (description-pane
     (spacing (:thickness 3)
       (clim-extensions:lowering ()
-        (scrolling (:scroll-bar :vertical :height 200)
+        (scrolling (:scroll-bar :vertical)
           (make-pane 'application-pane
-                     :name 'description))))))
+                     :name 'description
+                     :end-of-line-action :wrap*
+                     :end-of-page-action :scroll))))))
   (:layouts
    (default
     (spacing (:thickness 3)
-      (vertically ()
-        (1/6 category-test-pane)
+      (vertically (:height (* 3/2 *height*))
+        (1/7 category-test-pane)
         (:fill (horizontally (:min-width (* 15/8 *width*))
                  (1/3 options-pane)
                  (:fill (vertically ()
                           (:fill (spacing (:thickness 3)
                                    (clim-extensions:lowering ()
-                                     backend-pane)))
-                          (1/5 description-pane))))))))
+                                     backend-pane)))))))
+        (1/7 description-pane))))
    (side-by-side
     (spacing (:thickness 3)
-      (vertically ()
-        (1/6 category-test-pane)
+      (vertically (:height (* 3/2 *height*))
+        (1/7 category-test-pane)
         (:fill (horizontally (:min-width (* 3 *width*))
                  (1/6 options-pane)
                  (:fill (vertically ()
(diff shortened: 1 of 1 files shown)


dkochmanski:
First remarks from testing (without looking at the code):
1. "Use separate application frame" starts a new frame,but there is no output if "Recording" is "no" (it seems to draw output and then redraw empty output on top of it)
2. Category/Tests lists should have their own full row or at least more space
3. I can navigate through tests with keyboard arrows but not through categories. I'd like to have a behavior, that when I press "down" arrow after the last test in a category the next category is selected (and first test from it)
4. Even if recording is set to "no", benchmarks print "… operations/s; recording: yes"  -- now that I look into that I see two radio buttons with "recording", that should be only one.
5. I'd like to have the output console (where "writing …" is shown) to also indicate, that a new test has been selected when that occurs manually for preview
6. extensions of "Print All" (at least for png" have names in capital letters "PNG"
7. when I switch side-by-side "Backend"s pane has no test shown (I need to press test again to see the content)

Examples/drawing-tests.lisp
```diff
@@ -1,100 +1,356 @@
...
 (defparameter *width* 500)
 (defparameter *height* 700)
 (defparameter *border-width* 5)
 
-(defstruct drawing-test name description drawer)
+(defstruct drawing-test category name description display-function)
 
-(defmacro define-drawing-test (name arglist description &body body)
+(defun drawing-test-keyname (category name)
+  (concatenate 'string category "-" name))
```
dkochmanski: is there a need for this to be a string? hashtable access could be better if you had used `alexandria:symbolicate` and have symbols as keys

Examples/drawing-tests.lisp
```diff
@@ -1,100 +1,356 @@
...
-                         :drawer (lambda ,arglist ,@body))))
+  (alexandria:with-gensyms (g-category g-name)
+    `(let ((,g-name ,name)
+           (,g-category ,category))
+       (setf (gethash (drawing-test-keyname ,g-category ,g-name) *drawing-tests*)
+             (make-drawing-test :category ,g-category
+                                :name ,g-name
+                                :description ,description
+                                :display-function (lambda (,frame ,stream ,@arglist) ,@body)))
+       (pushnew ,g-category *drawing-tests-categories* :test #'string-equal))))
```
dkochmanski: ditto, maybe category could be a symbol?

Examples/drawing-tests.lisp
```diff
@@ -1,100 +1,356 @@
...
+
+(defun get-display (category name)
+  (let ((drawing-test (gethash (drawing-test-keyname category name) *drawing-tests*)))
+    (when drawing-test
```
dkochmanski: `(when-let ((drawing-test …)) …)`

dkochmanski: also this function seems not to be used. am I right? if so, please remove it.

Examples/drawing-tests.lisp
```diff
@@ -1,100 +1,356 @@
...
+      (drawing-test-display-function drawing-test))))
 
 (define-application-frame drawing-tests ()
   ((recording-p :initform t)
    (signal-condition-p :initform nil)
-   (current-selection :initform nil))
+   (current-selection :initform nil)
+   (application-frame-backend :initform :clx-ttf)
+   (frames :initform (make-hash-table))
+   (application-frame-recording-p :initform nil))
```
dkochmanski: why do we have two different settings for recording?

dkochmanski:
backend pane width is too small to host "usual" descriptions with M-q wrapping in sources. Since window is not resizable this looks bad (we have line breaks just after fresh soft line break.

dkochmanski:
ah, I see, categories are navigated with M-left M-right. why not M-up M-down?

dkochmanski:
these changes could be split into more fine-grained commits (introducing functionality step by step) - it would be easier to review.

Examples/drawing-tests.lisp
josrr: The second recording option is specific to the frame created with the "Use separate application frame" button.

josrr:
About first remarks:

 1. Is `(redisplay-frame-pane :around ((frame application-frame) pane &key force-p))`
      method that is clearing the pane after drawing it. The only way I've found to fix that is using `clime:never-repaint-background-mixin`.
 2. Done.
 3. Done.
 4. The text changes if you use the first recording option. The second option below label "Application frame" is for the application frame created by the "Use separate application frame" button.
 5. The description of the test is printed; some tests have no description.
 6. I will do it.
 7. I don't see that behavior.
 8. I will do it.

Examples/drawing-tests.lisp
josrr: No, you are right. I will change it.

josrr:
About the points 1 and 7 of "first remarks":  the content of the panels is deleted because they are of the `application-pane` class that is subclass of `clime:always-repaint-background-mixin`; I'm sure it's the `(handle-repaint :before (region (always paint-repaint-mix-background)))` method that draws a rectangle with the background ink of the panel.

About 8:   activate-gadget and deactivate-gadget dont have any effect in `text-field-pane` objects.

josrr:
I think I'm done with all the required changes, please, when you have some time review it.

dkochmanski:
This looks good to me. I still have some small issues but they are *not* related to the scope of this work. Thank you. Please claim a bounty on the bounty source. Good work!

PR #686, opened 2019-01-13
LAYOUT-EXAMPLES: Fix Unintended Variable Capture

PuercoPop:
Although the variable `esop` is not used in the body of the only place where the macro is being used.


scymtym:
Thank you. Merged with minor modifications.

Issue #693, opened 2019-01-27
Proposal to change the way presentation highlighting looks


lokedhs:
The current implementation simply draws a black rectangle around the presentation. In this bug report I propose a slightly different version:

- The coordinates of the bounding box are adjusted so that they (in most cases) cover the entire presentation. The old version makes the box one pixel smaller, which can cause the actual content to slow outside the box in some cases.
- The background colour is slightly brighter (or, slightly darker if the background is already white)
- The border is brighter than before.

![screenshot from 2019-01-27 20-15-24](https://user-images.githubusercontent.com/1028755/51800827-7409ed80-2270-11e9-9ca6-fbc0813e33c2.png)

The above screenshot shows a subexpression of a mathematical formula being highlighted.

![screenshot from 2019-01-27 20-16-08](https://user-images.githubusercontent.com/1028755/51800828-779d7480-2270-11e9-9085-8e0861461ab5.png)

In this screenshot, a URL is highlighted.

lokedhs:
To test the implementation, evaluate the following code:

```
(defun highlight-colour (colour)
  (let ((delta 0.1))
    (multiple-value-bind (i h s)
        (clim:color-ihs colour)
      (clim:make-ihs-color (if (> (+ i delta) 1)
                               (- i delta)
                               (+ i delta))
                           h
                           s))))

(defmethod clim:highlight-output-record ((record clim:output-record) stream state)
  (clim:with-identity-transformation (stream)
    (multiple-value-bind (px1 py1 px2 py2)
        (clim:bounding-rectangle* record)
      (let ((x1 (truncate (+ px1 0.5)))
            (y1 (truncate (+ py1 0.5)))
            (x2 (truncate px2))
            (y2 (truncate py2)))
        (let ((medium (clim:sheet-medium stream)))
          (ecase state
            (:highlight
             (let ((bg-colour (highlight-colour (clim:medium-background medium))))
               (clim:draw-rectangle* medium x1 y1 x2 y2 :filled t :ink bg-colour)
               (clim:draw-rectangle* medium x1 y1 x2 y2 :filled nil :ink (clim:make-rgb-color 0.7 0.7 0.7))
               (clim:replay-output-record (clim:stream-output-history stream) stream (clim:make-rectangle* x1 y1 x2 y2))))
            (:unhighlight
             (clim:repaint-sheet stream (clim:make-rectangle* x1 y1 (1+ x2) (1+ y2))))))))))
```

PR #694, opened 2019-01-27
Better visual appearance of presentation highlights

lokedhs:
Fixes https://github.com/McCLIM/McCLIM/issues/693


lokedhs:
There are rendering problems with certain elements. This needs a bit more research.

PR #697, opened 2019-01-31
Fix path shapes

dkochmanski:


dkochmanski:
please pull again. I think I was wrong with decrementing coordinates by 0.5 after all.

dkochmanski:
Technically speaking it is not an off-by-one error but an approximation problem. Imagine rectangle:

`0.5 0.5 -> 9.5 9.5`

it is approximated with a rectangle on screen:

`1 1` (and distance width/height is 9) ` -> `9 9`

because our coordinate rounding in CLX is `(lambda (e) (floor (+ e 0.5)))`. Now lets draw the rectangle border (outline) so it is guaranteed to fit in the bounding rectangle like for highlight:

`1.5 1.5 -> 8.5 8.5`

that gives us lines
`2 2 -> 2 9`
`2 2 -> 9 2`
`2 9 -> 9 9`
`9 2 -> 9 9`

due to the same rounding routine.

as I've mentioned in "highlight" comment, we could fix that with drawing rectangle with functional design (which is not implemented yet).

dkochmanski:
eventually we could use
`(draw-design (region-difference (make-rectangle …) (make-rectangle 1+ 1+ 1- 1-)))` and that should give us a similar result as with the functional design.

lokedhs:
Looks good now.

PR #699, opened 2019-02-02
clim-stream-pane: specify default dimensions

### LINKED ISSUE #47: guided-tour examples
Verify which Guide-Tour examples from the manual doesn't work and fix them (for instance hello-world does load clim and clim-clx, but it should load mcclim system).

dkochmanski:
X protocol can't handle windows of width/height = 0. It can handle them when
size is 1x1, but this doesn't make much sense as a default. We provide defaults
for 100x100. Note, that stream-width/stream-height are somewhat inconsistent
with default width/height intiargs despite some visible effort to make them as
an extension of said mechanism. Fixes #47.

### DIFF
--- a/Core/clim-core/panes.lisp
+++ b/Core/clim-core/panes.lisp
@@ -2519,8 +2519,8 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
    (user-min-height :accessor %pane-user-min-height)
    (user-max-height :accessor %pane-user-max-height)
    ;; size required by the stream
-   (stream-width :initform 0 :accessor stream-width)
-   (stream-height :initform 0 :accessor stream-height))
+   (stream-width :initform 100 :accessor stream-width)
+   (stream-height :initform 100 :accessor stream-height))
   (:documentation
    "This class implements a pane that supports the CLIM graphics,
     extended input and output, and output recording protocols."))


dkochmanski:
note to myself: when working on this PR add a fix for #742.

kc-:
> stream-width/stream-height are somewhat inconsistent

Here's an annotation in the spac talking about how useful the zero defaults for space-requirements are: http://bauhh.dyndns.org:8000/clim-spec/29-3.html#_6589

> Could you explain a bit more [...] how the stream width and height interact with the width and height of the X window?
> My expectation is that the stream width and height are turned into space requirements

That's right: `run-frame-top-level` calls `%realize-mirror`, which calls `compose-space`. That's also where the fault lies (if it is a fault): #47 talks about the hello-world example of the guided tour fails for `clim-stream-pane`, but works for `basic-pane`. The difference is in the `compose-space` method, which, for `clim-stream-pane`, uses the `stream-output-history` to determine a space hint. No records in the history means that the result is an empty rectangle.

```
(defmethod compose-space ((pane clim-stream-pane) &key width height)
  (declare (ignorable width height))
  (let* ((w (bounding-rectangle-max-x (stream-output-history pane)))
         (h (bounding-rectangle-max-y (stream-output-history pane)))
         (width (max w (stream-width pane)))
         (height (max h (stream-height pane))))
    (make-space-requirement
     :min-width (clamp w 0 width)
     :width width
     :max-width +fill+
     :min-height (clamp h 0 height)
     :height height
     :max-height +fill+)))
```

For `basic-pane`, the specialization is on `pane`, and the code reads:
```
(defmethod compose-space ((pane pane) &key width height)
  (make-space-requirement :width (or width 200)
			  :height (or height 200)))
```


This discussion is not the first time for magic default literals being put into or removed from the code. As an example, here's a commit removing default `width` and `height`arguments from `realize-mirror-aux`: a468fc64, and the corresponding reintroduction, four days later: 9dedf6bd.


If I had to decide what to do, I'd normalize the default value for minimal stream size between all the code paths, and I'd do it in pane.lisp, not in the clx backend. My 2ct.

dkochmanski:
@kc- do you mean that this solution is OK in your opinion and should be merged? or some further changes are required? 

@scymtym is this explanation good enough for you?

kc-:
@dkochmanski Yes, I think the solution is OK, because my opinion is that having sensible defaults for new frames should be part of the framework, not the backend.

I'm not sure whether it is complete for all code paths, and I'd also suggest using the same width and height default values everywhere.

scymtym:
@dkochmanski It would be good to have some of this discussion reflected as a comment in the code or the commit message so the next person looking at the code will be made aware of its history.

kc-:
Here's an attempt at normalizing the defaults: f370eac1.

dkochmanski:
superseded by #800.

Issue #704, opened 2019-02-03
Scigraph: error on dialog that change data symbol


admich:
If a user right click on the legend of a graph and choose from the menu "Change Data Symbol" a dialog starts but this error is reported:

> error while parsing arguments to NIL DESTRUCTURING-BIND:
>   unknown keyword: :PRINTER; expected one of :NAME-KEY, :DOCUMENTATION-KEY, :PARTIAL-COMPLETERS, :DESCRIPTION
>    [Condition of type SB-KERNEL::DEFMACRO-LAMBDA-LIST-BROKEN-KEY-LIST-ERROR]
> 
> Restarts:
>  0: [ABORT] ABORT
>  1: [TRY-AGAIN] Try executing the command COM-POP-EDIT-DATASET again.
>  2: [ABORT] Return to application command loop.
>  3: [SCIGRAPH-TOP-LEVEL] Abort to SCIGRAPH Top Level
>  4: [ABORT] abort thread (#<THREAD "Frame Top Level" RUNNING {1003DA6223}>)
> 
> Backtrace:
>   0: ((LAMBDA (#:TYPE1) :IN "/home/admich/.roswell/local-projects/admich/McCLIM/Core/clim-core/presentation-defs.lisp") ((MEMBER 0 1 2 3 4 ...) :NAME-KEY PRINC-TO-STRING :PRINTER PRESENT-LINE-STYLE :HIGHLI..
>   1: (EXPAND-PRESENTATION-TYPE-ABBREVIATION-1 ((MEMBER 0 1 2 3 4 ...) :NAME-KEY PRINC-TO-STRING :PRINTER PRESENT-LINE-STYLE :HIGHLIGHTER ...) NIL)
>   2: (EXPAND-PRESENTATION-TYPE-ABBREVIATION ((MEMBER 0 1 2 3 4 ...) :NAME-KEY PRINC-TO-STRING :PRINTER PRESENT-LINE-STYLE :HIGHLIGHTER ...) NIL)
>   3: (CLIM:ACCEPT DASH-PATTERN :VIEW (RADIO-BOX-VIEW :ORIENTATION :HORIZONTAL) :STREAM #<CLIM-INTERNALS::ACCEPTING-VALUES-STREAM {100235ABA3}> :PROMPT "Line Style" :DEFAULT 0 :DISPLAY-DEFAULT NIL :QUERY-ID..
[...]

PR #706, opened 2019-02-10
Scigraph dialogs

admich:
With this PR the Scigraph dialogs works and now it is possible to change the line type and the symbol of the data with dialogs.

This solve issue #704 

To try it:
1. load scigraph
2. open scigraph demo `(gr:make-demo-frame)` 
3. right click on one of the two data presentation on the legend of the graph and choose from the menu "Change Data Symbols..."

For the moment in the dialog the line types are printed as number and not graphically as expected from Scigraph code. To improve this it will be necessary in the future:
1. Use `:printer` option of completion presentation in `accept-present-default` methods.
2.  Implement the possibiliy to have for gadget labels not only string but also graphics (pixmap for examples)

### DIFF
--- a/Apps/Scigraph/dwim/package.lisp
+++ b/Apps/Scigraph/dwim/package.lisp
@@ -41,7 +41,8 @@ advised of the possiblity of such damages.
                 #:frame-manager
                 #:find-frame-manager
                 #:suggest
-                #:expression)
+                #:expression
+                #:accept)
   (:shadow #:interactive-stream-p
 
            #:menu-choose
@@ -54,7 +55,6 @@ advised of the possiblity of such damages.
            #:bounding-rectangle*
            #:redisplay
            #:redisplayable-format
-           #:accept
            #:accepting-values
            #:accept-values
            #:accept-variable-values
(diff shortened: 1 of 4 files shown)


slyrus:
Paraphrasing @dkochmanski , there are a few things going on here:

1. encapsulating-streams fixes
2. CLIM 2.2 feature implementation
3. scigraph cleanup

I'd prefer to see these split up into separate PRs/commits. The encapsulating-stream stuff seems pretty non-controversial, although it suggests that we might want to be a bit more systematic about making sure that the protocol specified in C.1.1 is fully supported.

I don't have a strong opinion about the CLIM 2.2 stuff, but others may.

And I'm all for the scigraph cleanup, subject to the above.

Finally, yet it's a trivial function, but we probably should just re-write highlight-completion-choice instead of slurping it in from the CLIM-TOS stuff.

dkochmanski:
I think you mean @scymtym. I do agree, that this PR should be split in three PRs.

As of copying CLIM-TOS code, I see no problem with that (unless there are technical issues or we could benefit from such rewrite) - licenses are compatible this way.

Regarding CLIM 2.2 incorporation, before we start having these options we could have a chapter in the manual which lists differences from the CLIM 2.0 specification.

admich:
Following yours suggestions, I will close this PR and I will open three different PRs.

slyrus:
Ah, right @scymtym made those suggestions. @admich thanks for breaking this up. Looking forward to seeing the other PRs.

PR #707, opened 2019-02-11
clim2.2 presentations options.

admich:
Following the suggetion of @slyrus I split PR #706 in three part. This is the first part about clim 2.2 presentation options. These options are required by Schigraph as reported in issue #704. 

This PR only introduce the options but not implements their functionality. 

@slyrus also complained about `highlight-completion-choice` function that it is copied from CLIM-TOS. The function is still there  because @dkochmanski was ok with this, but if you want to remove it, it is not a problem. Until McCLIM don't use `highlighter` options in `accept-present-default` methods we can also use `nil` as default value for `highlighter`.

Finally: if McCLIM don't want implement clim2.2 features it is also ok but it will be necessary to remove the options in Scigraph code (maybe with a compatibility function `define-presentation-type-abbreviation` in DWIM package).

### DIFF
--- a/Core/clim-core/presentation-defs.lisp
+++ b/Core/clim-core/presentation-defs.lisp
@@ -1752,7 +1752,14 @@ protocol retrieving gestures from a provided string."))
   (defconstant +completion-options+
     '((name-key 'default-completion-name-key)
       documentation-key
-      (partial-completers '(#\Space)))))
+      (partial-completers '(#\Space))
+      (printer #'write-token)
+      (highlighter #'highlight-completion-choice))))
+
+;;;; This function is copied from CLIM Franz code
+(defun highlight-completion-choice (continuation object stream)
+  (with-text-face (stream :bold)
+    (funcall continuation object stream)))
 
 (define-presentation-type completion (sequence
                                       &key (test 'eql) (value-key 'identity))
@@ -1826,7 +1833,9 @@ protocol retrieving gestures from a provided string."))
   (make-presentation-type-specifier `(completion ,elements)
                                     :name-key name-key
                                     :documentation-key documentation-key
-                                    :partial-completers partial-completers)
+                                    :partial-completers partial-completers
+                                    :printer printer
+                                    :highlighter highlighter)
   :options #.+completion-options+)
 
 (define-presentation-type-abbreviation member-sequence (sequence
@@ -1835,7 +1844,9 @@ protocol retrieving gestures from a provided string."))
    `(completion ,sequence ,@(and testp `(:test ,test)))
    :name-key name-key
    :documentation-key documentation-key
-   :partial-completers partial-completers)
+   :partial-completers partial-completers
+   :printer printer
+   :highlighter highlighter)
   :options #.+completion-options+)
 
 (defun member-alist-value-key (element)
@@ -1858,10 +1869,14 @@ protocol retrieving gestures from a provided string."))
                 :value-key member-alist-value-key)
    :name-key name-key
    :documentation-key documentation-key
-   :partial-completers partial-completers)
+   :partial-completers partial-completers
+   :printer printer
+   :highlighter highlighter)
   :options ((name-key 'default-completion-name-key)
             (documentation-key 'member-alist-doc-key)
-            (partial-completers '(#\Space))))
+            (partial-completers '(#\Space))
+            (printer #'write-token)
+            (highlighter #'highlight-completion-choice)))
 
 (define-presentation-type subset-completion (sequence
                                              &key (test 'eql)
@@ -1870,7 +1885,9 @@ protocol retrieving gestures from a provided string."))
             documentation-key
             (partial-completers '(#\Space))
             (separator #\,)
-            (echo-space t))
+            (echo-space t)
+            (printer #'write-token)
+            (highlighter #'highlight-completion-choice))
   :inherit-from t)
 
 (define-presentation-method presentation-typep (object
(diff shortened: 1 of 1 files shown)


nsrahmad:
Wait! most of the changes in CLIM 2.2 are also present in Symbolics Genera CLIM 2.0 guide 
(Available at http://www.bitsavers.org/pdf/symbolics/software/genera_8/Common_Lisp_Interface_Manager__CLIM__Release_2.0.pdf ) . 

This particular change can be seen in below screenshot:
clim-2

slyrus:
@nsrahmad given that,  what should we do here? commit these changes and annotate the CLIM spec?

nsrahmad:
From page 1 of `mcclim.pdf` manual:
> This manual documents McCLIM which is a mostly complete implementation
of the CLIM 2.0 specification and its revision 2.2.

So merging these changes should be okay. We already have many changes in McCLIM from CLIM 2.2:
```
[nsrahmad@archlinux Core]$ grep -ir "Franz"
clim-core/gadgets.lisp:;; - :CHOICES initarg to RADIO-BOX and CHECK-BOX is from Franz' user
clim-core/bordered-output.lisp:  ;; The Franz User guide implies that &key isn't needed.
clim-core/frames.lisp:;;; From Franz Users Guide
clim-core/dialog.lisp:using  the method on p. 305 of the Franz CLIM user guide, or by using  the
clim-core/commands.lisp:;;; Franz user manual says that this slot is setf-able
clim-core/presentation-defs.lisp:;;; auto-activate is described in the Franz user guide; it controls whether an
clim-core/presentation-defs.lisp:;;; According to the Franz User's guide, the destination object is
clim-core/dialog-views.lisp:;;; Views described in the Franz User manual (CLIM 2.2). Sec. 8.6.13
clim-core/presentations.lisp:        ;; without saying what that is!  Fortunately, the Franz CLIM user guide
clim-core/incremental-redisplay.lisp:;;; The Franz user guide says that updating-output does
clim-core/incremental-redisplay.lisp:     ;; pane argument found in the Franz User Guide.
clim-basic/graphics.lisp:;;; the Franz user guide is to be believed). What a mess. I think we
clim-basic/ports.lisp:  (:documentation "Sets the cursor associated with SHEET. CURSOR is a symbol, as described in the Franz user's guide."))
clim-basic/recording.lisp:;;; From the Franz CLIM user's guide but not in the spec... clearly necessary.
[nsrahmad@archlinux Core]$ 
````
Adding a comment like above in the source should be enough. No need to annotate the spec since annotations are for clarifications of that particular version.

I also recall one issue where these two versions conflicted, and CLIM 2.2 was implemented in McCLIM. (See https://github.com/McCLIM/McCLIM/issues/477 )

**Comment by dkochmanski** [comment]
imho all incorporated changes from franz manual should be put in our documentation. We supply CLIM spec and McCLIM manual. If we extend CLIM package such changes should end up as modifications to CLIM spec shipped with McCLIM.

Also I wouldn't hold to Franz 2.2 unless it makes sense in *some particular situation we encounter*. Also in mcclim.pdf (I'm taking it from introduction.txi):

```
This manual documents McCLIM @value{MCCLIMVERSION} which is a mostly
complete implementation of the CLIM 2.0 specification and its revision
2.2. To our knowledge version~2.2 of the CLIM specification is only
documented in the ``CLIM 2 User's Guide'' by Franz. While that document
is not a formal specification, it does contain many cleanups and is
often clearer than the official specification; on the other hand, the
original specification is a useful reference. This manual will note
where McCLIM has followed the 2.2 API.
```

**Comment by slyrus** [comment]
@dkochmanski @nsrahmad are you OK with me merging this as is?

File: Core/clim-core/presentation-defs.lisp [hunk 1]
```diff
@@ -1752,7 +1752,14 @@ protocol retrieving gestures from a provided string."))
   (defconstant +completion-options+
     '((name-key 'default-completion-name-key)
       documentation-key
-      (partial-completers '(#\Space)))))
+      (partial-completers '(#\Space))
+      (printer #'write-token)
+      (highlighter #'highlight-completion-choice))))
+
+;;;; This function is copied from CLIM Franz code
+(defun highlight-completion-choice (continuation object stream)
```
pouar: You forgot to wrap `defun highlight-completion-choice` in an `eval-when` as this is failing to compile

pouar: ok, that might not be it, as I still keep getting
```
; caught ERROR:
;   (during macroexpansion of (%DEFINE-PRESENTATION-TYPE COMPLETION ...))
;   The function CLIM-INTERNALS::HIGHLIGHT-COMPLETION-CHOICE is undefined.
;   It is defined earlier in the file but is not available at compile-time.
```

pouar: ok, that was it, but I still had my flatpak build in my directory without the eval-when and was confusing asdf when I tried to build outside of flatpak. So my second comment was my fault, but my first comment still applies.

pouar: Want me to submit a pr with the fix? I can do that if you want.

pouar: I think you guys are doing a pretty good job developing McCLIM btw, and you guys are probably a lot better at it than I am. I'm still just learning how to use the toolkit.

dkochmanski: Hey, thank you for analyzing and reporting that issue. Yes, please make a pull request with a fix.

PR #708, opened 2019-02-11
Encapsulating stream

admich:
Following the suggetion of @slyrus I split PR #706 in three part. This is the second  part about encapsulating stream.

In this PR I add some methods and function to encapsulating-stream. This is not a systematic check of the encapsulating stream as suggested by @slyrus but I add only the function needed to display the Scigraph dialogs.

### DIFF
--- a/Core/clim-basic/encapsulate.lisp
+++ b/Core/clim-basic/encapsulate.lisp
@@ -464,6 +464,9 @@ if there is one, or STREAM"
 
 ;;; Extended Output Streams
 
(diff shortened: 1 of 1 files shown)


admich:
Here the changes you requested: 
1. removed the extra space
2. write a simpler version of `do-graphics-with-option`. I use your code but I don't use `call-next-method`. I try it and it works and maybe it is correct but I find it a little obscure (which is exactly the next method if I change the parameter?).
3. Use `encapsulating-stream-stream` function instead of `slot-value`.

PR #709, opened 2019-02-11
Scigraph: remove accept from dwim package

admich:
Following the suggetion of @slyrus I split PR #706 in three part. This is the third part about Scigraph accept function.

The ACCEPT version in DWIM package produced a SHEET-IS-NOT-A-CHILD
error when a dialog with gadget was closed.

This is due to how QUERY-IDENTIFIER parameter was passed to the
CLIM:ACCEPT function.

Because the Scigraph version in McCLIM is a CLIM2 only application the
DWIM:ACCEPT compatibility layer is no more necessary.

For this reason this commit remove the function ACCEPT from DWIM
package and now Scigraph use directly the CLIM:ACCEPT function.

### DIFF
--- a/Apps/Scigraph/dwim/package.lisp
+++ b/Apps/Scigraph/dwim/package.lisp
@@ -41,7 +41,8 @@ advised of the possiblity of such damages.
                 #:frame-manager
                 #:find-frame-manager
                 #:suggest
-                #:expression)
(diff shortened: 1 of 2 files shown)


slyrus:
Looks good to me. Thanks!

Issue #714, opened 2019-02-15
generic-list-panel: fails to redraw it with fewer elements than it originally had


josrr:
When items in a "generic-list-pane" are changed to a smaller list of items, the size of the pane is not recalculated according to the new number of items, causing glitches when redrawing it.

![generic-list-pane01](https://user-images.githubusercontent.com/3723736/52865727-df720d00-3102-11e9-86f7-f8e57a8f9f17.png)

Now, we change the items:
![generic-list-pane02](https://user-images.githubusercontent.com/3723736/52865731-e3059400-3102-11e9-8de3-b8dbbd9b2abd.png)

Seems good:
![generic-list-pane03](https://user-images.githubusercontent.com/3723736/52865733-e4cf5780-3102-11e9-84c8-f81ae0942e7a.png)

But, now we move the scroll-bar:
![generic-list-pane04](https://user-images.githubusercontent.com/3723736/52865735-e6008480-3102-11e9-8494-086be948b0db.png)

Issue #710, opened 2019-02-12
Aborted command invokes redisplay


dkochmanski:
Redisplay should only be invoked after successful command completion. Steps to reproduce: compile this file (it should open the window right away) and type `C-c` a few times in the interactor:

```common-lisp
(in-package #:clim-user)

(defun display (frame pane)
  (let ((x (random 100))
        (y (random 100)))
   (draw-rectangle* pane x y (+ x 100) (+ y 100))))

(define-application-frame clim-asdf ()
  ()
  (:panes (int :interactor :end-of-line-action :allow)
          (app :application :display-function 'display))
  (:layouts (default (horizontally () int app))))

(run-frame-top-level (make-application-frame 'clim-asdf))
```

Expected: no redisplay is invoked (rectangle on the right doesn't change its position)
Observed: redisplay is invoked

![image](https://user-images.githubusercontent.com/6871201/52636045-cb909600-2ecb-11e9-90dd-e82824edbef3.png)
![image](https://user-images.githubusercontent.com/6871201/52636059-d519fe00-2ecb-11e9-8cf5-9de2d98fadaf.png)

dkochmanski:
nb: "just" hitting enter does the same (no command is being executed).

josrr:
The [specification](http://bauhh.dyndns.org:8000/clim-spec/29-4.html#_1669) says that redisplaying must occur after the execution of a command if the option `:display-time` is `:command-loop`. But this option can also have the values `nil` and `t`; other behaviors should be followed in the latter two cases.

dkochmanski:
yes, that's what I mean. default display time for application pane is command-loop.

josrr:
I have a fix to this issue, I wanted to do a pull request, but I messed up:  the new branch (redisplay-fix) that I created  also includes the changes to the drawing tests, so here is the code:

[frames.lisp](https://github.com/josrr/McCLIM/blob/redisplay-fix/Core/clim-core/frames.lisp#L473)

PR #715, opened 2019-02-17
fix to issue #710 and change in the space requirements of the list-pane

josrr:
This implements the behavior of the option :display-time described in the [specification](http://bauhh.dyndns.org:8000/clim-spec/29-4.html#_1669).

Also solves #714.

### DIFF
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -496,9 +496,13 @@ documentation produced by presentations.")
               (*partial-command-parser* partial-command-parser)
               (interactorp (typep *query-io* 'interactor-pane)))
          (restart-case
-             (progn
-               (redisplay-frame-panes frame :force-p first-time)
-               (setq first-time nil)
+             (labels ((exec-cmd (frame command)
+                        (when command
+                          (execute-frame-command frame command)
+                          (redisplay-frame-panes frame))))
+               (when first-time
+                 (redisplay-frame-panes frame)
+                 (setq first-time nil))
                (if query-io
                    ;; For frames with an interactor:
                    (progn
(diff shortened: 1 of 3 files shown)


Core/clim-core/frames.lisp
```diff
@@ -496,9 +496,13 @@ documentation produced by presentations.")
               (*partial-command-parser* partial-command-parser)
               (interactorp (typep *query-io* 'interactor-pane)))
          (restart-case
-             (progn
-               (redisplay-frame-panes frame :force-p first-time)
-               (setq first-time nil)
+             (labels ((exec-cmd (frame command)
+                        (when command
+                          (execute-frame-command frame command)
+                          (redisplay-frame-panes frame))))
```
dkochmanski: this is not right – command may change dynamic variable values and that would leak into redisplay.

dkochmanski: if that's fine I'll fix it on my own (a solution will be similar in spirit to what you did)

Core/clim-core/frames.lisp
```diff
@@ -496,9 +496,13 @@ documentation produced by presentations.")
               (*partial-command-parser* partial-command-parser)
               (interactorp (typep *query-io* 'interactor-pane)))
          (restart-case
-             (progn
-               (redisplay-frame-panes frame :force-p first-time)
-               (setq first-time nil)
+             (labels ((exec-cmd (frame command)
```
lokedhs: Why does this function accept a `FRAME` argument? It's already available in the outer scope.

dkochmanski:
also fix for 714 is incorrect (we shouldn't clip max-height but rather recompute extent of the viewport). I'm closing this (see #717 for 2/3 changes without the list-pane one).

PR #718, opened 2019-02-20
Refactor word wrap

dkochmanski:
Word wrap is factored into a clim-agnostic utility. This is a first of a stream-output improvement patch series.

### DIFF
--- a/Core/clim-basic/utils.lisp
+++ b/Core/clim-basic/utils.lisp
@@ -550,8 +550,8 @@ STREAM in the direction DIRECTION."
 
 (defun bisect (start end predicate-fn &optional predicament-fn)
   "Finds the rightmost index meeting the PREDICATE-FN between START and END. It
-is assumed that START always meets the predicate while END may but doesn't have
-to meet it. That means that function always return some index.
+is assumed that START meets the predicate while END doesn't. These indexes are
+*not* tested.
 
 PREDICATE-FN INDEX
 Should return NIL if index does not meet the predicate and something else
@@ -561,8 +561,6 @@ PREDICAMENT-FN INDEX-1 INDEX-2
 Returns next index between its arguments for test.  If there is nothing more to
 test must return NIL. When not supplied default function looks always for an
 index being halfway between INDEX-1 and INDEX-2."
-  (when (funcall predicate-fn end)
-    (return-from bisect end))
   (unless predicament-fn
     (setf predicament-fn (lambda (last-good last-bad)
                            (let ((predicament (floor (+ last-good last-bad) 2)))
@@ -578,19 +576,6 @@ index being halfway between INDEX-1 and INDEX-2."
             (setf last-good current-guess)
             (setf last-bad current-guess))
      finally (return last-good)))
-
-;; Implementing line breaking as defined in Unicode[1] is left as an excercise
-;; for the reader. [1] https://unicode.org/reports/tr14/ -- jd 2019-01-08
-(defun line-break-opportunities (string start end &optional (break-characters '(#\space)))
-  "Returns a sequence of string indexes where line can break."
-  (loop
-     with space-indexes = (make-array 1 :fill-pointer 0 :adjustable t :element-type 'fixnum)
-     for i from start below (1- end)
-     do (when (member (aref string i) break-characters :test #'char=)
-          (vector-push-extend i space-indexes))
-     finally
-       (vector-push-extend (1- end) space-indexes)
-       (return space-indexes)))
 
 ;;; Command name utilities that are useful elsewhere.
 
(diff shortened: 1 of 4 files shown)


Core/clim-basic/utils.lisp
```diff
@@ -743,3 +728,195 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+                                      (new-start (alexandria:last-elt char-breaks)))
+                                 (multiple-value-bind (breaks remaining-cost)
+                                     (breaks-rec 0 margin new-start end i)
+                                   (values (append char-breaks breaks) remaining-cost))))))
+                        (T #1=(return-from breaks-rec
+                                (values best-breaks best-cost))))
+                finally #1#)))
+    (breaks-rec initial-offset margin start end 0)))
+
+;; Implementing line breaking as defined in Unicode[1] is left as an excercise
```
scymtym: Either remove this comment or explain what the downsides of the algorithm below compared the Unicode algorithm are.

Core/clim-basic/utils.lisp
```diff
@@ -743,3 +728,195 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+characters (width of a character m is taken as a reference value).
+
+BREAK-STRATEGY may be:
+- symbol T implementing a default line breaking by word strategy,
+- symbol NIL implementing a line breaking by character strategy,
+- function accepting index which should return T for break opportunity,
+- list of characters which are break opportunities (i.e space),
+- vector of string indexes which are break opportunities.
+
+START/END denote string beginning and ending offset."
```
scymtym: START/END designate the sub-sequence of STRING for which line breaks should be computed

Core/clim-basic/utils.lisp
```diff
@@ -743,3 +728,195 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+                                      (number width)))))
+  (unless initial-offset (setq initial-offset 0))
+  (assert (< start end))
+  (when (null break-strategy)
+    (return-from line-breaks
+      (%line-breaks-1 string width initial-offset margin start end)))
+  (let ((opportunities (etypecase break-strategy
+                         (function
+                          (coerce (loop for i from start below end
+                                     when (funcall break-strategy i)
```
scymtym: Wouldn't it make sense to pass the string as well?

Core/clim-basic/utils.lisp
```diff
@@ -743,3 +728,195 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+                        ((null best-break)
+                         (return-from breaks-rec
+                           (if (> offset 0)
+                               (list* start (breaks-rec 0 margin start end i))
+                               (let* ((char-breaks (%line-breaks-1 string width
+                                                                   offset margin
+                                                                   start opportunity))
+                                      (new-start (alexandria:last-elt char-breaks)))
+                                 (append char-breaks (breaks-rec 0 margin new-start end i))))))
+                        (T #1=(return-from breaks-rec
```
scymtym: why is `T` upper case?

Core/clim-basic/utils.lisp
```diff
@@ -743,3 +728,195 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+;;; beginning they are kept in the previous line.
+(defun %line-breaks-2 (string width initial-offset margin start end opportunities
+                       &aux (width-fn (etypecase width
+                                        (function width)
+                                        (number (lambda (string start end)
+                                                  (declare (ignore string))
+                                                  (* width (- end start)))))))
+  (labels ((skip-whitespace (string index)
+             (and index (position #\space string :start index :test-not #'char=)))
+           (breaks-rec (offset margin start end next-opportunity-index
```
scymtym: As far as I can tell, all recursive calls keep the initial `margin`, so this parameter is not needed.

Core/clim-basic/utils.lisp
```diff
@@ -743,3 +728,195 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+                                   (values (append char-breaks breaks) remaining-cost))))))
+                        (T #1=(return-from breaks-rec
+                                (values best-breaks best-cost))))
+                finally #1#)))
+    (breaks-rec initial-offset margin start end 0)))
+
+;; Implementing line breaking as defined in Unicode[1] is left as an excercise
+;; for the reader. [1] https://unicode.org/reports/tr14/ -- jd 2019-01-08
+(defun line-breaks (string width &key (break-strategy t) initial-offset margin (start 0) end)
+  "Function takes a string and returns a list of indexes where it should be split.
```
scymtym: Can the string contain #\Newline characters? If so, how are they handled?

Core/clim-basic/utils.lisp
```diff
@@ -743,3 +728,195 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
      ((and *print-pretty* *print-readably*)
       (simple-pprint-object ,sink ,self))
      (t ,@body)))
+
+
+
+(defun %line-breaks-1 (string width initial-offset margin start end)
```
scymtym: Add a comment explaining what this functions does.

Core/clim-basic/utils.lisp
dkochmanski: I had this remark made twice today! what a coincidence. I have a habit of putting capital T (and for consistency NIL) where they have a special meaning because I find small-case t as a single letter distinctive and easy to miss. Do you want me to change that?

Core/clim-basic/utils.lisp
dkochmanski: It can and how they are handled is not a concern of `line-breaks` (it all depends on `break-strategy` - i.e newline may be counted as a "break" character). The utility doesn't put any meaning to the characters (it is a responsibility of the break-strategy). Unicode algorithm should take such things into account (hard and soft breaks).

From the wider perspective (McCLIM-wise): stream-write-string splits the string by newline first so it is always the case, that there are no newlines in a string passed to `line-breaks`

```
(defmethod stream-write-string ((stream standard-extended-output-stream) string
                                &optional (start 0) end)
  (let ((seg-start start)
        (end (or end (length string))))
    (with-cursor-off stream
      (loop for i from start below end do
           (when (char= #\Newline
                        (char string i))
             (seos-write-string stream string seg-start i)
             (seos-write-newline stream)
             (setq seg-start (1+ i))))
      (seos-write-string stream string seg-start end))))
```

All breaks in seos-write-string are in fact soft newline breaks (candidates for rewrapping). I plan to use that quality when I get to the rewrapping functionality.

Core/clim-basic/utils.lisp
dkochmanski: My intention was to leave it at the caller discretion to make a closure over the string if that is necessary. From `line-breaks` perspective this interpretation makes more sense (imho).

Core/clim-basic/utils.lisp
dkochmanski: addressed in the newest commit

Core/clim-basic/utils.lisp
dkochmanski: addressed in the newest commit

Core/clim-basic/utils.lisp
dkochmanski: addressed in the newest commit

Core/clim-basic/utils.lisp
dkochmanski: addressed in the newest commit

Core/clim-basic/utils.lisp
scymtym: I would say stick to what rest of the code base does (I try to do that by not using keywords in `loop`).

Core/clim-basic/utils.lisp
scymtym: I think the first paragraph of what you explained here is a good candidate for inclusion in the documentation string.

Core/clim-basic/utils.lisp
dkochmanski: fixed (force-pushed)

Core/clim-basic/utils.lisp
dkochmanski: "default line breaking by word strategy" is a purposefully vague wording. Break opportunities are one thing, but by default we may want to break with minimum raggedness too (for instance), or we want to switch to some other approach (which is not related to unicode). Also adding into a docstring something what we wish we were doing misses the purpose of the documentation (imo).

scymtym: I meant this first paragraph:

> It can [contain newlines] and how they are handled is not a concern of line-breaks [...]

Mentioning this in the documentation string would make sense in my opinion. I was wondering about that after reading the documentation string.

dkochmanski: ah, makes sense. force-pushed (the last docstring sentence mentions it now).

PR #723, opened 2019-03-03
New margins

dkochmanski:


Documentation/Manual/Texinfo/extensions.txi
```diff
@@ -27,6 +28,75 @@ the manual. Drei documentation is provided as a separate document.
...
+@end deffn
+
+@anchor{Macro clime:with-temporary-margins}
+@deffn {Macro} {clime:with-temporary-margins} (stream &key (move-cursor t) left right top bottom) &body body
+@findex clime:with-temporary-margins
+Executes a body in a dynamic environment where stream's margins are
+augumented with provided one. Not all margins have to be specified. If
+they are not current margin values are taken as defaults.
+
+Each margin must be in on of following formats: @code{(:relative
```
lokedhs: Should this be `one`?

Documentation/Manual/Texinfo/extensions.txi
```diff
@@ -27,6 +28,75 @@ the manual. Drei documentation is provided as a separate document.
...
+space)} or @code{(:absolute space)}. Space may be specified as for
+@code{:x-spacing} and @code{:y-spacing} for horizontal and vertical
+margins accordingly. If margin is ``absolute'' then it corresponds to
+its exact placement in stream coordinates. ``relative'' margins are
+relative to the stream viewport region.
+
+If the boolean @code{move-cursor} is T then cursor is left where it
```
lokedhs: Perhaps change to `the cursor`?

Documentation/Manual/Texinfo/extensions.txi
```diff
@@ -27,6 +28,75 @@ the manual. Drei documentation is provided as a separate document.
...
+on a fresh page, and the final position is where the cursor is placed
+right before the page ends. Coordinates depend on current margins and
+text alignment.
+@end deffn
+
+@anchor{Function clime:stream-page-region}
+@deffn {Function} {clime:stream-page-region} stream
+@findex clime:stream-page-region
+Functions returns a region which corresponds to the stream page
+format. This region implies the stream margins.
```
lokedhs: What does the region imply? Is this a typo, and the intended word is `includes`?

dkochmanski: wording is correct, margins are implied by the region (to be exact: its bounding rectangle).

Documentation/Manual/Texinfo/extensions.txi
dkochmanski: fixed

Core/clim-basic/text-formatting.lisp
```diff
@@ -0,0 +1,130 @@
...
+                               (thunk top    y1 + :vertical)
+                               (thunk right  x2 - :horizontal)
+                               (thunk bottom y2 - :vertical)))))))
+
+(defmethod (setf sheet-region) :before (sheet-region (stream standard-page-layout))
+  (unless (region-equal sheet-region (sheet-region stream))
+    (slot-makunbound stream '%page-region)))
+
+(defun valid-margin-spec-p (margins)
+  (ignore-errors ; destructuring-bind may error; that yields invalid spec
```
lokedhs: I'm not a huge fan of `IGNORE-ERRORS`. It hides legitimate errors that should crash the application rather than return nil. Couldn't this actively check for the error cases and explicitly return nil in those cases?

Core/clim-basic/text-formatting.lisp
```diff
@@ -128,3 +128,88 @@
...
+

+;;; Mixin is used to store text-style and ink when filling-output it is invoked.
+
+(defclass %filling-output-graphics-state (gs-ink-mixin gs-text-style-mixin)
+  ()
+  (:default-initargs :ink +foreground-ink+ :text-style *default-text-style*))
+
+(defclass filling-output-mixin (gs-ink-mixin gs-text-style-mixin)
+  ((lbs :accessor line-break-strategy :initarg :line-break-strategy
```
lokedhs: When looking at the list of slots in the debugger, seeing a list of names of the form `LBS`, `ALB`, etc, can be confusing. How about longer, more informative names here?

Core/clim-basic/text-formatting.lisp
```diff
@@ -128,3 +128,88 @@
...

+;;; formatting functions
+(defun format-textual-list (sequence printer
+                            &key stream separator conjunction
+                              suppress-separator-before-conjunction
+                              suppress-space-after-conjunction)
+  "Outputs the SEQUENCE of items as a \"textual list\" into
+STREAM. PRINTER is a function of an item and a stream. Between each
+two items the string SEPARATOR is placed. If the string CONJUCTION is
+supplied, it is placed before the last item.
```
lokedhs: Shouldn't the docstring document the suppress keywords as well?

Core/clim-basic/text-formatting.lisp
```diff
@@ -0,0 +1,130 @@
...
+;;;    Author: Daniel Kochmański <daniel@turtleware.eu>
+;;;   License: LGPL-2.1+ (See file 'Copyright' for details).
+;;; ---------------------------------------------------------------------------
+;;;
+;;;  (c) copyright 2002 by Alexey Dejneka <adejneka@comail.ru>
+;;;  (c) copyright 2019 by Daniel Kochmański <daniel@turtleware.eu>
+;;;
+;;; ---------------------------------------------------------------------------
+;;;
+;;; Page layout may has numerous properties which arrange things on a
```
lokedhs: Typo: `may have`

Core/clim-basic/text-formatting.lisp
```diff
@@ -0,0 +1,130 @@
...
+;;; stream with having a broader picture in mind. Text on a page may
+;;; have alignment and direction, margins, columns, paragraph settings
+;;; and much more. This file is a backon abstraction which may be used
```
lokedhs: Is the word `backon` a typo? I don't know this word.

dkochmanski: beacon° I'll fix it

Core/clim-basic/text-formatting.lisp
dkochmanski: This is simply copied from clim-core/text-formatting.lisp.

Core/clim-basic/text-formatting.lisp
dkochmanski: what is an error case for destructuring-bind? `error`? how that would be different from ignore-errors? also this calls nothing outside this scope, so if anything causes the error it is malformed margin specification structure, hence should return nil. I'm against using ignore-errors for blocks which call anything arbitrary too.

Core/clim-basic/text-formatting.lisp
dkochmanski: slots have documentation, so this should impose a problem. slot-names are purely internal (even more- not used, we access them only via well-defined accessors).

Core/clim-basic/text-formatting.lisp
lokedhs: Good point. I guess the best I can come up with is using something like lambda-fiddle (available on QL). However, I'm not sure that is a complete solution anyway.

Core/clim-basic/text-formatting.lisp
lokedhs: I mentioned the debugger (the inspector, actually) which is where I usually see the slot names. If they are descriptive it helps when looking at object instances.

It's not a major issue, but my opinion was merely that since the name isn't actually used anywhere, it doesn't hurt to have them descriptive.

dkochmanski: imo it is too verbose and will break the indentaiton (which is nicely done with first keywords starting at the same column). I've used short names for aesthetical reasons.

lokedhs:
Looks good to me.

PR #724, opened 2019-03-03
In Extensions/render, merge color.lisp into utilities.lisp so that ty…

fittestbits:
…pe "octet" is defined before it's used.

Type "octet" was defined in color.lisp, but was used by a couple of functions in utilities.lisp. If the functions depending on "octet" were moved to color, then only one macro and one function would be left in utilities.lisp. And, since the functions in color.lisp only depend on simple lisp math functions or other functions defined in color.lisp, the cleanest solution seems to be to combine the two files.


dkochmanski:
lgtm, thanks!

PR #725, opened 2019-03-04
add dejavu font path in gentoo/funtoo system.

jingtaozf:
In Gentoo/Funtoo linux system, the font path for dejavu is `/usr/share/fonts/dejavu/`.

It will raise a warning if such path doesn't exist.


dkochmanski:
thanks!

PR #728, opened 2019-03-05
`defun highlight-completion-choice` needs to be inside an `eval-when`

pouar:
Sorry it took so long, I was asleep at the time you responded


scymtym:
Applied (with small modifications) to master. Thank you.

PR #729, opened 2019-03-06
New implementation of copy and paste

lokedhs:
This is a new implementation of copy and paste. It provides the following benefits compared to the old implementation:

- Support for copying other object types than just plain text. Text, HTML and images are currently supported.
- Better API that is not tied to CLX
- Support for clipboard in addition to selections
- Less invasive in terms of what events are caught by the implementation
- Integration with Drei
- Integration with CLIM presentations
- New "copy selection to clipboard" command
- Removed several ugly hacks from the previous implementation. Unfortunately not all can be removed because of underlying architecture limitations
- More consistent behaviour when pasting into interactor panes. Previously, in some cases paste didn't work depending on what state the editor was in.


Backends/CLX/port.lisp
```diff
@@ -213,6 +212,19 @@
...
+;; comment was already there. -- lokedhs 2019-02-16
+;;
+;; Incredibly crappy broken unportable Latin 1 encoder which should be
+;; replaced by various implementation-specific versions.
+(flet ((latin1-code-p (x)
+	 (not (or (< x 9) (< 10 x 32) (< #x7f x #xa0) (> x 255)))))
+  (defun string-encode (string)
+    (delete-if-not #'latin1-code-p (map 'vector #'char-code string)))
+  (defun exactly-encodable-as-string-p (string)
+    (every #'latin1-code-p (map 'vector #'char-code string))))
```
dkochmanski: - is string-encode used anywhere? if not then please remove it.
- exactly-encodable-as-string-p is a local function to one of %realize-mirror methods, please put it in flet
- can we implement exactly-encodable-as-string-p as the following? if so, please do that. where does latin1 requirement come from?

```
(handler-case (coerce string 'base-string)
  (:error () nil)
```

alternatively (if it must be latin-1):

```
(handler-case (babel:string-to-octets string :encoding :latin-1)
  (:no-error () t)
  (error () nil))
```

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,76 @@
+(in-package :climi)
+
+;;;
+;;;  Functions dealing with copying to the clipboard.
+;;;
+
+(deftype representation-type-name () '(member :string :html :image))
+
+(defgeneric copy-to-clipboard-with-port (port sheet clipboard-p object presentation-type)
+  (:documentation "Method to be implemented by backends."))
```
dkochmanski: documentation should say, what this function is supposed to do, not who should implement it. you may put that in a comment (and update documentation).

Most notably I do not understand what is clipboard-p, why it needs both port an sheet etc.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,76 @@
...
+content of the clipboard will be available after the sheet has been
+removed."
+  (copy-to-clipboard-with-port (port sheet) sheet t object presentation-type))
+
+(defun copy-to-selection (sheet object &key presentation-type)
+  "Copy OBJECT to the selection.
+SHEET is the owner of the selection, and it is not guaranteed that the
+content of the selection will be available after the sheet has been
+removed."
+  (copy-to-clipboard-with-port (port sheet) sheet nil object presentation-type))
```
dkochmanski: what does it mean to copy to selection? how is it different from copying to the clipboard? is it just that "copy-to-clipboard" is copy and "copy-to-selection" is paste?

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,76 @@
...
+
+(defmethod clear-clipboard-with-port (port sheet clipboard-p)
+  "Fallback implementation if the implementation does not support a clipboard.."
+  nil)
+
+(defun clear-clipboard (sheet)
+  (clear-clipboard-with-port (port sheet) sheet t))
+
+(defun clear-selection (sheet)
+  (clear-clipboard-with-port (port sheet) sheet nil))
```
dkochmanski: same story, I don't understand this interface duplication.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,76 @@
...
+;;;
+
+(deftype representation-type-name () '(member :string :html :image))
+
+(defgeneric copy-to-clipboard-with-port (port sheet clipboard-p object presentation-type)
+  (:documentation "Method to be implemented by backends."))
+
+(defmethod copy-to-clipboard-with-port ((port clim:port) (sheet clim:sheet) clipboard-p object presentation-type)
+  "Fallback implementation if the implementation does not support a clipboard."
+  nil)
```
dkochmanski: put default method as `(:method …)` option in defgeneric. This docstring is a no-op, so please put that in comment.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,76 @@
...
+content of the selection will be available after the sheet has been
+removed."
+  (copy-to-clipboard-with-port (port sheet) sheet nil object presentation-type))
+
+(defgeneric clear-clipboard-with-port (port sheet clipboard-p))
+
+(defmethod clear-clipboard-with-port (port sheet clipboard-p)
+  "Fallback implementation if the implementation does not support a clipboard.."
+  nil)
+
```
dkochmanski: ditto, `(:method …)` and docstring -> comment.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,76 @@
...
+  (copy-to-clipboard-with-port (port sheet) sheet t object presentation-type))
+
+(defun copy-to-selection (sheet object &key presentation-type)
+  "Copy OBJECT to the selection.
+SHEET is the owner of the selection, and it is not guaranteed that the
+content of the selection will be available after the sheet has been
+removed."
+  (copy-to-clipboard-with-port (port sheet) sheet nil object presentation-type))
+
+(defgeneric clear-clipboard-with-port (port sheet clipboard-p))
```
dkochmanski: please add documentation option to this function.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,76 @@
+(in-package :climi)
+
```
dkochmanski: please explain in a toplevel comment what is clipboard, what is selection, how they interact with each other, what are differences and similarties. in other words provide these newly introduced concepts summary.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,76 @@
...
+;;;  The following functions implement the standard API to request
+;;;  content from the clipboard.
+;;;
+
+(defclass clipboard-send-event (window-event)
+  ((content :initarg :content
+            :reader clipboard-event-content)
+   (type    :initarg :type
+            :reader clipboard-event-type))
+  (:documentation "Event containing the result of a clipboard or selection request"))
```
dkochmanski: if it is part of core it may be worth considering putting it in `events.lisp` instead.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,76 @@
...
+              (and (listp type)
+                   (every (lambda (v) (typep v 'climi::representation-type-name)) type)))
+    (error "Invalid type: ~s" type))
+  (call-next-method))
+
+(defun request-selection-content (pane type)
+  (request-clipboard-content-with-port (port pane) pane nil type))
+
+(defun request-clipboard-content (pane type)
+  (request-clipboard-content-with-port (port pane) pane t type))
```
dkochmanski: ditto, I don't know what these duplicated interfaces are for

Backends/CLX/port.lisp
lokedhs: This comes from the old implementation. It was using a function that used to reside in `text-selection.lisp` (out of convenience, I think).

I'm wondering why they even call `EXACTLY-ENCODABLE-AS-STRING-P` in the first place. Why not always set the `WM_NAME` property? It definitely shouldn't hurt, right? That would allow us to eliminate this function, as well as the two `WHEN` forms.

Core/clim-basic/clipboard.lisp
lokedhs: `CLIPBOARD-P` is a boolean specifying whether the function should operate on the clipboard or the selection. Perhaps having a keyword argument `:CLIPBOARD` or `:SELECTION` would be better? Another alternative is to split it into two separate functions.

The reason both `PORT` and `SHEET` is specified is because the generic function needs to dispatch on the port. Note that this function should never be called by users.

The user-facing API consists of two functions: `COPY-TO-CLIPBOARD` and `COPY-TO-SELECTION`. Both of these functions then simply call `COPY-TO-CLIPBOARD-WITH-PORT`, with the first argument simply being the sheet's port, and `CLIPBOARD-P` being used to discriminate which of the previously mentioned two functions were called.

I agree that all of this information should go into the docstring. I'If you're OK with the API, then I'll write the documentation. Otherwise I'll change the API first.

Core/clim-basic/clipboard.lisp
lokedhs: No. As explained in the overview document (which has not been merged into this change yet), there are two separate concepts: The *selection* is the the highlighted text. That's the text that is pasted when you middle-click in most X11 applications.

The *clipboard* is a separate thing. It's where the information that is pasted using Control-V in most applications copy from. When typing Control-C, the content of the selection is copied to the clipboard.

Thus, when text is highlighted, `COPY-TO-SELECTION` is called to inform the system that something has been selected. In a later commit, a separate command will be added that implements the behaviour of Control-C in most applications, and it checks if some text is selected, extracts it, and calls `COPY-TO-CLIPBOARD` on it.

Core/clim-basic/clipboard.lisp
lokedhs: This is similar to `COPY-TO-SELECTION` and `COPY-TO-CLIPBOARD` where the two user-facing functions are directed to a single backend implementation. Just like in that case, the term *clipboard* in the backend function might not be ideal, since it actually works with both the selection and the clipboard.

Backends/CLX/port.lisp
lokedhs: I decided to always set `_NET_WN_NAME`. All applications I've checked (such as Emacs or Gnome-Terminal) seems to do it, even when the name is pure ASCII.

Core/clim-basic/clipboard.lisp
lokedhs: I've added some documentation to this function.

Core/clim-basic/clipboard.lisp
lokedhs: Done, and documentation added.

Core/clim-basic/clipboard.lisp
lokedhs: Done.

Core/clim-basic/clipboard.lisp
lokedhs: Done.

Core/clim-basic/clipboard.lisp
lokedhs: Documented.

Core/clim-basic/clipboard.lisp
lokedhs: Moved. Is the location in that file a good place to put it? (after the other window-related events, but before timer-event)

Core/clim-basic/clipboard.lisp
lokedhs: I have added a comment section at the beginning of the file explaining some of the terminology.

Core/clim-basic/clipboard.lisp
dkochmanski: please split this new protocol between selection and clipboard modules. fake unification (with clipboard-p) buys us nothing but still imposes extra cognitive effort on backend implementers (even if selection doesn't make sense there).

also remove extra functions (i.e instead of clear-clipboard and clear-clipboard-with-port make one generic function clear-sheet-clipboard accepting both port and a sheet (that's how i.e set-sheet-cursor is implemented, so it would be consistent).

So on example of clear-sheet-*:

Remove clear-clipboard-with-port and change clear-clipboard / clear-selection to generic functions with arguments (port sheet). moreover move clear-selection to selection module.

Core/clim-basic/clipboard.lisp
dkochmanski: yes, events is a good file

Core/clim-basic/clipboard.lisp
dkochmanski: please see my other comment.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,159 @@
+(in-package :climi)
```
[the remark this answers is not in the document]
lokedhs: Changed

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,159 @@
...
+;;;
+;;;    Selection: The currently selected data (for example using
+;;;      shift-drag in a stream pane). X11 allows an application to
+;;;      request the content of the selection, usually by pressing
+;;;      middle-click to paste. Other windowing systems might not
+;;;      allow one application to request the content of the selection
+;;;      from another application.
+;;;
+;;;    Clipboard:
+;;;
```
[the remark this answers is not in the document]
lokedhs: Changed

Core/clim-basic/clipboard.lisp
lokedhs: Changed according to recommendation.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+
+Please refer to the documentation for REQUEST-CLIPBOARD-CONTENT for
+details.")
+  (:method ((port clim:port) (pane clim:sheet) type)
+    nil))
+
+(defmethod request-selection-content :around (port name type)
+  (unless (or (typep type 'climi::representation-type-name)
+              (and (listp type)
+                   (every (lambda (v) (typep v 'climi::representation-type-name)) type)))
```
dkochmanski: maybe: `(unless (every (lambda (v) (typep v representation-type-name)) (ensure-list type)) …`

but even better: define a type (`satisfies`) against this predicate and call check-type here for clarity. or even better better define `representation-type` type instead of `representation-type-name` which takes that into account (i.e by satisfies) and check-type on that.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+
+After calling this function, if the clipboard is active and can be
+presented as the requested type, an event of type CLIPBOARD-SEND-EVENT
+will be delivered to PANE. This event will contain the requested data
+in one of the requested formats.
+
+TYPE is a list of keywords (or a single keyword, in which case it is
+interpreted as a list of a single element) that indicates the type of
+data that can be accepted. If the clipboard content can be represented
+in more than one format, the first matching type will be used.")
```
dkochmanski: any keyword? can I put there ":i-like-pancakes"? if not, what are the values which may be put here?

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+(defgeneric request-clipboard-content (port pane type)
+  (:documentation "Backend implementation of REQUEST-CLIPBOARD-CONTENT.")
+  (:method ((port clim:port) (pane clim:sheet) type)
+    nil))
+
+(defgeneric request-selection-content (port pane type)
+  (:documentation "Request the content of the selection.
+
+Please refer to the documentation for REQUEST-CLIPBOARD-CONTENT for
+details.")
```
dkochmanski: referring to request-clipboard-content gives me: "backend implementation of request-clipboard-content". please put docstrings separate for each after putting in separate files. Also, this should describe what it does, not where it is implemented. Backend implementer will surely appreciate that.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+content of the selection, or NIL,NIL if it is not held.")
+  (:method ((port clim:port))
+    nil))
+
+;;;
+;;;  The following functions implement the standard API to request
+;;;  content from the clipboard.
+;;;
+
+(defgeneric request-clipboard-content (port pane type)
```
dkochmanski: generic function request-clipboard-content is defined twice (see below). the second docstring is of course more acceptable.

Core/clim-core/presentations-clipboard.lisp
```diff
@@ -0,0 +1,24 @@
...
+(defun convert-clipboard-content (obj output-type &key type check-only)
+  (funcall-presentation-generic-function convert-clipboard-content
+                                         obj
+                                         (or type (presentation-type-of obj))
+                                         output-type
+                                         check-only))
+
+(defun supported-clipboard-types (obj type)
+  (loop
+    for output-type in '(:string :html)
```
dkochmanski: this shows string html, but representation-type-name has also image. how does it work in practice?

Backends/CLX/clipboard.lisp
```diff
@@ -0,0 +1,296 @@
...
+      (set-stored-object port selection
+                         (if success-p
+                             (make-instance 'clx-clipboard-stored-object
+                                            :content obj
+                                            :type presentation-type
+                                            :owner sheet)
+                             nil))
+      success-p)))
+
+(defmethod clim-extensions:copy-to-selection ((port clx-clipboard-port-mixin) sheet obj &key presentation-type)
```
dkochmanski: please use `clime` abbreviation

Backends/CLX/clipboard.lisp
```diff
@@ -0,0 +1,296 @@
...
+
+(defun set-stored-object (port selection value)
+  (ecase selection
+    (:primary (setf (selection-content port) value))
+    (:clipboard (setf (clipboard-content port) value))))
+
+;;; Event classes
+
+(defclass clx-selection-event (window-event)
+  ())
```
dkochmanski: get rid of all this event hackery and put calls to process-selection-* directly in event-handler in input.lisp.

Core/clim-basic/stream-input.lisp
```diff
@@ -220,11 +220,16 @@ keys read."))
...
-                           (or (typep event 'key-press-event)
-                               (typep event 'pointer-button-press-event))))))
+                 (not (or (and (typep sheet 'clim-stream-pane)
+                               (not (typep sheet 'interactor-pane))
+                               (or (typep event 'key-press-event)
+                                   (typep event 'pointer-button-press-event)))
+                          (and (typep sheet 'interactor-pane)
+                               (or (typep event 'key-press-event)
+                                   (typep event 'pointer-button-press-event)
+                                   (typep event 'clipboard-send-event)))))))
```
dkochmanski: terrible! :-) but makes sense

Examples/clipboard-demo.lisp
```diff
@@ -0,0 +1,101 @@
+(defpackage :clipboard-demo
```
dkochmanski: this demo needs some work, here is a screenshot:
image

optimally inspector is smaller, wrapping is made by word and whole frame has bigger width (so there is no need to scroll). moreover paste should be performed in a separate pane than instructions for clarity. alternatively a horizontal line should be drawn between instructions and the are where paste is put.

Core/clim-core/builtin-commands.lisp
```diff
@@ -428,3 +428,36 @@
...
+    (t clipboard-object clim:global-command-table :tester ((obj presentation)
+                                                           (supported-clipboard-types obj (presentation-type presentation))))
+    (obj presentation)
+  (make-instance 'clipboard-object :content obj :type (presentation-type presentation)))
+
+(clim:define-command (com-copy-to-clipboard :command-table clim:global-command-table :name "Copy Presentation to Clipboard")
+    ((obj clipboard-object :prompt "Object"))
+  (let ((pane (clim:frame-top-level-sheet clim:*application-frame*)))
+    (copy-to-clipboard (clim:port pane) pane (clipboard-object/content obj)
+                       :presentation-type (clipboard-object/type obj))))
```
dkochmanski: I can see that I can "Copy Presentation to Clipboard" disregarding its type (i.e forms in the interactor), but nothing happens to be in the clipboard when there is no translator. Is there a way to make the command abort, fail, or refuse to accept presentations which doesn't have appropriate translators?

Steps to reproduce:
run stream-test
issue any command
run command `copy presentation to clipboard` and select previous form
try to paste (nothing happens)


Second issue: copy text command if there is no text selected should not complete successfully either.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+is actually performed. Use the :CHECK-ONLY argument to check whether a
+conversion is supported.
+
+PORT is the port that the sheet belongs to. It should be the primary
+argument that individual sheets specialise on.
+
+SHEET is the owner of the selection or clipboard.
+
+If CLIPBOARD-P is true, this call is a request to copy to the
+clipboard, if false it's a copy to the selection.
```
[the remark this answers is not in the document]
lokedhs: Fixed

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+;;;  between applications.
+;;;
+;;;    Selection:
+;;;
+;;;      The currently selected data (for example using shift-drag in
+;;;      a stream pane). X11 allows an application to request the
+;;;      content of the selection, usually by pressing middle-click to
+;;;      paste. Other windowing systems might not allow one
+;;;      application to request the content of the selection from
+;;;      another application.
```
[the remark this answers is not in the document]
lokedhs: My opinion is that the description in question discusses the fundamental concepts, for example the fact that there are two "boxes" that can hold information for transfer between processes.

The file `text-selection.lisp` deals with the shift-drag highlighting of text. It uses the above-mentioned protocol, but does not define it. They are two very different things.

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+
+(defgeneric local-selection-content (port)
+  (:documentation "Returns the content of the selection in this Lisp image.
+If the global selection is currently held, the value returned is the
+same as what would be sent in response to a REQUEST-CLIPBOARD-CONTENT
+call.
+
+If content is available, this function should return two values: The
+content of the selection, and the presentation-type of the object.
+Otherwise NIL,NIL should be returned.
```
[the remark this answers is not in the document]
lokedhs: Fixed

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+
+This function is used by the copy selection to clipboard operation.
+This is in order to present the same beahviour as most non-CLIM
+applications exhibit: They are are able perform the copy
+operation (usually invoked by control-c) even when not owning the
+system-wide selection.
+
+The simplest implementation of this function would simply return the
+content of the selection, or NIL,NIL if it is not held.")
```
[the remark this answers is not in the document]
lokedhs: Fixed

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+same as what would be sent in response to a REQUEST-CLIPBOARD-CONTENT
+call.
+
+If content is available, this function should return two values: The
+content of the selection, and the presentation-type of the object.
+Otherwise NIL,NIL should be returned.
+
+This function is used by the copy selection to clipboard operation.
+This is in order to present the same beahviour as most non-CLIM
+applications exhibit: They are are able perform the copy
```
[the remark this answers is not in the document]
lokedhs: Fixed

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+
+;;;
+;;;  The following functions implement the standard API to request
+;;;  content from the clipboard.
+;;;
+
+(defgeneric request-clipboard-content (port pane type)
+  (:documentation "Backend implementation of REQUEST-CLIPBOARD-CONTENT.")
+  (:method ((port clim:port) (pane clim:sheet) type)
+    nil))
```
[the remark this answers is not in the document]
lokedhs: Fixed

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,174 @@
...
+details.")
+  (:method ((port clim:port) (pane clim:sheet) type)
+    nil))
+
+(defmethod request-selection-content :around (port name type)
+  (unless (or (typep type 'climi::representation-type-name)
+              (and (listp type)
+                   (every (lambda (v) (typep v 'climi::representation-type-name)) type)))
+    (error "Invalid type: ~s" type))
+  (call-next-method))
```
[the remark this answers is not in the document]
lokedhs: Fixed

Core/clim-basic/clipboard.lisp
lokedhs: Fixed. I decided to not create a new type, as that would be confusing in cases where a list is not allowed (most uses of this type).

Core/clim-basic/clipboard.lisp
lokedhs: Yes. Any keyword can be used. Unsupported types are ignored.

Core/clim-core/presentations-clipboard.lisp
lokedhs: This is intentional. Support for sending images to external processes has not been implemented.

Core/clim-basic/clipboard.lisp
lokedhs: Fixed.

Core/clim-basic/clipboard.lisp
lokedhs: Fixed.

Backends/CLX/clipboard.lisp
lokedhs: Fixed

Backends/CLX/clipboard.lisp
lokedhs: The event hackery originates from the old implementation. I decided to keep it since I wanted to make sure the clipboard content accessed from the main application thread, and not the event thread (in order to avoid race conditions).

I now note that even the `HANDLE-EVENT` is processed in the main CLX event loop rather than the application thread. Clearly this was a problem even with the old implementation.

This means that moving everything to `input.lisp` is definitely possible, and will not cause any more problems than we're already facing. However, the threading problem exists, and I don't know what should be done to fix it.

I'm pretty sure there are other places where objects are accessed from both the event thread and the application thread without locking (I'd presume anything that implements `HANDLE-EVENT`).

Examples/clipboard-demo.lisp
lokedhs: I have made changes to the example code according to these recommendations. Please take a look and let me know if it's better.

Core/clim-basic/clipboard.lisp
dkochmanski: what's the point of not erroring on types which are *not* supported?

Core/clim-basic/clipboard.lisp
```diff
@@ -0,0 +1,166 @@
...
+data that can be accepted. If the clipboard content can be represented
+in more than one format, the first matching type will be used.")
+  (:method ((port clim:port) (sheet clim:sheet) type)
+    nil))
+
+(defmethod request-clipboard-content :around (port name type)
+  (unless (every (lambda (v)
+                   (typep v 'representation-type-name))
+                 (alexandria:ensure-list type))
+    (error "Invalid type: ~s" type))
```
dkochmanski: also this contradicts that "unsupported types are ignored". it is not consistent.

Backends/CLX/clipboard.lisp
dkochmanski: fixed… how?

Backends/CLX/clipboard.lisp
dkochmanski: handle-event is always issues from the application event loop. event queue is a thread-safe mailbox, so queue and handle are not prone to races of any kind. that doesn't change the fact that handling this communication with X is local to the port (and only send-notify is passed to the rest of the system). For setting clipboard from the application thread – this should be made secure thread-wise.

I still maintain that this handle-event hackery should be removed.

Examples/clipboard-demo.lisp
```diff
@@ -0,0 +1,116 @@
...
+    (clim:redisplay-frame-pane frame pane)))
+
+(defun display-clipboard-demo (frame stream)
+  (declare (ignore frame))
+  (format stream "Select text using shift and the left mouse button~%")
+  (format stream "You can also use the command \"Copy to clipboard\" to select a presentation.~%~%")
+  (let ((obj (make-instance 'markup-text :text "This text can be copied in HTML form")))
+    (clim:stream-present stream obj (clim:presentation-type-of obj)))
+  (format stream "~%~%You can paste the clipboard into editor fields using Shift-Insert,~%")
+  (format stream "or you can Shift-click the middle mouse button to paste from the selection.~%~%")
```
dkochmanski: shift-click-middle-button gesture doesn't work here.

Examples/clipboard-demo.lisp
dkochmanski: this looks much better. please also add command to clear "paste pane". moreover "menu" lacks "copy presentation to clipboard".

Core/clim-basic/clipboard.lisp
lokedhs: The argument is that if you have a backend which does not, for example, support copying of `:STRING` objects, then you don't want to have an exception thrown whenever you type C-k in a Drei pane (which copies the string to the clipboard).

However, when calling either `REQUEST-*` or `COPY-*`, then you want to throw an exception if the caller specified some keyword other than the main ones supported (currently, `:STRING`, `:HTML` and `:IMAGE`).

You only want the backend to deal with the types it supports. The main check to verify that the caller used a proper type is verified upstream (in the `:AROUND` method).

Core/clim-core/presentations-clipboard.lisp
lokedhs: Oops. It's not intentional. I thinking of the wrong file. Fixed.

Backends/CLX/clipboard.lisp
lokedhs: Good question. It is now.

Core/clim-core/builtin-commands.lisp
lokedhs: I've added the commands.

lokedhs: For the failure if nothing is selected, how should the failure manifest itself? It's definitely possible to print a message to `*STANDARD-OUTPUT*`, but that might not always be preferable. The information is there, it's just a matter of how to report it back to the user.

The other issue is a trickier one. When you selected the previous form, you're actually not selecting it. Rather, you are reissuing the previous command. The way to show this is the case is to do the following:

  - Type `Clear Paste Pane` on the commandline
  - Paste some text info the paste pane
  - Type `Copy Presentation to Clipboard` followed by space so that the commandline prompts for a presentation.
  - Click on the `Clear Paste Pane` command that was previously typed

After this, the paste pane is cleared. What happens is that it looks like the `Clear Paste Pane` command was chosen, but it was in fact executed instead.

I believe this is a bug that is unrelated to the clipboard feature. It shouldn't even be possible to select the previous command at this point, since its presentation does not have a converter (one can, and probably should, be written, but that's s separate issue).

Core/clim-basic/clipboard.lisp
lokedhs: I wrote most of the details in the comment to the previous note. There was a bug which has not been fixed, but the main point is that the end-user API should check for calid types, but the backends should ignore types they are unable to process.

Examples/clipboard-demo.lisp
lokedhs: Shift-middle-click is only implemented where you can type text (designed to behave similar to GNOME applications). It is of course possible to implement it in the clipboard-demo.

Core/clim-basic/clipboard.lisp
dkochmanski: I still don't get why valid keywords are not specified given error is signalled in before method.

Core/clim-core/builtin-commands.lisp
dkochmanski: you may trigger "command aborted". that will prevent pane being redisplayed(!)

as of the issue, please report it with steps to reproduce then.

Examples/clipboard-demo.lisp
dkochmanski: I know, I've tried to put it in the edit box.

Core/clim-basic/clipboard.lisp
lokedhs: As per previous discussion on IRC, I've removed the check.

Backends/CLX/clipboard.lisp
lokedhs: After actually removing the code that implemented the hackery, I agree with you. It's much cleaner this way, and doesn't introduce any new concurrency issues that didn't already exist.

I've pushed a fix.

Core/clim-core/builtin-commands.lisp
lokedhs: That sounds like a good idea. Only problem is that I don't know how to trigger a command aborted. I'll experiment.

I will report the issue separately with a simple test case. I should be able to create one that doesn't require copying and pasting stuff.

Examples/clipboard-demo.lisp
lokedhs: That's strange. It works for me. Can you provide an exact step-by-step explanation as to what you do? Perhaps there is some edge-case where the event isn't delivered in some circumstances. The more exact you can be, the better.

Core/clim-core/builtin-commands.lisp
lokedhs: I have now updated the copy command so that it triggers an abort when the selection is not active.

Fixed in 77cb2feaf6e370dcd7ed33d41193c22f02105dfa

Backends/CLX/clipboard.lisp
```diff
@@ -0,0 +1,227 @@
...
+(defun find-stored-object (port selection)
+  (ecase selection
+    (:primary (selection-content port))
+    (:clipboard (clipboard-content port))))
+
+(defun set-stored-object (port selection value)
+  (ecase selection
+    (:primary (setf (selection-content port) value))
+    (:clipboard (setf (clipboard-content port) value))))
+
```
dkochmanski: `find-stored-object` and `set-stored-object` should have a shared lock (may be a slot in port, for instance clipboard-lock). they are accessed from both input processing thread and the application command loop(s). Sadly neither bordeaux-threads nor clim-sys defines read-write locks, because this would be a classic example.

```
(process-next-event-thread
 (:event :selection-notify
         (event-handler)
         (process-selection-notify)
         (process-string-reply)         ; not always strings!
         (sheet-event-queue clipboard-send-event))

 (:event :selection-request
         (event-handler)
         (process-selection-request)
         (find-stored-object))

 (:event :selection-clear
         (event-handler)
         (process-selection-clear)
         (set-stored-object))

 (:event :mixed-event-drag-with-key-modifiers
         (dispatch-event :around)
         (eos/shift-click)
         (pane-clear-markings)
         (clime:clear-selection)
         (clx-clear-clipboard)
         (find-stored-object))

 (:event :mixed-event-release
         (dispatch-event :around)
         (eos/shift-release)
         (clime:copy-to-selection)
         (clx-copy-to-clipboard)
         (set-stored-object)))

(command-loop-thread
 (:command (com-copy-text-to-clipboard) ; selection ->
           (clime:local-seleciton-content)
           (find-stored-object))
 (:command (com-copy-text-to-clipboard) ; -> clipboard
           (clime:copy-to-clipboard)
           (clx-copy-to-clipboard)
           (set-stored-object))
 (:command (com-copy-to-clipboard)
           (clime:copy-to-clipboard)
           (clx-copy-to-clipboard)
           (set-stored-object))
 (:drei (probably-handle-event)
        (push-start-and-copy)
        (clime:copy-to-clipboard)
        (clx-copy-to-clipboard)
        (set-stored-object)))
```

Backends/CLX/clipboard.lisp
```diff
@@ -0,0 +1,227 @@
...
+                                      when (member type targets)
+                                        return type)
+                            when v
+                              return v)))
+      (if selected-type
+          (xlib:convert-selection selection selected-type window :mcclim time)
+          ;; ELSE: The clipboard doesn't support content of this type, send a negative response here
+          nil))))
+
+(defun process-string-reply (port selection content type)
```
dkochmanski: this function name is bogus, reply may be image according to above. `process-clipboard-reply`?

Backends/CLX/clipboard.lisp
```diff
@@ -0,0 +1,227 @@
...
+(defun process-selection-clear (port selection)
+  (set-stored-object port selection nil))
+
+;;;
+;;;  Paste support
+;;;
+
+(defmethod clx-request-clipboard-content ((port clx-clipboard-port-mixin) pane selection type)
+  (setf (clipboard-outstanding-request-pane port) pane)
+  (setf (clipboard-outstanding-request-type port) (if (listp type) type (list type)))
```
dkochmanski: alexandria:ensure-list

Backends/CLX/clipboard.lisp
dkochmanski: this are simple accessors, so I don't think it is possible that incomplete state is read.

Backends/CLX/clipboard.lisp
lokedhs: I agree. Changed.

Backends/CLX/clipboard.lisp
lokedhs: Changed.

Core/clim-basic/text-selection.lisp
```diff
@@ -52,46 +52,6 @@
...
-(defclass selection-event (window-event)
-  ((selection :initarg :selection
-              :reader selection-event-selection)))
-
-(defclass selection-clear-event (selection-event)  ())
-(defclass selection-notify-event (selection-event) ())
-(defclass selection-request-event (selection-event)
-  ((requestor :initarg :requestor :reader selection-event-requestor)))
-
-
```
dkochmanski: there are remaining implementations of these protocols in codebase (i.e Null backend). please grep for them and remove them. Also, if you remove protocols please remove them from documentation too.

dkochmanski:
superseded

PR #732, opened 2019-03-08
scigraph: y label with transformed glyphs

admich:


scymtym:
Merge into master with minor modifications. Thanks.

PR #738, opened 2019-03-18
stream-cursor-initial-position: fix a few regressions

### LINKED ISSUE #737: Enabling margins adds extra vertical space in interactor pane
If a command outputs graphics wrapped in `WITH-ROOM-FOR-GRAPHICS` and the interactor pane uses text margins, some extra space is added after the graphics. If the margins are removed, this doesn't happen.

This is the result of running the test program:
with-margins
If the margins are removed, then the output is as expected:
without-margins

dkochmanski:
Since margin introduction we can't assume that page starts at cursor
position (values 0 0). Drei seemed to make this assumption what kind
of broke interactor - spurious new lines from read-command. We fix the
cursor position in pdf and ps backends too since we're at it.

Moreover we adjust margins for with-output-to-output-record to match
the viewport (spotting this issue was a result of investigating #737.

Closes #737.


lokedhs:
This fixes all the observed issues. Also tested with Climaxima.

PR #740, opened 2019-03-20
Make PRESENTATION-HISTORY-ADD into a generic function

lokedhs:
This fix changes PRESENTATION-HISTORY-ADD from a regular function to a
generic function. This function is not exported, so there should be no
change in behaviour.

The purpose of making this function generic is to allow software to
override what is stored in the presentation history.

### DIFF
--- a/Core/clim-core/presentation-defs.lisp
+++ b/Core/clim-core/presentation-defs.lisp
@@ -665,13 +665,14 @@ history will be unchanged."
          (when ,added
            (decf (fill-pointer (presentation-history-array ,history))))))))
 
-(defun presentation-history-add (history object ptype)
-  "Add OBJECT and PTYPE to the HISTORY unless they are already at the head of
- HISTORY"
-  (multiple-value-bind (top-object top-ptype)
-      (presentation-history-top history ptype)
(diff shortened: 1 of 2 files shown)


nsrahmad:
This is not the right approach.

If I understand correctly, you do not want certain commands to appear in `interactor` history. Instead of modifying `history`, you can try making `history-yank-next` and `history-yank-previous` as generic functions and specializing on `stream`.

These functions are here:
https://github.com/McCLIM/McCLIM/blob/52641f72a9c1043710ce72ebc2d5f10ff5f320ca/Core/clim/input-editing-drei.lisp#L186-L232

lokedhs:
@nsrahmad You are right that those two functions should be generic, and I've added a fix to that effect. However, I still feel that the original function could be generic too, at it can be argued that it doesn't hurt if it is.

Would the current proposed change be acceptable?

dkochmanski:
the question is: is there a need for that? also, if presentation-history-add is to be a generic function all other operators on the history should be generic functions too. It doesn't make sense otherwise. If there is no need for that, please remove the first commit from the pull request.

lokedhs:
I need to do more research into a proper solutions, so closing this for now.

Issue #755, opened 2019-04-16
Some dialog gadget views don't change the query value.


admich:
The following dialog gadget views don't change the query value:

- radio-box-view
- check-box-view
- slider-view used with integer type

It is possible to see that with the demo  "Accepting Value Test > Accepting With Gadget", if you change the radio box, the check-box or the integer slider the returned value is not correct.

This is due to some changes made to my PR #682 when it was merged.

I can sumbmit a new PR with the fix in few time.

PR #758, opened 2019-04-16
Fix-755

admich:
Before this commit the following dialog gadget views don't change the query value:

- radio-box-view
- check-box-view
- slider-view used with integer type

This fix issue #755


scymtym:
Applied with minor changes. Thank you.

Issue #754, opened 2019-04-13
command-menu pane initially displays wrong


heshrobe:
In an application frame, with command menu whose display function is clim:display-command-menu: Initially the frame displays incorrectly with the command menu items text printing vertically and overwriting the pane above the menu.

You can get the same behavior by calling 
(redisplay-frame-pane <application-frame> <menu-pane> :force-p t).  

You can then get the display correct by issuing any command in the application-frame or by 
by (redisplay-frame-pane <application-frame> <menu-pane> :force-p nil).  

I've done enough poking around to determine that display-comand-menu eventually calls formatting-item-list and that in  the output-record hierarchy produced by this the text-records consist of a sequence of one-character records whose x remains 0 and whose y increases (i.e. it's writing the string vertically).  

I'm not sure what's causing this.

dkochmanski:
That's what most probably is happening:

- space allocation protocol is invoked too late (see #748), only after initial display
- menu pane width is 0 and end-of-line-action is `:wrap`, so to prevent infinite recursion it wraps after at least one character printed

dkochmanski:
workaround would be ensuring, that eol-action is `:allow`, proper fix requires reviewing space allocation code flow and fixing that.

heshrobe:
Thanks Daniel,

Sort of what I thought vaguely.  I’ll look more into it more (once I’m less frustrated).

> On Apr 13, 2019, at 12:50, Daniel Kochmanski <notifications@github.com> wrote:
> 
> That's what most probably is happening:
> 
> 	• space allocation protocol is invoked too late (see #748), only after initial display
> 	• menu pane width is 0 and end-of-line-action is :wrap, so to prevent infinite recursion it wraps after at least one character printed
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

Howard Shrobe

Principal Research Scientist and
Director CyberSecurity@CSAIL
MIT Computer Science and Artificial Intelligence Laboratory
hes@csail.mit.edu
617-253-7877 work
617-312-6439 cell

heshrobe:
I've spent a lot of time trying to narrow the problem down.  Here's a much more succinct description:

Redipslay-frame-pane calls window-clear (if :force-p is t).

Window-clear on the command-menu-pane leaves the mirror-region of the pane set to x 0:1 y 0:1.

At that point trying to do output wraps after the first character.  Here's a simple test case:

[...]

PR #759, opened 2019-04-19
Add forms to handle command menu panes in same way as other panes.  F…

heshrobe:
This addresses issue #754:

In an application frame, with command menu whose display function is clim:display-command-menu: Initially the frame displays incorrectly with the command menu items text printing vertically and overwriting the pane above the menu.

You can get the same behavior by calling
(redisplay-frame-pane :force-p t).

You can then get the display correct by issuing any command in the application-frame or by
by (redisplay-frame-pane :force-p nil).

The ultimate cause of the problem is that command-menu panes are handled differently than other standard pane types as the application-frame definition is expanded  (probably an error of omission).  This results in the command-menu-pane having a different class of pane as its parent and this class mishandles the resize request leading to the pane having a mirror-region of size 0:1, 0:1.

The fix simply provides the code to treat command-menu-panes like other panes:

In frames.lisp:
```
(defun do-pane-creation-form (name form)
  (cond
    ;; Single form which is a function call
    ((and (= (length form) 1)
	  (listp (first form)))
     `(coerce-pane-name ,(first form) ',name))
    ;; Standard pane denoted by a keyword (i.e `:application')
    ((keywordp (first form))
     (case (first form)
       (:application `(make-clim-application-pane
                       :name ',name
                       ,@(cdr form)))
       (:interactor `(make-clim-interactor-pane
                      :name ',name ,@(cdr form)
                      ,@(cdr form)))
       (:pointer-documentation `(make-clim-pointer-documentation-pane
                                 :name ',name
                                 ,@(cdr form)))
       **(:command-menu `(make-clim-command-menu-pane
       			:name ',name
       			,@(cdr form)))**
       (otherwise `(make-pane ,(first form) :name ',name ,@(cdr form)))))
    ;; Non-standard pane designator fed to the `make-pane'
    (t `(make-pane ',(first form) :name ',name ,@(cdr form)))))
```
and in panes.lisp

```
(defun make-clim-command-menu-pane (&rest options)
  (apply #'make-clim-stream-pane :type 'command-menu-pane options))
```

### DIFF
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -730,6 +730,9 @@ documentation produced by presentations.")
        (:pointer-documentation `(make-clim-pointer-documentation-pane
                                  :name ',name
                                  ,@(cdr form)))
+       (:command-menu `(make-clim-command-menu-pane
+       			:name ',name
+       			,@(cdr form)))
        (otherwise `(make-pane ,(first form) :name ',name ,@(cdr form)))))
     ;; Non-standard pane designator fed to the `make-pane'
     (t `(make-pane ',(first form) :name ',name ,@(cdr form)))))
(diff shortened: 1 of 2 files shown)


dkochmanski:
lgtm, thanks!

PR #761, opened 2019-04-20
Fix pane spacing specifications that use :line and :character

heshrobe:
Currently a pane specification that uses :height '(n :line) or :width '(n character) causes an error.  This is because the function spacing-value-to-device-units  (in panes.lisp) calls text-style-character-width or stream-line-height on the pane.  The problem is that the pane can be a composite pane (particularly if there are scroll bars or borders) and these panes don't handle the method.  The fix is to maximize the heights (or widths) of the children panes.

### DIFF
--- a/Core/clim-core/panes.lisp
+++ b/Core/clim-core/panes.lisp
@@ -876,11 +876,28 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
            (:pixels (car x))
            (:point  (* (car x) (graft-pixels-per-inch (graft pane)) 1/72))
            (:mm     (* (car x) (graft-pixels-per-millimeter (graft pane))))
-           (:character (* (car x) (text-style-character-width (pane-text-style pane)
-                                                               (sheet-medium pane)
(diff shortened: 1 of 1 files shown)


dkochmanski:
there is no need to introduce auxiliary protocol on top of auxiliary protocol. here's a pull request which doesn't introduce new generic functions: #763.

heshrobe:
You’re right about that.  I somehow missed that spacing-value-to-device-units was a method (since none of its arguments are qualified).  I think 763 looks fine.  I can test it if you’d like.

> On Apr 23, 2019, at 06:04, Daniel Kochmanski <notifications@github.com> wrote:
> 
> there is no need to introduce auxiliary protocol on top of auxiliary protocol. here's a pull request which doesn't introduce new generic functions: #763.
> 
> —
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

Howard Shrobe

Director CyberSecurity@CSAIL
hes@csail.mit.edu
617-253-7877 work
617-312-6439 cell

Issue #751, opened 2019-04-10
Loss of command table on redefining application frame


heshrobe:
If you have defined an application frame with a command table, then define commands for that application and then redefine the application-frame (e.g. when working with slime) then you lose all the commands.

The fix to this is straightforward: In the macro-expansion of define-application-frame, you always generate this form: (clim:define-command-table <table-name>).  This should be changed to (unless (clim:find-command-table <table-name> :errorp nil) (clim:define-command-table <table-name>))

dkochmanski:
I'm not sure if we are allowed to that according to the spec: http://bauhh.dyndns.org:8000/clim-spec/28-2.html#_1480

```
:command-table name-and-options, where name-and-options is a list consisting of the name of the application frame's command table followed by some keyword-value pairs. The keywords can be :inherit-from or :menu (which are as in define-command-table). The default is to create a command table with the same name as the application frame. 
```

heshrobe:
I think the spec is silent on the issue of what happens when an application-frame is redefined.  II can look at the original CLIM’s implementation to see what it does.  But it definitely had the behavior of not losing the menu items in the command table on redefinition of the application-frame.  The code might need to be more complicated than my suggestion due to the options part.  If the options change, maybe the implication is that you’d need to make a new command table.  The core point for me was that losing my menu items when I redefine an application-frame during development is a major pain in the posterior.

> On Apr 10, 2019, at 12:25, Daniel Kochmanski <notifications@github.com> wrote:
> 
> I'm not sure if we are allowed to that according to the spec: http://bauhh.dyndns.org:8000/clim-spec/28-2.html#_1480
> 
> :command-table name-and-options, where name-and-options is a list consisting of the name of the application frame's command table followed by some keyword-value pairs. The keywords can be :inherit-from or :menu (which are as in define-command-table). The default is to create a command table with the same name as the application frame. 
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

Howard Shrobe

Principal Research Scientist and
Director CyberSecurity@CSAIL
MIT Computer Science and Artificial Intelligence Laboratory
hes@csail.mit.edu
617-253-7877 work
617-312-6439 cell

heshrobe:
On further consideration, this really comes to what the behavior of define-comand-table is.  Define-application-frame simply generates a call to define-command-table with all the arguments provided.  

[...]

PR #762, opened 2019-04-20
Fix the bug where define-command-table on an existing table throws aw…

heshrobe:
…ay the menu items. Issue #751

Define-command-table loses old commands when redefining the command table.  Define-application-frame expands into (among many other things) an invocation of define-command-table so if during development the application-frame is redefined, you lose all the commands in the command menu.

### DIFF
--- a/Core/clim-core/commands.lisp
+++ b/Core/clim-core/commands.lisp
@@ -223,19 +223,27 @@ designator) inherits menu items."
 
 (make-command-table 'user-command-table)
 
-(defmacro define-command-table (name &key inherit-from menu inherit-menu)
-  `(let ((old-table (gethash ',name *command-tables* nil))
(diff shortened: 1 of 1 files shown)


Core/clim-core/commands.lisp
```diff
@@ -223,19 +223,27 @@ designator) inherits menu items."
...
+	     ,@(when menu-supplied-p
+		`((setq menu (menu-items-from-list ',menu))))
+	     old-table)
+	   (make-command-table ',name
+			       :inherit-from ,(if inherit-supplied-p
+						  `',inherit-from-arg
+						  `'(global-command-table))
+			       :inherit-menu ,inherit-menu
+			       :menu ',menu
+			       :errorp nil)))))
```
dkochmanski: please see comment in the linked issue.

PR #765, opened 2019-05-19
This makes the menu behave more like menus of other gui frameworks

DanielKeogh:
An open menu will now no longer close during the first click when the mouse-button is released unless it is released over a button with a command.
Clicking on an open menu will close it, as will clicking outside of the menu.

To support this, it was necessary to track which menu was open.
The active menu is now tracked on the standard-application-frame in a slot named 'active-menu'.
I figured that this was an ok design choice, as the standard-application-frame already has two menu-specific slots on it.


Core/clim-core/menu.lisp
```diff
@@ -170,7 +175,10 @@ account, and create a list of menu buttons."
 	    (adopt-frame manager submenu-frame)
 	    (enable-frame submenu-frame)
 	    (with-sheet-medium (medium raised)
-	      (medium-force-output medium))))))))
+	      (medium-force-output medium))))))
+    (when frame
```
dkochmanski: functions earlier would signal a condition if frame is  nil, so there is no need for this test

Core/clim-core/menu.lisp
```diff
@@ -50,6 +50,11 @@
       (dispatch-repaint button (sheet-region button))
       (stream-force-output button))))
 
+(defgeneric toggle-arm-branch (button))
+
+(defmethod toggle-arm-branch ((button menu-button-pane))
+  (arm-menu button))
+
```
dkochmanski: please rename arm-menu function to toggle-arm-branch and merge their behaviors. also write a comment above defgeneric form which describe how we achieve desired behavior.

Core/clim-core/menu.lisp
```diff
@@ -179,7 +187,10 @@ account, and create a list of menu buttons."
       (disown-frame frame-manager submenu-frame)
       (disarm-gadget sub-menu)
       (dispatch-repaint sub-menu +everywhere+)
-      (setf submenu-frame nil) )))
+      (setf submenu-frame nil)))
+  (with-application-frame (frame)
+    (when frame
+      (setf (slot-value frame 'active-menu) nil))))
```
dkochmanski: make active-menu an accessor on frame (here you use setf on slot-value and earlier you use with-slots - this is not consistent).

Core/clim-core/menu.lisp
```diff
@@ -192,8 +203,19 @@ account, and create a list of menu buttons."
...
-(defmethod handle-event ((pane menu-button-submenu-pane) (event pointer-button-release-event))
-  (destroy-substructure (menu-root pane)))
+(defmethod toggle-arm-branch ((button menu-button-submenu-pane))
+  (if (slot-value button 'armed)
+      (destroy-substructure button)
+      (arm-branch button)))
+
+(defmethod handle-event :before (pane (event pointer-button-release-event))
+  (when (not (or (typep event 'window-repaint-event)
+		  (typep event 'pointer-motion-event)))
```
dkochmanski: s/when not/unless/

Core/clim-core/menu.lisp
```diff
@@ -192,8 +203,19 @@ account, and create a list of menu buttons."
...
+    (with-application-frame (frame)
+      (when (and (not (typep pane 'menu-button)) frame)
+	(with-slots (active-menu) frame
+	  (when active-menu
+	    (destroy-substructure (menu-root active-menu))))))))
```
dkochmanski:
```
(unless (typep pane 'menu-button)
  (with-application-frame (frame)
    (when (and frame (active-menu frame))
      (destroy-sub …)
```

Core/clim-core/frames.lisp
```diff
@@ -113,6 +113,7 @@ input focus. This is a McCLIM extension."))
 	     :initform nil)
    (menu-bar-pane :initform nil
                   :accessor frame-menu-bar-pane)
+   (active-menu :initform nil)
```
dkochmanski: make it an accessor %frame-active-menu

DanielKeogh:
Will do these things. Thanks for review.

After this I fill fix the weird double-open that occurs when you mouse-down a menu-item that has no sub-menu, and then drag the mouse over a menu-item that does, then release.
(e.g. in menu demo, press down on right menu, then release over left.)

Without this pullrq you are left with an open menu that can only be closed by clicking on it directly.

Core/clim-core/menu.lisp
DanielKeogh: Can't merge arm-menu and toggle-arm-branch.
arm-menu when menu is already armed. toggle-arm-branch is solely for pointer-button-press-event.

I can remove toggle-arm-branch by overriding handle-event for menu-button-submenu-pane.

Core/clim-core/frames.lisp
DanielKeogh: Done.

Core/clim-core/menu.lisp
DanielKeogh: Done

Core/clim-core/menu.lisp
DanielKeogh: Okies

Core/clim-core/menu.lisp
DanielKeogh: In the other place I accessed the slot twice. Here only once. I try to always use with-accessors/with-slots in any place where there are multiple uses, and avoid it when there is only one.
I changed this to use the accessor in this case.

Core/clim-core/menu.lisp
DanielKeogh: Changed, thanks.

DanielKeogh:
I left two feedbackos unresolved.
I'm not too happy with the existence of:
`(defmethod handle-event :before (pane (event pointer-button-release-event)) ...)`
Given that it adds overhead to every pointer-button-release-event, but like I said in the irc channel, there aren't any hooks for adding temporary event handlers at this stage.

Core/clim-core/menu.lisp
```diff
@@ -192,8 +196,17 @@ account, and create a list of menu buttons."
...
     (arm-menu sub-menu)))
 
-(defmethod handle-event ((pane menu-button-submenu-pane) (event pointer-button-release-event))
-  (destroy-substructure (menu-root pane)))
+(defmethod handle-event ((pane menu-button-submenu-pane) (event pointer-button-press-event))
+  (if (slot-value pane 'armed)
+      (destroy-substructure pane)
+      (arm-branch pane)))
+
+(defmethod handle-event :before (pane (event pointer-button-release-event))
```
dkochmanski: why isn't this specialized on pane as `menu-button-pane`?

dkochmanski: then handle-event would trigger only for panes of this type and there would be no unnecessary overhead

dkochmanski: ah, nevermind.

dkochmanski:
I'm wondering if it wouldn't be a good idea to make menu be responsive to keyboard events (as in `port-keyboard-input-focus`), then we could use our (currently malfunctioning) protocol `note-input-focus-changed` and specialize for menu, so when it loses focus then it is closed. As a bonus we could implement navigating menu with arrows. What do you think?

nb: I've tested your branch and it works as advertised.  I have too much on my plate to work on what I've suggested above, so it is up to you: should we merge the branch or work on it further?

DanielKeogh:
Clearly, menu navigation with arrows is a good idea.

Let me think about it and come back. I will need to think about is whether or not focus-change is the right abstraction for this to be hooked on.

I am expecting that the "focus" abstraction is not clearly defined as, "focus changes when something is clicked on", and some of the time focus changed events could be raised by a mouse-over or other shortcuts. (It would be extremely annoying if the menu was closed because of this)

If this is the case, then we would want something like:
`(defgeneric can-yield-focus (event-causing-focus-change item-with-focus item-to-be-focused))`
This would let us refuse to yield focus under certain circumstances. Notably, this would be useful for implementing modal dialog (pop-up that stops you from clicking the window behind it).

I will look at how this protocol works on the weekend and decide.

dkochmanski:
I don't think we need to introduce new protocols. if menu loses keyboard focus then it should close. currently keyboard focus only changes upon click.

nb: I'm anxious about introducing more protocol complexity onto what we have now. One of my hazy plans for the future is to build a chinese walls between basic backend abstraction, windowing abstraction and application abstraction (ports/grafts vs sheets/streams vs frames/panes), so one is built on top of another instead of having a monolith.

DanielKeogh:
The menu changes are right, I think.

The change to focus is very aggressive. I also haven't looked at the document on how focus is supposed to work yet.

dkochmanski:
I'm closing this for now, if the pull request reemerge it should be done after we sort out #851.

PR #767, opened 2019-05-22
bug fix subset-alist definition

admich:
Fix a bug in definition of `sublist-alist` presentation abbreviation.


dkochmanski:
thanks!

PR #770, opened 2019-05-29
Cleanup translators

dkochmanski:
This pull request is a prerequisite for the selection extension. It is required to allow translating copied elements (i.e images).

First it extracts presentation-translators code from presentations and then some refactor follows. Some issues are fixed along the way. Main benefit of this change is improved readability.


Core/clim-core/presentation-translators.lisp
```diff
@@ -450,46 +450,37 @@ and used to ensure that presentation-translators-caches are up to date.")
...
+             (funcall (tester translator) (presentation-object presentation)
+                      :presentation presentation :context-type context-type
+                      :frame frame :window window :x x :y y :event event))
+         (or (tester-definitive translator)
+             (null (decode-parameters context-type))
+             (presentation-typep
+              (call-presentation-translator translator presentation context-type
+                                            frame event window x y)
+              context-type))
+         t)))
```
scymtym: The final `t` should not be needed.

Core/clim-core/presentation-translators.lisp
```diff
@@ -482,53 +482,38 @@ and used to ensure that presentation-translators-caches are up to date.")
...
+                  when (and (or (not for-menu) (eql for-menu (menu translator)))
+                            (test-presentation-translator
+                             translator presentation context-ptype
+                             frame window x y
+                             :event event :modifier-state modifier-state
+                             :for-menu for-menu :button button))
+                  do (funcall func translator presentation context))))
+           (mopscp (context record)
+             "maps recursively over all presentations in record, including record."
+             (map-over-output-records (curry #'mopscp context) record)
```
scymtym: This changes `map-over-output-records-containing-position` to `map-over-output-records`. Are you sure this is an improvement? I'm asking because methods on `map-over-output-records-containing-position` could have a more efficient implementation than linear iteration with position-based filtering.

Core/clim-core/presentation-translators.lisp
dkochmanski: original code returned either NIL or T, if we skip T then it will return a generalized boolean. It makes sense and does not contradict the spec, it is just that this change doesn't change function semantics. Should I change it?

scymtym: No, it's fine.

Core/clim-core/presentation-translators.lisp
dkochmanski: you are right, I've changed the code

dkochmanski:
I've added one more commit which fixed the issue where the same translators were mapped on the same presentation multiple times. This could happen i.e when we have defined command as a menu command (then it would be in different input contexts and call-menu had duplicated entries).

dkochmanski:
I've fixed the commit message.

regarding floor I've added a comma (which is clearly needed)

`Despite the comment above the local function, floor was called without`

Core/clim-core/presentation-translators.lisp
```diff
@@ -483,14 +483,23 @@ and used to ensure that presentation-translators-caches are up to date.")
...
+                                        context-ptype
+                                        (frame-command-table frame))))
+               (setf maybe-translators
+                     (remove-if (lambda (translator)
+                                  (member (cons translator presentation)
+                                          processed
+                                          :test #'equal))
+                                maybe-translators))
+               (loop
+                  for translator in maybe-translators
```
scymtym: would 
```lisp
for key = (cons translator presentation)
unless (member key processes :test #'equal)
do (push key processed)
and …
```
instead of the `remove-if` work? Seems simpler and less consy.
```

Comment (dkochmanski) [reply Core/clim-core/presentation-translators.lisp]: I'm a poor loop reader, that would be
```
unless
do
and
when
do
```

how about

```
(let* ((context-ptype (first context))
                    (maybe-translators (find-presentation-translators
                                        (presentation-type presentation)
                                        context-ptype
                                        (frame-command-table frame))))
               (dolist (translator maybe-translators)
                 (let ((key (cons translator presentation)))
                   (unless (member key processed :test #'equal)
                     (push key processed)
                     (when (and (or (not for-menu) (eql for-menu (menu translator)))
                                (test-presentation-translator
                                 translator presentation context-ptype
                                 frame window x y
                                 :event event :modifier-state modifier-state
                                 :for-menu for-menu :button button))
                       (funcall func translator presentation context))))))
```

scymtym: Fine with me.

dkochmanski: fixed

dkochmanski:
I've retracted the change to map-applicable-translators and applies it to call-presentation-menu (see commit message)

scymtym:
> I've retracted the change to map-applicable-translators and applies it to call-presentation-menu (see commit message)

Looks good to me (apart from stylistic choices, but I can accept those).

PR #777, opened 2019-06-06
clim-fig example: fix undo behaviour.

kc-:
Actions get pushed onto the history, com-undo should behave like a
FIFO. The previous code undid the actions in the order they were performed.

### DIFF
--- a/Examples/clim-fig.lisp
+++ b/Examples/clim-fig.lisp
@@ -1,5 +1,4 @@
 ;;; -*- Mode: Lisp; Package: CLIM-DEMO -*-
-
 ;;;  (c) copyright 2001 by
 ;;;           Arnaud Rouanet (rouanet@emi.u-bordeaux.fr)
 ;;;           Lionel Salabartan (salabart@emi.u-bordeaux.fr)
@@ -157,11 +156,13 @@
       (set-status-line " ")
       (copy-from-pixmap canvas-pixmap 0 0 pixmap-width pixmap-height pane 0 0)
       (deallocate-pixmap canvas-pixmap)
-      (with-output-as-presentation (pane nil 'figure
-                                         :single-box t)
-        (draw-figure pane x1 y1 x y
-		     :cp-x1 cp-x1 :cp-y1 cp-y1
-		     :cp-x2 cp-x2 :cp-y2 cp-y2))
+      (let ((new-record (with-output-to-output-record (*standard-output*)
+                          (draw-figure pane x1 y1 x y
+		                       :cp-x1 cp-x1 :cp-y1 cp-y1
+		                       :cp-x2 cp-x2 :cp-y2 cp-y2))))
+        (push new-record (clim-fig-undo-list *application-frame*))
+        (stream-add-output-record *standard-output* new-record)
+        (replay new-record *standard-output*))
       (setf (clim-fig-redo-list *application-frame*) nil))))
 
 (defun handle-move-object (pane figure first-point-x first-point-y)
@@ -217,6 +218,7 @@
 (define-application-frame clim-fig ()
   ((drawing-mode :initform :line :accessor clim-fig-drawing-mode)
    (output-record :accessor clim-fig-output-record)
+   (undo-list :initform nil :accessor clim-fig-undo-list)
    (redo-list :initform nil :accessor clim-fig-redo-list)
    (current-color :initform +black+ :accessor clim-fig-current-color)
    (line-style :initform (make-line-style) :accessor clim-fig-line-style)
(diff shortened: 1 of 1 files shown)


Examples/clim-fig.lisp
```diff
@@ -375,8 +375,7 @@
   (let* ((output-history (clim-fig-output-record *application-frame*))
 	 (record-count (output-record-count output-history))
          (record (and (not (zerop record-count))
-		      (elt (output-record-children output-history)
-			   (1- record-count)))))
+		      (first (output-record-children output-history)))))
```
dkochmanski: This fix is not correct, because it depends on assumption, that output-record-children are in some particular order. There is no guarantee for that. In fact this code is correct for standard-sequence-output-record (where add is an operation `vector-push-extend`). In case of tree-output-record (for some reason) it is reversed and that may require further investigation. It is not stated in the spec what is the order (http://bauhh.dyndns.org:8000/clim-spec/16-2.html#_859).

Also record-count in this commit is left unusued (then there would be no need to investigate).

A correct fix would be collecting output records on a separate list where the order is known and access them from there (redo command does something in this spirit).

To have access to the output-record when it is drawn you need to use with-output-to-output-record and then add it to the stream.

That's at least my impression how this should be tackled.

kc-: Okay, 2f3fa9c fixes the behaviour without any dependence on the internals of tree-output-record, per your suggestion.

dkochmanski: please squash commits so there is only one.

Examples/clim-fig.lisp
```diff
@@ -372,21 +375,18 @@
...
 
 (define-clim-fig-command com-undo ()
-  (let* ((output-history (clim-fig-output-record *application-frame*))
-	 (record-count (output-record-count output-history))
-         (record (and (not (zerop record-count))
-		      (elt (output-record-children output-history)
-			   (1- record-count)))))
-    (if record
+  (let ((latest-record (pop (clim-fig-undo-list *application-frame*))))
+    (if latest-record
```
dkochmanski: `(if-let ((latest-record (pop …))) …)`

Examples/clim-fig.lisp
```diff
@@ -372,21 +375,18 @@
...
         (progn
-          (erase-output-record record *standard-output*)
-          (push record (clim-fig-redo-list *application-frame*)))
+          (erase-output-record latest-record *standard-output*)
+          (push latest-record (clim-fig-redo-list *application-frame*)))
         (beep))))
 
 (define-clim-fig-command com-redo ()
   (let* ((record (pop (clim-fig-redo-list *application-frame*))))
     (if record
```
dkochmanski: if-let

dkochmanski:
this looks good, thank you. please replace (if (let )) with if-let (available in alexandria) and squash commits.

dkochmanski:
found while testing: I can't add ellipse to the canvas. If I use version from before this commit I can.

dkochmanski:
you've also missed clearing undo list in com-clear

kc-:
You're absolutely right.
I also have a question about this command -- is it really the intended behaviour that after, for example:

 - add rectangle
 - add line
 - add ellipse
 - press CLEAR,

you use REDO three times to restore the state before having pressed CLEAR? In other words, shouldn't CLEAR produce a compound undo containing all records of the canvas?

dkochmanski:
it should, yes

kc-:
All right, then I propose these changes:

`(define-clim-fig-command com-undo ()
  "Undo the previous command, which might have been either 'draw a new object', or the CLEAR command.
   In the former case, remove the record and add it to the redo list; in the latter case, replay the
   output-history."
  (let ((latest-record (pop (clim-fig-undo-list *application-frame*))))
    (cond
      ((listp latest-record)
       (loop for record in latest-record do
            (stream-add-output-record *standard-output* record)
            (replay record *standard-output* (bounding-rectangle record))))
      ((clim-internals::output-record-ancestor-p
        (stream-output-history *standard-output*) latest-record)
       (erase-output-record latest-record *standard-output*)
       (push latest-record (clim-fig-redo-list *application-frame*)))
      (T (beep)))))`

`(define-clim-fig-command com-clear ()
  (setf (clim-fig-undo-list *application-frame*)
        (list (coerce (output-record-children (clim-fig-output-record
						 *application-frame*))
			'list)))
  (setf (clim-fig-redo-list *application-frame*) (list))
  (window-clear *standard-output*))`

dkochmanski:
lgtm, thank you!

kc-:
Unfortunately, I missed that simply adding records to a pane does not create a presentation. To be honest, I was not even aware that clim-fig allowed editing objects after they had been drawn -- I was alerted to this on IRC.
b0a1a77 fixes the oversight.

dkochmanski:
please make a pull request from this branch, I can't "remerge" already merged branch.

Issue #783, opened 2019-06-21
automatic-clx-server-path doesn't parse all conceivable paths correctly.


kc-:
`automagic-clx-server-path` does not function properly unless there's a ':' in that string.

On my machine (macos with XQuartz), `$DISPLAY` is set to `computer-name.network-name`, and the fault leads to the `use-localhost` restart.

Since the `use-localhost` restart calls `parse-clx-server path` with the empty string as `:host` argument, so automagic-clx-server-path should accept the empty string when the environment variable DISPLAY is set to it, as well.

4ce5e0e fixes this.

dkochmanski:
fixed by recent merge.

PR #784, opened 2019-06-21
Fix Issue #783, automagic-clx-server-path

kc-:
The function supposes an occurence of at least one #\: in the DISPLAY
environment variable. This isn't guaranteed (for example, with XQuartz, DISPLAY
is set to computer-name.network-name), and in fact the use-localhost
restart of helpfully-automagic-clx-server-path passes an empty string onwards.


dkochmanski:
lgtm, thanks!

Issue #712, opened 2019-02-14
[$1000] streams and gadgets doesn't mix well


dkochmanski:
The problem is how events are processed. Currently stream-input and event queue are the same object, so streams "steal" input from the pane (and handle-event can't be used to its fullest). That in practice leads to a situation, where we can't have native gadgets with presentations.

I have plans for solving this issue (I'll start around 2-3 months from now - 2019-02-15), but if someone else wants to solve it before that please go ahead (I'm not calling dibs :-).

pouar:
I'm going to guess this is why throw-highlighted-presentation doesn't work inside of activate-callback

pouar:
Ok, apparently my issue was apparently unrelated to this bug, as the gadgets for some reason in my case were simply weren't working with incremental redisplay.

pouar:
Should I open a separate bug report?

PR #785, opened 2019-06-21
[wip] input-refactor: separate input-buffer from event-queue

dkochmanski:
We remove invalid assumption that input-buffer is same thing as
event-queue for standard-extended-input-stream. To do that we split
the initargs and slots between standard-sheet-input-mixin and
standard-input-stream (along with appropriate accessors).

This change is incomplete, because stream reading operations still
assume that the input-buffer is an event-queue what leads to debugger
popping all the time. To make this commit atomically complete we
replace passages in the code which make this invalid assumption with
event-sheet-queue access (and add appropriate comment that this is
invalid). This will be fixed in upcoming commit. Software after this
change will work as before except that SEIS will signal unbound-slot
when we try to access STREAM-INPUT-BUFFER from the user code (access
from McCLIM has been replaced).

- add :type vector to the input-buffer slots
- untie input-buffer and event-queue
- remove stream-input-buffer references from sheet-input
- replace stream-input-buffer with sheet-event-queue where appropriate
- minor refactor of some code passages (stylistic)

### DIFF
--- a/Core/clim-basic/input.lisp
+++ b/Core/clim-basic/input.lisp
@@ -453,17 +453,10 @@ use condition-variables nor locks."))
                         (make-instance 'concurrent-event-queue)
                         (make-instance 'simple-event-queue))
           :reader sheet-event-queue
-          :initarg :input-buffer)
+          :initarg :event-queue)
    (port :initform nil
          :initarg :port
          :reader port)))
-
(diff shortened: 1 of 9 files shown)


dkochmanski:
Related to #712.

dkochmanski:
this branch is obsolete (see refactor-input-streams for ongoing progress)

Issue #742, opened 2019-03-22
make-clim-stream: space value is not based on inner stream


dkochmanski:
Steps to reproduce:

Compile the following:
```
(in-package #:clim-user)
(define-application-frame bug-xxx () ()
  (:pane :application  :width '(64 :character) :height '(16 :line)))
(run-frame-top-level (make-application-frame 'bug-xxx))
```

Expected: single pane of width 64x16
Observed:
```
There is no applicable method for the generic function
  #<STANDARD-GENERIC-FUNCTION CLIM:STREAM-LINE-HEIGHT (2)>
when called with arguments
  (#<CLIM-CLX::CLX-876-BORDER-PANE "(Unnamed Pane)" {10066E0933}>).
   [Condition of type SB-PCL::NO-APPLICABLE-METHOD-ERROR]
See also:
  Common Lisp Hyperspec, 7.6.6 [:section]
```

The issue is caused by the fact that application-pane is wrapped in bordered pane and scrolling pane, if we add `:borders nil :scroll-bars nil` debugger won't be invoked and application will start without a problem. That said it is only a symptom.

`width` and `height` are passed to the border (or scroller) pane instead of the inner application pane, so during compose-space user-specified options are merged against one of the panes upon creation. That calls in turn `spacing-vaue-to-device-units`:

```
(defmethod spacing-value-to-device-units (pane x)
  (cond ((realp x) x)
        ((consp x)
         (ecase (cadr x)
           (:pixels (car x))
           (:point  (* (car x) (graft-pixels-per-inch (graft pane)) 1/72))
           (:mm     (* (car x) (graft-pixels-per-millimeter (graft pane))))
           (:character (* (car x) (text-style-character-width (pane-text-style pane)
                                                               (sheet-medium pane)
                                                               #\m)))
           (:line  (* (car x)
                      (stream-line-height pane)))))))
```

`:line` option depends on stream having line-height and of course bordered-pane doesn't have it. Arguably that should depend on `text-style-height` (think about tit in this ticket!). `:width` "just" passes because even bordered-pane seems to have text-style associated with it, but that is equally wrong.

Here comes `make-clim-stream-pane`:
```
(defun make-clim-stream-pane (&rest options
                              &key (type 'clim-stream-pane)
                                (scroll-bar :vertical)
                                (scroll-bars scroll-bar)
                                (borders t)
                                &allow-other-keys)
  (with-keywords-removed (options (:type :scroll-bar :scroll-bars :borders))
[...]

PR #786, opened 2019-06-25
Rename the spacing unit :pixels to :pixel (discussed in issue #742).

kc-:
The unit names are specified in http://bauhh.dyndns.org:8000/clim-spec/17-2.html#_922.


dkochmanski:
lgtm, thank you!

PR #787, opened 2019-06-27
Fix drei tests

kc-:
Preparing for McCLIM#779, I dusted off the drei test suite.

Three failing tests remain (both to do with `drei-lisp-syntax::arglist-for-form`): for argument lists of macros and generic methods, default arguments are not returned by this function -- but the tests indicate that they should be, or once were (I only ran tests on SBCL 1.5.3).

I'd like some feedback on this, please.

To run the tests:
```
(ql:quickload "drei-mcclim/test")
(drei-tests:run-tests)
```
or, if you want to reproduce this without running the entire test suite:

```
(ql:quickload "drei-mcclim/test")
(5am:run! 'drei-tests::make-instance-form-traits)
(5am:run! 'drei-tests::arglist-for-form-1)
```


Libraries/Drei/Tests/testing.lisp
```diff
@@ -101,6 +101,10 @@ display. Used for testing.")
            (with-bound-drei-special-variables (,drei :minibuffer nil)
              ,@body))))))
 
+(defmethod climb:publish-selection (port selection object object-type)
+  "Empty implementation to allow with-drei-environment to contain editing operations."
+  (declare (ignore port selection object object-type)))
```
dkochmanski: this should be at all times specialized on the first argument (at least). you may use eql-specializer for instance (see the bottom of file Tests/tests.lisp). given it is not used at all here maybe just remove the method from PR. other things look good to me.

kc-: > given it is not used at all here maybe just remove the method from PR. other things look good to me.

That function is called from all tests that edit a buffer, I'll paste from the first failed test:

- Without having loaded `mcclim`:
```
 BACKWARD-KILL-PARAGRAPH []: 
      Unexpected Error: #<UNDEFINED-FUNCTION PUBLISH-SELECTION {100456A493}>
The function CLIM-EXTENSIONS:PUBLISH-SELECTION is undefined...
```

- After having loaded `mcclim`:
```
 BACKWARD-KILL-PARAGRAPH []: 
      Unexpected Error: #<SB-PCL::NO-APPLICABLE-METHOD-ERROR {100A3C9503}>
  #<STANDARD-GENERIC-FUNCTION CLIM-EXTENSIONS:PUBLISH-SELECTION (4)>
when called with arguments
  (NIL :CLIPBOARD "This is the last paragraph." STRING).
See also:
  The ANSI Standard, Section 7.6.6..
```

dkochmanski: then it should be fixed in drei to not call publish-selection if there is no sheet. fix will probably look like this (in kill-ring):

```
(defun push-start-and-copy (chain vector)
  (push-start chain vector)
  (when (every #'characterp vector)
    (when-let ((pane (drei:editor-pane (drei:drei-instance))))
      (clime:publish-selection pane :clipboard (coerce vector 'string) 'string))))
```

(previously we did not call when-let but only let)

kc-: Okay, I made the change.

dkochmanski:
lgtm, thanks!

PR #788, opened 2019-06-27
refactor 712 partial

dkochmanski:
This change unifies wait-function interpretation across different APIs and removes recursive locks from the concurrent-event-queue. Some minor fixes follow. This change is part of work for #712 without any stream-related changes (which are work in progress).

### DIFF
--- a/Backends/CLX/clipboard.lisp
+++ b/Backends/CLX/clipboard.lisp
@@ -105,14 +105,28 @@
         (to-type (selection-object-type requested-object))
         (table (climi::get-object-table requested-object)))
     (labels ((wait-for-request ()
-               (or (not (clipboard-outstanding-request port))
-                   (not (eq selection (selection-object-content requested-object)))))
+               (flet ((wait-fn ()
+                        (or (not (clipboard-outstanding-request port))
+                            (not (eq selection (selection-object-content requested-object))))))
+                 (declare (dynamic-extent #'wait-fn))
+                 ;; event-listen-or-wait will return on any event
+                 ;; becoming available in the queue, that's why we
+                 ;; loop until our condition is met (or timeout).
+                 (loop
+                    with timeout-time = (1+ (climi::now))
+                    for decay = (- timeout-time (climi::now))
+                    do
+                      (cond ((wait-fn)   (return t))
+                            ((< decay 0) (return nil))
+                            (t (event-listen-or-wait sheet
+                                                     :wait-function #'wait-fn
+                                                     :timeout decay))))))
              (negotiate-selection-type ()
                (xlib:convert-selection selection :targets window :mcclim nil)
-               (event-listen-or-wait sheet :wait-function #'wait-for-request :timeout 1)
                ;; Either window is closed or peer does not understand
                ;; targets. We will try to request string in that case.
-               (unless (clipboard-outstanding-request port)
+               (unless (and (wait-for-request)
+                            (clipboard-outstanding-request port))
                  (setf (clipboard-outstanding-request port) requested-object
                        (selection-object-content requested-object) '(:utf8_string)))
                ;; We look for the first proposed type which is suitable for the
@@ -134,12 +148,12 @@
       (when-let ((from-type (selection-object-type requested-object)))
         (setf (selection-object-content requested-object) selection)
         (xlib:convert-selection selection from-type window :mcclim nil)
-        (when (event-listen-or-wait sheet :wait-function #'wait-for-request :timeout 1)
-          (when-let ((result (clipboard-outstanding-request port)))
-            (call-selection-translator (selection-object-content result)
-                                       (selection-object-type result)
-                                       to-type
-                                       table)))))))
+        (when-let ((result (and (wait-for-request)
+                                (clipboard-outstanding-request port))))
+          (call-selection-translator (selection-object-content result)
+                                     (selection-object-type result)
+                                     to-type
+                                     table))))))
 
 (macrolet ((frob (selection)
              `(defmethod request-selection ((port clx-selection-mixin)
--- a/Core/clim-basic/input.lisp
+++ b/Core/clim-basic/input.lisp
@@ -43,16 +43,15 @@
 (declaim (inline now compute-decay))
 (defun now ()
   (/ (get-internal-real-time)
-     1.0
-     internal-time-units-per-second))
+     #.(float internal-time-units-per-second)))
 
 ;; Given two alarm times compute remaining time (in seconds). If both times are
 ;; NIL returns NIL.
 (defun compute-decay (time-1 time-2)
-  (cond ((and time-1 time-2) (- (min time-1 time-2) (now)))
-        (time-1 (- time-1 (now)))
-        (time-2 (- time-2 (now)))
-        (T NIL)))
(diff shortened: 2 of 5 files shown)


dkochmanski:
adding wip tag, something is wrong on ecl with this branch. will investigate it later.

Core/clim-basic/input.lisp
```diff
@@ -288,79 +298,71 @@ use condition-variables nor locks."))
...
+         (check-schedule queue)
+         (with-lock-held (lock)
+           (if-let ((result (%event-queue-read queue)))
+             (return result)
+             (condition-wait cv lock decay))))))
 
 (defmethod event-queue-read-no-hang ((queue concurrent-event-queue))
   (do-port-force-output queue)
-  (with-recursive-lock-held ((event-queue-lock queue))
+  (check-schedule queue)
```
scymtym: See previous comment.

Core/clim-basic/input.lisp
```diff
@@ -383,30 +385,26 @@ use condition-variables nor locks."))
...
-  (with-recursive-lock-held ((event-queue-lock queue))
+  (with-lock-held ((event-queue-lock queue))
     (call-next-method)
     (condition-notify (event-queue-processes queue))))
 
 (defmethod event-queue-peek ((queue concurrent-event-queue))
   (do-port-force-output queue)
-  (with-recursive-lock-held ((event-queue-lock queue))
-    (check-schedule queue)
+  (check-schedule queue)
```
scymtym: Why is this not in the critical section?

Backends/CLX/clipboard.lisp
```diff
@@ -105,14 +105,32 @@
...
         (table (climi::get-object-table requested-object)))
     (labels ((wait-for-request ()
-               (or (not (clipboard-outstanding-request port))
-                   (not (eq selection (selection-object-content requested-object)))))
+               (flet ((wait-fn ()
+                        (or (not (clipboard-outstanding-request port))
+                            (not (eq selection (selection-object-content requested-object)))))
+                      (now ()
+                        (/ (get-internal-real-time)
+                           1.0
```
scymtym: Is this to force float coercion? If so, there has to be a cleaner idiom.

dkochmanski: it is cargo-culted from input.lisp. how about

(/ (float (get-internal-real-time)
   internal-time-units-per-second))

or alternatively:

(/ (get-internal-real-time)
   #.(flaot internal-time-units-per-second))

?

Core/clim-basic/input.lisp
```diff
@@ -288,79 +298,71 @@ use condition-variables nor locks."))
 (defmethod event-queue-peek ((queue simple-event-queue))
   (do-port-force-output queue)
   (check-schedule queue)
-  (or (first (event-queue-head queue))
-      (when (process-next-event (event-queue-port queue) :timeout 0)
-        (first (event-queue-head queue)))))
+  (process-next-event (event-queue-port queue) :timeout 0)
```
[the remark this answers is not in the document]
dkochmanski: it is deliberate choice. in concurrent-event-queue we do not call process-next-event at all (because it runs on its own thread).

This unconditional call is made to ensure event processing progress (i.e we may have a wait function which yields true only after some event is processed by the port, not event queued - like selection-request events).

Also assuming that scheduler appends events as we go, we may never feel a need to call process-next-event causing mouse or keyboard events being dropped.

Core/clim-basic/input.lisp
```diff
@@ -288,79 +298,71 @@ use condition-variables nor locks."))
...
+                     (eql reason :wait-function)))
+              (return (values nil :wait-function)))
+             (t (check-schedule queue)))))
 
 (defmethod event-queue-listen ((queue simple-event-queue))
   (event-queue-listen-or-wait queue :timeout 0))
 
 

+;;; XXX SBCL doesn't reacquire lock when condition-variable
```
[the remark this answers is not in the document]
dkochmanski: new version
```
;;; XXX SBCL doesn't reacquire lock when condition-variable
;;; prematurely returns (i.e due to timeout). That means we can't use
;;; recursive locks for the concurrent-event-queue. --jd 2019-06-28
```

Core/clim-basic/input.lisp
```diff
@@ -288,79 +298,71 @@ use condition-variables nor locks."))
...
   (let ((lock (event-queue-lock queue))
         (cv (event-queue-processes queue)))
     (loop
        as decay = (compute-decay nil (event-schedule-time queue))
-       do (with-recursive-lock-held (lock)
-            (if-let ((result (%event-queue-read queue)))
-              (return result)
-              (condition-wait cv lock decay))))))
+       do
+         (check-schedule queue)
```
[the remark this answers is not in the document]
dkochmanski: because that would cause a deadlock (that's the main reason why the initial implementation had recursive locks).
check-schedule may call event-append which has a critical section on its own, so this call outside the critical section is safe.

One could wonder, if access to the scheduler-mixin slots is safe. It is, because there is only one consumer (read/peek/listen etc.) who calls check-schedule - i.e producent doesn't call it. We always consume queue from the CLIM loop (and it is not exported API except the method `schedule-event` which operates on sheets and these operations are also called from CLIM loop.

Core/clim-basic/input.lisp
```diff
@@ -149,39 +149,49 @@ use condition-variables nor locks."))
...
-    (when decay (maxf decay 0))
-    (loop
-       as result = (%event-queue-read queue)
-       if result do (return result)
-       else do (process-next-event port :timeout decay))))
+  (check-schedule queue)
+  (process-next-event (event-queue-port queue) :timeout 0)
+  (loop
+     with port = (event-queue-port queue)
+     as decay = (event-schedule-time queue)
```
dkochmanski: this is wwonrg, decay is confused with the a schedule time (fixing).

dkochmanski:
wait-function commit changes:
- message commit fixed as suggested
- instead of defining local `now` function we access `climi::now`

event-queue commit changes:
- message improved as suggested
- function `now` looks like this
```
(defun now ()
  (/ (get-internal-real-time)
     #.(float internal-time-units-per-second)))
```
- decay is computed only when needed
- comment about SBCL improved as per suggestion
- issue in concurrent event-queue-listen-or-wait fixed
code did clamp decay to 0.01 when wait-function was present, but it did not set decay to 0.01 when decay was NIL. That caused that when timeout was not supplied we weren't  busy-waiting for the wait-function returning T.
- event-queue-read is fixed to not confuse decay with the scheduled-time (regression)

Notes:
check-schedule is called outside the lock to avoid recursive locking. It is a safe operation, because it may be invoked only from a thread of the queue consumer (CLIM loop). check-schedule may call queue-append which is protected by its own means.

I've pushed changes and now I'm going to test both simple-event-queue and concurrent-event-queue on SBCL and on ECL. I'll leave a comment about the results. (exploratory testing).

dkochmanski:
rebased against current master

dkochmanski:
I've tested it with ECL and indeed some problems followed. Said problems were a result of CLX safety 0 declaration though, so I'm satisfied with this branch. I've pushed one more commit which removes rudimentary class (we need to move some things to core, like distribute-event).

PR #789, opened 2019-06-28
Add missing contrasting dash patterns

### LINKED ISSUE #780: contrasting-dash-pattern-limit
`(contrasting-dash-pattern-limit (find-port))`
return 5 but the specification says (10.3.2) that it must return at least 8

scymtym:
Fixes #780.


scymtym:
This is what all combination of contrasting inks and dash patterns look like: screenshot

dkochmanski:
lgtm, thanks!

Issue #779, opened 2019-06-13
drei: rotate-yank does not remove first yank


dkochmanski:
Steps to reproduce:
1. open gadget test demo
2. C-k for `Editor` substring in text field
3. C-y to yank (Editor is re-appended)
4. M-y to rotate yank

- expected: first yank should be replaced
- observed: first yank is appended to the existing one, if there are more items then rotate yank replaces the second appended string with history

PR #790, opened 2019-07-01
Fix rotate-yank (#779).

kc-:
com-rotate-yank would only replace input when the previous command was a yank
rotation as well. Make it recognize com-yank as valid predecessor.


dkochmanski:
lgtm. thanks!

PR #795, opened 2019-07-12
Fix #744: Don't skip xlib:put-image for images larger than 2^11.

### LINKED ISSUE #744: Big images display incorrectly
Scaling images may break for some cases. Code to reproduce attached below. Original image and screenshots are attached further below.

Steps to reproduce:
1. compile code
2. run demo
3. check different scaling variants (0.1-2.1)

Observed: for some scaling wrong memory region is read
Expected: flawless scaling

Here my used code :
```common-lisp
(define-application-frame image-transform-demo ()
  ((image :initarg :image
          :accessor image-transform-demo/image))
  (:panes (image-demo :application
                      :display-function 'display-image-demo
                      :scroll-bars :both
                      :incremental-redisplay t)
          (scale-slider :slider
                        :value 0.9
                        :show-value-p nil
                        :orientation :horizontal
                        :min-value 0.1
                        :max-value 5
                        :value-changed-callback #'slider-updated-callback
                        :drag-callback #'slider-updated-callback))
  (:layouts (default (vertically ()
                       (19/20 (labelling (:label "Test Image"
					  :align-x :center
					  :label-alignment :top)
				image-demo))
                       (1/20 (labelling (:label "Scale")
			       scale-slider))))))

(defmethod initialize-instance :after ((obj image-transform-demo) &key)
  (setf (image-transform-demo/image obj)
        (make-pattern-from-bitmap-file #P "~/1.jpg")))

(defun slider-updated-callback (gadget value)
  (declare (ignore gadget value))
  (let ((frame *application-frame*))
    (redisplay-frame-pane frame (find-pane-named frame 'image-demo))))

(defun display-image-demo (frame stream)
  (updating-output (stream)
    (let* ((image (image-transform-demo/image frame))
           (scale (gadget-value (find-pane-named frame 'scale-slider)))
           (tr (make-scaling-transformation scale scale)))
      (with-drawing-options (stream :transformation tr)
        (draw-design stream image)))))

(defun image-transform-demo ()
  (let ((frame (make-application-frame 'image-transform-demo)))
(run-frame-top-level frame)))
```

Original image : 
image

Here running with scale is 1.1 
image

Here running with scale is 0.9 
image

kc-:
Previously, requests to draw regions with larger edge size than 2048 were simply
skipped, but still marked as updated, with the effect that uninitialized
contents of the graphics memory were visible instead.

Fix it by halving the size of pixmaps until they are small enough for
xlib:put-image to handle.

### DIFF
--- a/Backends/CLX/medium.lisp
+++ b/Backends/CLX/medium.lisp
@@ -28,6 +28,8 @@
 
 (defgeneric X-pixel (port color))
 
+(defconstant +x11-pixmap-dimension-limit+ 2048)
+
 (defmethod X-pixel ((port clx-basic-port) color)
   (let ((table (slot-value port 'color-table)))
     (or (gethash color table)
@@ -359,6 +361,21 @@ translated, so they begin at different position than [0,0])."))
 
 
 
+(defun put-image-recursively (pixmap pixmap-context pixmap-image width height x0 y0)
+  (labels ((put-partial-image (width height x0 y0)
+             (cond
+               ((and (< width +x11-pixmap-dimension-limit+) (< height +x11-pixmap-dimension-limit+))
+                (xlib:put-image pixmap pixmap-context pixmap-image
+                                :src-x x0 :src-y y0 :x x0 :y y0
+                                :width width :height height))
+               ((> width height)
+                (put-partial-image (ceiling width 2) height x0 y0)
+                (put-partial-image (floor width 2) height (+ x0 (ceiling width 2)) y0))
+               (T
+                (put-partial-image width (ceiling height 2) x0 y0)
+                (put-partial-image width (floor height 2) x0 (+ y0 (ceiling height 2)))))))
+    (put-partial-image width height x0 y0)))
+
 ;;; XXX: both PM and MM pixmaps should be freed with (xlib:free-pixmap pixmap)
 ;;; when not used. We do not do that right now.
 (defun compute-rgb-mask (drawable image)
@@ -384,9 +401,7 @@ translated, so they begin at different position than [0,0])."))
                   (if (< (ldb (byte 8 0) elt) #x80)
                       (setf (aref mdata y x) 0)
                       (setf (aref mdata y x) 1)))))
-      (unless (or (>= width 2048) (>= height 2048)) ;### CLX bug
-        (xlib:put-image mm mm-gc mm-image :src-x 0 :src-y 0 :x 0 :y 0
-                        :width width :height height :bitmap-p nil))
+      (put-image-recursively mm mm-gc mm-image width height 0 0)
       (xlib:free-gcontext mm-gc)
       (push #'(lambda () (xlib:free-pixmap mm)) ^cleanup)
       mm)))
@@ -412,9 +427,7 @@ translated, so they begin at different position than [0,0])."))
            (loop for y fixnum from 0 below height do
                 (let ((elt (aref idata y x)))
                   (setf (aref pdata y x) (ash elt -8)))))
-      (unless (or (>= width 2048) (>= height 2048)) ;### CLX bug
-        (xlib:put-image pm pm-gc pm-image :src-x 0 :src-y 0 :x 0 :y 0
-                        :width width :height height))
+      (put-image-recursively pm pm-gc pm-image width height 0 0)
       (xlib:free-gcontext pm-gc)
       (push #'(lambda () (xlib:free-pixmap pm)) ^cleanup)
       pm)))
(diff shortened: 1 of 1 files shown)


Backends/CLX/medium.lisp
```diff
@@ -384,9 +399,10 @@ translated, so they begin at different position than [0,0])."))
                   (if (< (ldb (byte 8 0) elt) #x80)
                       (setf (aref mdata y x) 0)
                       (setf (aref mdata y x) 1)))))
-      (unless (or (>= width 2048) (>= height 2048)) ;### CLX bug
-        (xlib:put-image mm mm-gc mm-image :src-x 0 :src-y 0 :x 0 :y 0
-                        :width width :height height :bitmap-p nil))
+      (if (or (>= width 2048) (>= height 2048)) ;### CLX bug
```
scymtym: Maybe write this using `(and (< width 2048) (< height 2048))` and flipping the legs? Having the same spelling of the condition makes it a bit easier to read.

Backends/CLX/medium.lisp
```diff
@@ -361,6 +361,21 @@ translated, so they begin at different position than [0,0])."))
 
 ;;; XXX: both PM and MM pixmaps should be freed with (xlib:free-pixmap pixmap)
 ;;; when not used. We do not do that right now.
+(defun put-image-recursively (pixmap pixmap-context pixmap-image width height x0 y0)
+  (cond
+    ((and (< width 2048) (< height 2048))
```
scymtym: Does clx export symbolic constants for these numbers? If yes, we should use those. If not, we should make our own (or add them in clx).

Backends/CLX/medium.lisp
```diff
@@ -361,6 +361,21 @@ translated, so they begin at different position than [0,0])."))
...
+     (xlib:put-image pixmap pixmap-context pixmap-image :src-x x0 :src-y y0 :x x0 :y y0
+                     :width width :height height))
+    ((> width height)
+     (put-image-recursively pixmap pixmap-context pixmap-image (ceiling width 2) height x0 y0)
```
scymtym: You could do the recursion using a local `labels` function to avoid passing the non-changing arguments over and over. That would also be more efficient in some implementations.

Backends/CLX/medium.lisp
```diff
@@ -412,9 +428,10 @@ translated, so they begin at different position than [0,0])."))
            (loop for y fixnum from 0 below height do
                 (let ((elt (aref idata y x)))
                   (setf (aref pdata y x) (ash elt -8)))))
-      (unless (or (>= width 2048) (>= height 2048)) ;### CLX bug
-        (xlib:put-image pm pm-gc pm-image :src-x 0 :src-y 0 :x 0 :y 0
-                        :width width :height height))
+      (if (or (>= width 2048) (>= height 2048)) ;### CLX bug
```
scymtym: Since this occurs twice: maybe call `put-image-recursively` unconditionally in both cases? The function call overhead shouldn't matter in this context.

Backends/CLX/medium.lisp
kc-: All right, but how large a pixmap are we expecting here? Seeing as this software rendering business will be obsolete in [optimistically short amount of time], I personally think optimizing for cycles is maybe best done at another place.

scymtym: True. I see the main benefit in not repeating the function name and unchanged arguments.

Backends/CLX/medium.lisp
kc-: As we discussed on IRC, the number two to the eleventh is known to CLX. Let me quote from Backends/CLX/keysyms.lisp:

```
;;; The parameter STATE is a bit mask represented as the logical OR
;;; of individual bits.  Each bit corresponds to a modifier or a
;;; pointer button that is active immediately before the key was
;;; pressed or released.  The bits have the following meaning:
;;;
;;;   position  value    meaning
;;;   [...]
;;;    11      2048      button4
```

As we know from the release notes to System 286 [1], All Number are documented:

```
[5] Number are documented.

    As part of our policy of keeping things well documented, all numbers
    in the system must be documented.  Creation of a number without any
    documentation will result in an error.  Use of the DESCRIBE function
    on a number will print out its documentation.  Note that numbers are
    EQ now, so don't try to use one number for more than one purpose.
    (That's why there's an infinite number of them.) (Well, almost, 
    as we only have a sex bit machine.)
```

[1]: http://www.ibiblio.org/pub/docs/humor/fionavar/sys286_msg

Backends/CLX/medium.lisp
kc-: Agreed, I put the change in.

dkochmanski:
this looks good, shall we merge it?

scymtym:
I think this is the best we can do assuming we don't know the root cause for the restriction.

PR #798, opened 2019-07-21
SETF FRAME-PRETTY-NAME notifies frame manager, changes window title

scymtym:


Backends/CLX/mirror.lisp
```diff
@@ -1,11 +1,29 @@
...
+  (setf (xlib:wm-icon-name window) icon-name)
+  (xlib:change-property window
+                        :_NET_WM_ICON_NAME
+                        (babel:string-to-octets icon-name :encoding :utf-8)
+                        :UTF8_STRING 8))
+
+(defmethod port-set-mirror-name ((port clx-basic-port) mirror name)
+  (%set-window-name mirror name)
+  (%set-window-icon-name mirror name))
+
```
dkochmanski: do we want to introduce sheet protocol for names?

i.e something in this spirit:

```
(defclass named-sheet-mixin ()
  ((sheet-name :initarg :name :accessor sheet-name)))

(defmethod (setf sheet-name) :after ((sheet mirrored-sheet) new-name)
  (setf (port-set-mirror-name (sheet-direct-mirror sheet))))

(defmethod (setf sheet-name) :after ((sheet named-sheet-mixin) new-name)
  (note-sheet-name-changed sheet new-name))
```

Backends/CLX/port.lisp
```diff
@@ -214,33 +214,23 @@
...
+         (name (frame-name frame))
+         (pretty-name (frame-pretty-name frame)))
+    (setf (xlib:wm-hints window) (xlib:make-wm-hints :input :on))
+    (%set-window-name window pretty-name)
+    (%set-window-icon-name window pretty-name)
+    (xlib:set-wm-class window (string-downcase name) (string-capitalize name))
+    (setf (xlib:wm-protocols window) `(:wm_delete_window))
+    (xlib:change-property window
+                          :WM_CLIENT_LEADER (list (xlib:window-id window))
+                          :WINDOW 32)))
```
dkochmanski: frame references could be removed from here if we had created `:after` method on `realize-mirror` in `panes.lisp` which calls the name setter. I'm not sure what to do with wm-class though in that case.

Backends/CLX/frame-manager.lisp
```diff
@@ -226,3 +226,9 @@
           #+nil
           (dispatch-repaint (sheet-parent sheet)
                             (region-union new-region old-region)))))))
+
+(defmethod clime:note-frame-pretty-name-changed
+    ((frame-manager clx-frame-manager) frame new-name)
+  (when-let* ((top-level-sheet (frame-top-level-sheet frame))
+              (mirror (sheet-direct-xmirror top-level-sheet)))
+    (port-set-mirror-name (port frame-manager) mirror new-name)))
```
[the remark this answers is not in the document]
scymtym: Changed `sheet-direct-xmirror` to `sheet-direct-mirror`.

The `top-level-sheet` slot in `standard-application-frame` is initialized to `nil`, so the `when-let*` seems necessary.

Backends/CLX/mirror.lisp
```diff
@@ -21,13 +39,13 @@
 (defmethod raise-mirror ((port clx-basic-port) (sheet basic-sheet))
   (let ((mirror (sheet-xmirror sheet)))
     (when (and mirror
-	       (typep mirror 'xlib:window))
+               (typep mirror 'xlib:window))
```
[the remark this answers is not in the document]
scymtym: That's an accidental whitespace change. I would rather not address that question in this pull request, if that's OK.

Backends/CLX/mirror.lisp
scymtym: I'm all for it. My main difficulty with this would be knowing where to put (which file, which package, etc.) the pieces.

Backends/CLX/port.lisp
scymtym: The sheet would need a name and a "pretty name" to make that possible. Are those in scope for the sheet protocol extension you suggested above?

Backends/CLX/mirror.lisp
dkochmanski: all three forms would belong in sheets.lisp after sheet-mirrored-mixin and before sheets ads bounding rectangle.

Backends/CLX/port.lisp
dkochmanski: can't it have only name? is it because pretty name is a string?

I imagine, that method on top-level sheet would setf this pane's name to the frame's pretty name, I was thinking about sheets name as a string designator. sheets are considered in clim abstraction as windows, I don't think there is a necessity to have two names for them.

scymtym: Having a string as a name for the sheet would makes, yes. But currently `%realize-mirror` also uses the `frame-name` (i.e. not pretty) for the window class. If it hadn't access to the frame, and the sheet also didn't have this second name, how would that work?

dkochmanski: I know that it works like that right now, but is there a reason to not set wm-class to the same thing?

scymtym: I'm certainly not an expert regarding ICCCM, but it sounds like the concepts are actually different.

The two `WM_CLASS` strings 
* identify the application instance and class (not e.g. a document or state within the window) for resource lookup and i presume automated window management
* are typically set at startup and not changed (in fact, as i understand it, they are not allowed to change while the window is visible)

The `_NET_WM_NAME` property
* should be displayed as the title of the window (other applications typically include things like current document title or status)
* can, as far as i can tell, change at any time (other applications do this. for example, gedit changes this property when a document is selected)

So merging the two doesn't seem right. In addition, the `frame-name`/`frame-pretty-name` pair seems to match the `WM_CLASS`/`_NET_WM_NAME` pair pretty well, given the description in the CLIM spec.

That said, I already agreed that having this work for sheets without relying on an associated frame would be good. My conclusion is that the sheet has to mirror the name/pretty name distinction. And why shouldn't it? For the IMGUI case, this separation would allow IMGUI windows to work with resources and window managers in the normal way while being able to control and change the window title independently.

scymtym:
I updated the draft according to what I think was our agreement.

Remaining problems from my point of view:

1. Should there be an explicit generic function definition for `climb:note-frame-pretty-name-changed'? If so, where should it go?

2. Which of the new generic functions need documentation strings? Is any other documentation required?

3. Setting the pretty name of the frame changes the pretty name of the top-level sheet. But setting the pretty name of the top-level sheet directly does not change the pretty name of the frame. As a result, the visible window title and the pretty name of the frame can get out of sync.

One way to test the changes is playing with the following in the listener:

``` common-lisp
(loop :for i :to 10
      :do (setf (frame-pretty-name *application-frame*) (format nil "foo ~D" i))
          (sleep 1))

(loop :for i :to 10
      :do (setf (clime:sheet-pretty-name (frame-top-level-sheet *application-frame*))
                (format nil "foo ~D" i))
          (sleep 1))

(frame-pretty-name *application-frame*)
```

Backends/CLX/port.lisp
dkochmanski: sounds reasonable

Backends/CLX/port.lisp
```diff
@@ -214,33 +214,24 @@
...
-      (xlib:change-property window
-                            :WM_CLIENT_LEADER (list (xlib:window-id window))
-                            :WINDOW 32))))
+  (let* ((q (compose-space sheet))
+         (window (realize-mirror-aux
+                  port sheet
+                  :map nil
+                  :width (round-coordinate (space-requirement-width q))
+                  :height (round-coordinate (space-requirement-height q))))
+         (name (pane-name sheet))
```
dkochmanski: name should be sheet's property. panes are "sheets belonging to frames". In this case pane-name should be a method which resolves to sheet-name (to make this name consistent), so in clim-core/panes.lisp protocol class pane should not change

(name :initarg :name :initform "(Unnamed Pane)" :reader pane-name)

but we could add a comment that this slot name is the same as in named-sheet-mixin, so in top-level-sheet-pane both methods would refer to it.

Core/clim-basic/sheets.lisp
```diff
@@ -773,6 +776,11 @@ might be different from the sheet's native region."
               (sheet-transformation sheet)))))
     native-transformation))
 
+;;; Named sheets
+
+(defclass named-sheet-mixin ()
+  ((%pretty-name :initarg :pretty-name :accessor clime:sheet-pretty-name)))
```
dkochmanski: this should also incorporate name (see the previous remark)

Core/clim-core/frames.lisp
```diff
@@ -244,6 +244,10 @@ documentation produced by presentations.")
       (setf (slot-value frame 'layouts) nil))
     (setf (%frame-manager frame) fm)))
 
+(defmethod (setf frame-pretty-name) :after (new-value frame)
+  (when-let ((frame-manager (frame-manager frame)))
+    (clime:note-frame-pretty-name-changed frame-manager frame new-value)))
```
dkochmanski: as per some convention I don't remember where I saw it (probably in CLIM II spec) note-* methods are reserved for api consumers (i.e application programmers), and clim should do its things inside the methods which perform the actual thing. In this spirit we shout set sheet's pretty name here, not in note-frame-pretty-name-changed.

Core/clim-core/frames.lisp
```diff
@@ -683,6 +688,11 @@ documentation produced by presentations.")
 (defmethod bury-frame ((frame application-frame))
   (bury-sheet (frame-top-level-sheet frame)))
 
+(defmethod note-frame-pretty-name-changed
+    ((frame-manager frame-manager) frame new-name)
+  (when-let ((top-level-sheet (frame-top-level-sheet frame)))
+    (setf (sheet-pretty-name top-level-sheet) new-name)))
```
dkochmanski: ditto, note methods should not perform anything (only have default methods doing nothing to prevent errors)

package.lisp
```diff
@@ -1881,6 +1881,7 @@
    #:event-listen-or-wait
    #:schedule-event
    ;; sheets
+   #:sheet-pretty-name
```
dkochmanski: ditto, add sheet-name here

dkochmanski:
> I updated the draft according to what I think was our agreement.
> 
> Remaining problems from my point of view:
> 
>     1. Should there be an explicit generic function definition for `climb:note-frame-pretty-name-changed'? If so, where should it go?

there shouldn't. also note, that note-frame-pretty-name-changed is already present in climi by importing from clime (which also imports from climb), so it would be the same symbol. Adding this symbol to climb exports would only make it visible as external in this package. It is a conscious decision, all "core" components should have no name conflicts hence they share same climi package.
> 
>     2. Which of the new generic functions need documentation strings? Is any other documentation required?

I think we are fine as we are. If you feel otherwise add a node in extensions.txi "Frame and sheet names" which describes shortly the protocol and note method.
> 
>     3. Setting the pretty name of the frame changes the pretty name of the top-level sheet. But setting the pretty name of the top-level sheet directly does not change the pretty name of the frame. As a result, the visible window title and the pretty name of the frame can get out of sync.

IMO this is expected. If someone violates the abstraction boundaries they must understand the consequences. CLIM programmer doesn't have any obvious access to top-level-sheet-pane.
> 
> 
> One way to test the changes is playing with the following in the listener:
> 
> ```lisp
> (loop :for i :to 10
>       :do (setf (frame-pretty-name *application-frame*) (format nil "foo ~D" i))
>           (sleep 1))
> 
> (loop :for i :to 10
>       :do (setf (clime:sheet-pretty-name (frame-top-level-sheet *application-frame*))
>                 (format nil "foo ~D" i))
>           (sleep 1))
> 
> (frame-pretty-name *application-frame*)
> ```

Core/clim-basic/decls.lisp
```diff
@@ -196,6 +196,22 @@
 (defgeneric sheet-occluding-sheets (sheet child))
 (defgeneric map-over-sheets (function sheet))
 
+(defgeneric clim-extensions:sheet-name (sheet)
+  (:documentation "McCLIM extension: Return the name of SHEET.
+The returned name is a symbol and does not change.
+Only top-level sheets have a name. For sheets which are also panes, the
```
dkochmanski: "Only top-level sheets have a name" - top-level-sheet-pane is not an exported symbol, so it doesn't make sense to mention it here.  Maybe there should be a default method resolving to `unnamed-sheet` (name) and `"(Unnamed Sheet)"`? Also we may want to export named-sheet-mixin in clime.

Other than that, I think we should keep package prefix out of the definitions. All definitions reside in climi package, so there should be no conflicts and that way we'll be able to just skip the symbol from clime exports to change its home package.

Documentation/Manual/Texinfo/extensions.texi
```diff
@@ -2,13 +2,30 @@
...
+@node Frame and sheet names
+@section Frame and sheet names
+
+According to the @clim specification, a frame has a name and a pretty
+name the latter of which can be changed. @mcclim extends this design to
+top-level sheets in form of the following protocol:
+
+@include docstrings/fun-clim-extensions-sheet-name.texi
+@include docstrings/fun-clim-extensions-sheet-pretty-name.texi
+@include docstrings/fun-clim-extensions-setf-sheet-pretty-name.texi
```
dkochmanski: mixin who is the protocol owner is not mentioned here (but we've exported it).

Core/clim-basic/decls.lisp
```diff
@@ -196,6 +196,24 @@
...
+The returned name is a symbol and does not change.
+For sheets which are also panes, the returned name is identical to the
+pane name.")
+  (:method (sheet) nil))
+(defgeneric sheet-pretty-name (sheet)
+  (:documentation "McCLIM extension: Return the pretty name of SHEET.
+The returned name is a string and may change over time.
+The pretty name usually corresponds to the title of the associated
+window.")
+  (:method (sheet) "(Unnamed sheet)"))
```
dkochmanski: you may ignore this remark, but I wonder if default method like this wouldn't be better:
```
(if-let ((name (sheet-name sheet)))
  (format nil "~a" name)
  "(Unnamed sheet)")
```

dkochmanski:
in the future (if there is a factual need for it) we may introduce note-sheet-pretty-name-changed. For now there is no client who would need this. another potential thing is creating a small demo illustrating the functionality in clim-demo package.

PR #800, opened 2019-07-25
Specify and regularize default dimensions for panes. See #699

kc-:
Caveat: there is some history to this type of change. For example, a468fc6
removed some "magic numbers", then 9dedf6b reintroduced them again a few days
later. It is advisable to look through the commit history before working
on this.

The basic problem is that panes without contents might result in computed space
requirements of 0x0 pixels, which aren't sensible (and the X protocol can't
handle). For these situations, some code paths had default values for width and
height, others did not. This commit normalizes those defaults to 100x100 pixels.

### DIFF
--- a/Core/clim-core/panes.lisp
+++ b/Core/clim-core/panes.lisp
@@ -384,9 +384,8 @@ order to produce a double-click")
 (defmethod (setf medium-background) (ink (pane pane))
   (setf (medium-background (sheet-medium pane)) ink))
 
-(defmethod compose-space ((pane pane) &key width height)
-  (make-space-requirement :width (or width 200)
-			  :height (or height 200)))
+(defmethod compose-space ((pane pane) &key (width 100) (height 100))
+  (make-space-requirement :width width :height height))
 
 (defmethod allocate-space ((pane pane) width height)
   (declare (ignorable pane width height))
@@ -1808,9 +1807,8 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
   (dolist (child (alexandria:ensure-list contents))
     (sheet-adopt-child sheet child)))
 
-(defmethod compose-space ((bboard bboard-pane) &key width height)
-  (declare (ignore width height))
-  (make-space-requirement :width 300 :height 300))
+(defmethod compose-space ((bboard bboard-pane) &key (width 100) (height 100))
+  (make-space-requirement :width width :height height))
 
 ;;; VIEWPORT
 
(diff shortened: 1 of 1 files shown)


Core/clim-core/panes.lisp
```diff
@@ -1810,7 +1809,7 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
 
 (defmethod compose-space ((bboard bboard-pane) &key width height)
   (declare (ignore width height))
-  (make-space-requirement :width 300 :height 300))
+  (make-space-requirement :width 100 :height 100))
```
dkochmanski: for consistency make it obey specified width/height with default values 100/100.

Core/clim-core/panes.lisp
```diff
@@ -2536,8 +2535,8 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
    (user-min-height :accessor %pane-user-min-height)
    (user-max-height :accessor %pane-user-max-height)
    ;; size required by the stream
-   (stream-width :initform 0 :accessor stream-width)
-   (stream-height :initform 0 :accessor stream-height))
+   (stream-width :initform 100 :accessor stream-width)
+   (stream-height :initform 100 :accessor stream-height))
```
dkochmanski: this is good, be we need to fix also 

- `window-clear-method`
 (sets width/height to 0), 

there are other places which change stream width/height size, but imo we should leave them as they are.

kc-: I looked at that, but was under the (mistaken?) impression that that method was part of the deconstruction of panes. So I assumed it ran before without problems. 
What do we do there? Setting width/height to the 100^2 default is probably a bad idea.

Core/clim-core/panes.lisp
kc-: Oh, of course. Thanks.

Core/clim-core/panes.lisp
dkochmanski: I've confused myself. Default sizes require more thought and (maybe) minimal values should be supplied by backends, but window-clear won't be a problem, because frame is already created and compose-space protocol will take care of it.

Core/clim-core/panes.lisp
```diff
@@ -1808,9 +1807,9 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
   (dolist (child (alexandria:ensure-list contents))
     (sheet-adopt-child sheet child)))
 
-(defmethod compose-space ((bboard bboard-pane) &key width height)
+(defmethod compose-space ((bboard bboard-pane) &key (width 100) (height 100))
   (declare (ignore width height))
```
dkochmanski: width and height are not ignored anymore, please remove the declaration.

kc-: (Done, sorry about that.)

dkochmanski:
lgtm, thanks!

PR #806, opened 2019-07-27
Another round of documentation improvements

scymtym:
* Fix ex2 screenshot in manual

* Improve guided tour example programs. **Important**: I'm not sure whether the need to set `medium-background` as well as `pane-background` in `color-editor.lisp` is a bug or not. If it is a bug, it should be fixed instead of working around it in the example.

* Fix generation and deletion of listing snippets in guided tour `Makefile`


nsrahmad:
>Important: I'm not sure whether the need to set medium-background as well as pane-background in color-editor.lisp is a bug or not. If it is a bug, it should be fixed instead of working around it in the example.

>* Set pane and medium background to avoid repainting different
    regions with  different inks

I fail to spot the difference between setting both and just `medium-background`. setting `medium-background` should be enough.

Documentation/Guided-Tour/file-browser.lisp
```diff
@@ -1,25 +1,21 @@
-(eval-when (:compile-toplevel)
-  (asdf:oos 'asdf:load-op :clim)
-  (asdf:oos 'asdf:load-op :clim-clx))
+(cl:eval-when (:compile-toplevel :load-toplevel :execute)
+  (cl:map cl:nil #'asdf:load-system '(:mcclim :cl-fad)))
```
nsrahmad: I thing using `ql:quickload` will be better everywhere. It will be simpler, much less chance of failing, and wouldn't output tons of warnings.

Documentation/Manual/LaTeX/ex2-screenshot.png
```diff
@@ -0,0 +1 @@
+../Texinfo/figures/ex2-screenshot.png
```
nsrahmad: You will need to fix the LaTeX build script too.

dkochmanski:
I think we should delete latex version in this pr and use only texinfo as toplevel manual. This causes plenty of confusion among people that we "have" two manuals.

Dnia 29 lipca 2019 15:07:51 CEST, Nisar Ahmad <notifications@github.com> napisał(a):
>nsrahmad commented on this pull request.
>
>
>
>> @@ -0,0 +1 @@
>+../Texinfo/figures/ex2-screenshot.png
>
>You will need to fix the LaTeX build script too.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/McCLIM/McCLIM/pull/806#pullrequestreview-267768888

-- Wysłane za pomocą K-9 Mail.

scymtym:
> I fail to spot the difference between setting both and just medium-background. setting medium-background should be enough.

The difference only comes into play in a specific scenario:
1. Open the application (this sets medium and pane backgrounds based on the `:background` initarg)
2. Change the color (this only sets the medium background)
3. Make the frame bigger

The third step paints the freshly revealed part of the pane using the pane background which still is the ink initially supplied as `:background` and thus different from the current medium background.

Documentation/Guided-Tour/file-browser.lisp
scymtym: I can do that.

scymtym:
> I think we should delete latex version in this pr and use only texinfo as toplevel manual. This causes plenty of confusion among people that we "have" two manuals.

While I agree that we should ultimately do this, I think it is too early for that. I regularly check LaTeX version for comparison.

scymtym:
I addressed the remarks (using `ql:quickload`, why set pane background?, fix LaTeX build).

I added the following changes:

* Fix `@clim` and `@mcclim` macro again

* Use `@document{encoding,language}` correctly (The previous spelling of the UTF-8 encoding worked for HTML output, but not LaTeX)

* Improve `@glossentry` macro and use it

* Put all non-concept entries into a single index; attach entries to sensible letters (instead of the first letter of the package name)

Documentation/Manual/LaTeX/chap-first-application.tex
```diff
@@ -294,7 +294,7 @@ \section{Adding Functionality}
 
 \begin{figure}
 \begin{center}
-\inputfig{ex2.pdf_t}
+\includegraphics{ex2-screenshot.png}
```
nsrahmad: Image is much smaller than expected. Try `\includegraphics[width=.6\linewidth]{ex2-screenshot.png}`.

scymtym: Fixed.

nsrahmad:
Looks good now, Thanks.

PR #808, opened 2019-08-01
Add declarations for variables that have not been defined yet.

kc-:
Fixes these compilation warnings:
```
; file: [...]/McCLIM/Core/clim-core/dialog-views.lisp
[...]
; caught WARNING:
;   undefined variable: CLIM-INTERNALS::*ACCEPTING-VALUES-STREAM*
```

```
; file: [...]/McCLIM/Libraries/Drei/lisp-syntax.lisp
[...]
; caught WARNING:
;   undefined variable: DREI-LISP-SYNTAX::|initial-state |
```

### DIFF
--- a/Core/clim-core/dialog-views.lisp
+++ b/Core/clim-core/dialog-views.lisp
@@ -319,50 +319,33 @@ COMPLETION presentation type as a pop-up menu."))
   (unless default-supplied-p
     (setq default (funcall value-key (elt sequence 0))))
   (let ((record (updating-output (stream :unique-id query-identifier
-				  :cache-value default
-				  :record-type 'av-pop-up-menu-record)
-		  (with-output-as-presentation
-		      (stream query-identifier 'selectable-query)
-		    (surrounding-output-with-border
-			(stream :shape :inset :move-cursor t)
-		      (write-string (funcall name-key default) stream))))))
+                                  :cache-value default
+                                  :record-type 'av-pop-up-menu-record)
+                  (with-output-as-presentation
+                      (stream query-identifier 'selectable-query)
+                    (surrounding-output-with-border
+                        (stream :shape :inset :move-cursor t)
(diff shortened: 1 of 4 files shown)


Core/clim-core/dialog-views.lisp
```diff
@@ -172,6 +172,7 @@
 (define-presentation-method accept-present-default
     ((type completion) stream (view option-pane-view) default default-supplied-p
      present-p query-identifier)
+  (declare (special *accepting-values-stream*))
```
lokedhs: Shouldn't this be a globally declared `DEFVAR` instead?

Libraries/Drei/lisp-syntax.lisp
```diff
@@ -50,6 +50,8 @@
 ;;;
 ;;; the syntax object
 
+(declaim (special |initial-state |))
```
lokedhs: I honestly am not sure how declaim with special works, but I'd assume that results in a global special declaration, similar to how `DEFVAR` works?

kc-: Yes, that's what I think as well.

Core/clim-core/dialog-views.lisp
kc-: Semantically, yes. Practically, does it make a difference?

dkochmanski: yes, people usually look for where the variable is defined, if it is only declared then it may be confusing..

Libraries/Drei/lisp-syntax.lisp
dkochmanski: please make it defvar too, while it may be that they are the same it will be less surprising to read defvar and there would no such discussion in the first place (what kind of proves the point)

dkochmanski:
merging, thanks!

PR #813, opened 2019-08-07
Fix cross-references in listings for non-HTML output in manual

scymtym:


dkochmanski:
works for me. thanks!

Issue #585, opened 2018-10-30
streams: make stream access thread-safe


dkochmanski:
Right now when we write asynchronously to a stream output records get corrupted. Issue is reported here:

https://github.com/McCLIM/McCLIM/issues/315

This is opened as a separate task, because it is more an architecture enhancement. I plan to assign $600 bounty on this task before the next release, but if fixed earlier it will be paid anyway.

dkochmanski:
https://www.bountysource.com/issues/65501924-streams-make-seos-access-thread-safe

g--o:
> https://www.bountysource.com/issues/65501924-streams-make-seos-access-thread-safe

Hey is this still relevant?

dkochmanski:
yes it is. there is even a preferred way how to fix it (unless you have another proposal which we could discuss): check if the stream thread is the same from which operation is invoked and if not queue an event with the operation (and define handle-event method which works on such events). I've tried first with locks, but that didn't work for various reasons (and had a non-negligible performance penalty).

g--o:
@dkochmanski 
Hey, so I've been investigating for a while and with the documentation as bad as it is, it's pretty hard to figure out the exact source of the race condition and the flow of your architecture. I've seen your commits and the locking seem to be too general, and thus create deadlock in some of my tests. Note that even with a queue you'd need a lock and to know when exactly to dequeue, so understanding the race condition is crucial for solving the problem anyway.

As of my own findings - while i suspect as well that something's going on with recordings and repaints it definitely all starts with the the stream invoking scrolling methods like move-pane (according to my tests). Still, any pointers would be really great as to what your investigation led up to so far, so I can maybe try and continue from there to speed up the process. In particular: is the stream we interact with and the record stream the same? because this might be problematic as well

I'd really appreciate your thoughts on this, thanks!

dkochmanski:
Event queue already is thread-safe (see file input.lisp). Output stream may record its output (when it is an output-recording-stream), so it is indeed a case that the stream we interact with is recording its output. That's basically when the corruption happens if we call it asynchronously. Here is where the race condition may come from:

1. CLIM loop receives a repaint event and sets recording to NIL on the stream
2. We call draw-rectangle* on said stream what sets recording to T
3. repaint carries on and repaint is recorded (changing i.e current-output-record)
4. everything is jumbled up from this point on

[...]

Issue #315, opened 2017-08-30
Corrupted output to application frame


john-a-carroll:
In my application I have a window that displays diagnostic messages and a log of the progress of long computations. In Franz Allegro CLIM, a scrolling application pane works well.

But I have had only limited success with this in McCLIM. In mcclim-20160929 (with clx-20160929), new output was sometimes initially invisible and partially corrupted, but it was just about acceptable. In mcclim-20170725 (with clx-20170630) the situation is much worse. Output that arrives quickly gets corrupted and soon the window stops displaying any new output.

Here is an example that shows this behaviour:

```
(define-application-frame scrolling-frame ()
  ()
  (:pane
    (scrolling ()
      (make-pane 'application-pane
	:name 'display
        :end-of-page-action :scroll)))) ; same problem also with :allow

(defvar *scrolling-frame* nil)

(defun run-scrolling-frame ()
  (setq *scrolling-frame* (make-application-frame 'scrolling-frame))
  (run-frame-top-level *scrolling-frame*))

(sb-thread:make-thread #'run-scrolling-frame)

```
Run the following a few times - the output starts to get corrupted and soon becomes invisible.

```
(let ((*standard-output* (find-pane-named *scrolling-frame* 'display)))
  (dotimes (n 30)
    (format t "~A~%" n) (force-output) (sleep 0.01)))

```
Now open a new scrolling-frame and run the following a few times - the output is mostly OK, but with occasional glitches.

```
(let ((*standard-output* (find-pane-named *scrolling-frame* 'display)))
  (dotimes (n 30)
    (format t "~A~%" n) (force-output) (sleep 0.1)))

```
The fact that the behaviour changes with the speed of arrival of new output suggests to me that there is some sort of low level buffering or timing dependency. But I have no good ideas about how to find the problem.

dkochmanski:
this is a similar issue to:
https://github.com/robert-strandh/McCLIM/issues/170

from my investigation of the issue, output-records get corrupted due to race condition on streams.

clynbech:
If you are having problems reproducing the issue, try lowering the sleep value to something like 0.001.

PR #821, opened 2019-08-23
[wip] seos: now queueing operations

g--o:
#585

### DIFF
--- a/Core/clim-basic/events.lisp
+++ b/Core/clim-basic/events.lisp
@@ -5,6 +5,8 @@
 ;;;           Iban Hatchondo (hatchond@emi.u-bordeaux.fr)
 ;;;           Julien Boninfante (boninfan@emi.u-bordeaux.fr)
 ;;;           Robert Strandh (strandh@labri.u-bordeaux.fr)
+;;;  (c) copyright 2019 by
+;;;           Nir B. (go.prodml@gmail.com)
 
 ;;; This library is free software; you can redistribute it and/or
 ;;; modify it under the terms of the GNU Library General Public
@@ -96,6 +98,9 @@
 	   ,@options)
 	 (defmethod event-type ((event ,name))
 	   ',type)))))
+
+(define-event-class lambda-event (standard-event)
+  ((fun :initarg :fun :reader fun)))
 
 (define-event-class device-event (standard-event)
   ((modifier-state :initarg :modifier-state
--- a/Core/clim-basic/output.lisp
+++ b/Core/clim-basic/output.lisp
@@ -2,6 +2,8 @@
 
 ;;;  (c) copyright 1998,1999,2000 by Michael McDonald (mikemac@mikemac.com)
 ;;;  (c) copyright 2014 by Robert Strandh (robert.strandh@gmail.com)
+;;;  (c) copyright 2019 by
+;;;           Nir B. (go.prodml@gmail.com)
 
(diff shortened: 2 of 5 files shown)


dkochmanski:
it doesn't seem to be a complete pull request neither it does follow the approach I've proposed (to queue operations only when they are not executed in clim loop).

regarding copyright headers: changes are too small to add your name to file copyright. Also full names are to be put there (not initials), if you put initials who is the copyright owner?

g--o:
@dkochmanski it doesn't involve the approach you've proposed becaue that approach does not work.
The problem is with draw-text and move-sheet, or more generally rendering and scrolling, so the operations of the stream should be queued on a high level where the logic is clear.

As of the copyright notice I just followed the LGPL template as copying the code should involve getting in contact with whoever made anything there. once the diff becomes empty, names can be removed. How to identify the owner? simply corresponds to the email owner.

dkochmanski:
Saying that it "didn't work" is too vague, could you elaborate why it didn't work? Also there is no need to queue operations if they are invoked from the same thread, it is a waste performance-wise.

Speaking of copyrights, it is a good practice to add them only when significant changes are made, otherwise we'd have longer headers than the source code content. If you implement safe variants for draw-* in graphics that will be surely enough to put the name as a copyright notice.

g--o:
@dkochmanski Oh you mean checking if the stream is running in clim's thread? I'll try that
I honestly didn't agree with your analysis to begin with as the problem didn't exactly seem to stem from draw-rectangle. The draw-rectangle seemed to be invoked from the scrollbar (which is a "side-effect"), meaning: keeping the draw-rectangle but not scrolling works. Trying to queue each and every side effect separately makes little sense when the logical error is trying to render while you scroll.

As of the copyright stuff, look at the end of the day it's your repo (at least partially as a maintainer), i'm just trying to explain that this good practice makes no sense because while harder to maintain the list (e.g: removing names), it makes no clean cut and also asks to give up rights, which i get isn't impressive, but i'm supporting this side for correctness sake.

dkochmanski:
I think that you misunderstand the issue #585. Failure case shown in a #315 is just an example of things going awry; the ticket is about making all asynchronous drawing-related requests (like drawing a circle, a rectangle, printing a text) to the output stream work correctly. While it won't buy much for panes with a display time synchronized with a command loop it will be very useful for panes which are not redisplayed unless explicitly asked. I'm not insisting on a solution with queuing, if you have something better i'm all for it, but please remember that fixing #315 is not this task objective.

g--o:
@dkochmanski huh? the title is "make seos access thread-safe"
the description is 
"Right now when we write asynchronously to a stream output records get corrupted. Issue is reported here:#315"

Are you joking?

dkochmanski:
The problem is caused by output record corruption and issue #315 exhibits the problem. But the issue is deeper than just drawing text. I thought I've explained it in this comment (which was prompted by your question):
https://github.com/McCLIM/McCLIM/issues/585#issuecomment-522527288
As you can probably guess I am not joking.

g--o:
The comment you quoted literally explains nothing. You wanted seos to be thread safe. it is thread safe, plus i checked your hypothesis relevancy. I seriously don't see the problem

dkochmanski:
I won't accept this pull request as a solution for #585. I'm sorry if a lacking description caused you to spend time on something else.

g--o:
wow this is sad, im out

PR #822, opened 2019-08-23
improve event handling

dkochmanski:
- introduce uniform pointer grabbing handling
- introduce uniform pointer pressed handling
- make input focus work for any sheet
- minor fixes for utilities and ux

### DIFF
--- a/Backends/CLX/input.lisp
+++ b/Backends/CLX/input.lisp
@@ -136,7 +136,7 @@
                         :x x :y y
                         :graft-x root-x
                         :graft-y root-y
-                        :sheet (or (frame-properties (pane-frame sheet) 'focus) sheet)
+                        :sheet sheet
                         :modifier-state modifier-state :timestamp time)))
       ((:button-press :button-release)
        (let ((modifier-state (clim-xcommon:x-event-state-modifiers *clx-port* state))
@@ -404,11 +404,6 @@
 	       ;; The event initialization code will give us a
 	       ;; reasonable timestamp.
 	       :timestamp 0))))))))
-
-(defmethod port-frame-keyboard-input-focus ((port clx-basic-port) frame)
-  (frame-properties frame 'focus))
-(defmethod (setf port-frame-keyboard-input-focus) (focus (port clx-basic-port) frame)
-  (setf (frame-properties frame 'focus) focus))
 
 (defmethod port-grab-pointer ((port clx-basic-port) pointer sheet)
   (let ((mirror (sheet-xmirror sheet))
--- a/Core/clim-basic/ports.lisp
+++ b/Core/clim-basic/ports.lisp
@@ -74,33 +74,43 @@
     :accessor port-lock)
    (text-style-mappings :initform (make-hash-table :test #'eq)
                         :reader port-text-style-mappings)
+   (focused-sheet :initform nil :accessor port-keyboard-input-focus
+		  :documentation "The sheet the keyboard events are passed to.")
    (pointer-sheet :initform nil :accessor port-pointer-sheet
 		  :documentation "The sheet the pointer is over, if any")
+   ;; The difference between grabbed-sheet and pressed-sheet is that
+   ;; the former takes all pointer events while pressed-sheet receives
+   ;; replicated pointer motion events. -- jd 2019-08-21
    (grabbed-sheet :initform nil :accessor port-grabbed-sheet
-		  :documentation "The sheet the pointer is grabbing, if any")))
+		  :documentation "The sheet the pointer is grabbing, if any")
+   (pressed-sheet :initform nil :accessor port-pressed-sheet
+		  :documentation "The sheet the pointer is pressed on, if any")))
 
-(defmethod port-keyboard-input-focus (port)
-  (when (null *application-frame*)
-    (error "~S called with null ~S" 
-           'port-keyboard-input-focus '*application-frame*))
-  (port-frame-keyboard-input-focus port *application-frame*))
+(defgeneric note-input-focus-changed (sheet state)
+  (:documentation "~
+Called when a sheet receives or loses the keyboard input focus. State
+is T when focused and NIL otherwise. This is a McCLIM extension.")
+  (:method (sheet state)
+    (declare (ignore sheet state))))
 
-(defmethod (setf port-keyboard-input-focus) (focus port)
-  (when (null *application-frame*)
-    (error "~S called with null ~S" 
-           '(setf port-keyboard-input-focus) '*application-frame*))
-  ;; XXX: pane frame is not defined for all streams (for instance not for
-  ;; CLIM:STANDARD-EXTENDED-INPUT-STREAM), so this sanity check would lead to
-  ;; error on that.
-  ;; XXX: also should we allow reading objects from foreign application frames?
-  ;; This was the case on Genera and is requested by users from time to time...
-  #+ (or)
-  (unless (eq *application-frame* (pane-frame focus))
-    (error "frame mismatch in ~S" '(setf port-keyboard-input-focus)))
-  (setf (port-frame-keyboard-input-focus port *application-frame*) focus))
+(defmethod (setf port-keyboard-input-focus) :around (sheet port)
+  (let ((old-sheet (port-keyboard-input-focus port))
+        ;; Result may not reflect the actual focus.
+        (result (call-next-method))
+        (new-sheet (port-keyboard-input-focus port)))
+    (prog1 result
+      (unless (eq old-sheet new-sheet)
+        (when-let ((top-sheet (get-top-level-sheet new-sheet)))
+          (setf (focused-sheet top-sheet) new-sheet))
+        (note-input-focus-changed new-sheet t)
+        (note-input-focus-changed old-sheet nil)))))
 
-(defgeneric port-frame-keyboard-input-focus (port frame))
-(defgeneric (setf port-frame-keyboard-input-focus) (focus port frame))
+(defmethod (setf port-keyboard-input-focus) ((sheet top-level-sheet-mixin) port)
+  (if-let ((focus (focused-sheet sheet)))
+    (if (eq focus sheet)
+        (call-next-method)
+        (setf (port-keyboard-input-focus port) focus))
+    (call-next-method)))
 
 (defun find-port (&key (server-path *default-server-path*))
   (if (null server-path)
@@ -110,18 +120,18 @@
   (setq server-path
 	(funcall (get (first server-path) :server-path-parser) server-path))
   (loop for port in *all-ports*
-      if (equal server-path (port-server-path port))
-      do (return port)
-      finally (let ((port-type (get (first server-path) :port-type))
-		    port)
-		(if (null port-type)
-		    (error "Don't know how to make a port of type ~S"
-			   server-path))
-		(setq port
-		      (funcall 'make-instance port-type
-			       :server-path server-path))
-		(push port *all-ports*)
-		(return port))))
+     if (equal server-path (port-server-path port))
+     do (return port)
+     finally (let ((port-type (get (first server-path) :port-type))
+		   port)
+	       (if (null port-type)
+		   (error "Don't know how to make a port of type ~S"
+			  server-path))
+	       (setq port
+		     (funcall 'make-instance port-type
+			      :server-path server-path))
+	       (push port *all-ports*)
+	       (return port))))
 
 (defmethod destroy-port :before ((port basic-port))
   (when (and *multiprocessing-p* (port-event-process port))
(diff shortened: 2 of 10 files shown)


Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+         (setf sheet child))))
+
+;;; Function is responsible for making a copy of an immutable event
+;;; and adjusting its coordinates to be in the target-sheet
+;;; coordinates. Optionally it may change event's class.
+(defun dispatch-event-copy (target-sheet event &optional new-class
+                            &aux (sheet (event-sheet event)))
+  (if (eql target-sheet sheet)
+      (dispatch-event sheet event)
+      (let ((new-event (shallow-copy-object event)))
```
scymtym: Can you make `shallow-copy-object` optionally accept a new class and initargs to make this more straightforward?

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+      (dolist (s (do ((s new-pointer-sheet (sheet-parent s))
+                      (lis nil))
+                     ((or (null s) (graftp s) (eq s common-ancestor)) lis)
+                   (push s lis)))
+        (dispatch-event-copy s event 'pointer-enter-event)))))
+
+(defmethod distribute-event ((port basic-port) event)
+  (dispatch-event (event-sheet event) event))
+
+;;; In the most general case we can't tell whenever all sheets are
```
scymtym: "can't tell whether all"

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+;;; mirrored or not. So a default method for pointer-events operates
```
scymtym: "So this default method ..."

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+;;; under the assumption that we must deliver events to sheets which
```
scymtym: "to sheets which don't" or "to the sheet which doesn't"

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+;;; doesn't have a mirror and that the sheet grabbing and pressing is
```
scymtym: Maybe "that tracking of the grabbed sheet and pressed sheet is not ..."?

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+        (new-pointer-sheet (get-pointer-event-sheet event)))
+    (when grabbed-sheet
+      (unless (typep event '(or pointer-enter-event pointer-exit-event))
+        (dispatch-event-copy grabbed-sheet event))
+      (return-from distribute-event))
+    ;; Synthetize boundary change events and update the port.
+    (synthetize-enter-exit-events old-pointer-sheet new-pointer-sheet event)
+    (setf (port-pointer-sheet port) new-pointer-sheet)
+    ;; Set the pointer cursor.
+    (when-let ((cursor-sheet (or pressed-sheet new-pointer-sheet)))
```
scymtym: Shouldn't `cursor-sheet` be used at some point?

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+
+(defmethod distribute-event ((port basic-port) event)
+  (dispatch-event (event-sheet event) event))
+
+;;; In the most general case we can't tell whenever all sheets are
+;;; mirrored or not. So a default method for pointer-events operates
+;;; under the assumption that we must deliver events to sheets which
+;;; doesn't have a mirror and that the sheet grabbing and pressing is
+;;; not implemented by the port (we emulate it). -- jd 2019-08-21
 (defmethod distribute-event ((port basic-port) (event pointer-event))
```
scymtym: The commit message says

> [...] will behave slightly different, namely when pointer will reach different mirror our "pressed sheet" will lose focus and it won't receive events.

If I understand the new intended behavior correctly, I'm not sure it is good. Consider a frame with a scroll-bar on its right border and a larger frame "below" it (region of the first sheet is inside the region of the second sheet in graft coordinates). If I press and hold the pointer button to move the scroll-bar, I can let the pointer wander off the scroll-bar to the left (further inside the top frame), but if the pointer moves to the right onto the bottom frame (button still pressed), the scroll-bar will stop responding to pointer motion.

Core/clim-core/gadgets.lisp
```diff
@@ -1164,11 +1164,10 @@ and must never be nil.")
...
 (defmethod handle-event ((pane push-button-pane) (event pointer-button-release-event))
   (with-slots (armed pressedp) pane
+    (when (and armed pressedp)
+      (activate-callback pane (gadget-client pane) (gadget-id pane)))
     (setf pressedp nil)
-    (when armed
-      (activate-callback pane (gadget-client pane) (gadget-id pane))
-      (setf pressedp nil)
-      (dispatch-repaint pane +everywhere+))))
+      (dispatch-repaint pane +everywhere+)))
```
scymtym: Indentation looks funny here. Should all three forms be inside the `when`?

Backends/CLX/input.lisp
```diff
@@ -141,11 +141,10 @@
       ((:button-press :button-release)
        (let ((modifier-state (clim-xcommon:x-event-state-modifiers *clx-port* state))
              (button (decode-x-button-code code)))
-         (if (and (eq event-key :button-press)
-                  (member button '(#.+pointer-wheel-up+
-                                   #.+pointer-wheel-down+
-                                   #.+pointer-wheel-left+
-                                   #.+pointer-wheel-right+)))
+         (if (member button '(#.+pointer-wheel-up+
```
scymtym: If you look at 14fd2c21311ae13d3753df76b51b597c724d1e6a, the intention seems to have been to generate *no* event for `:button-release` of a scroll wheel "button". Could you explain whether that idea was wrong? Your change would lead to a `pointer-scroll-event` for `:button-release`. Is that the intention?

Backends/CLX/port.lisp
```diff
@@ -193,7 +193,8 @@
                                        event-mask))))
       (port-register-mirror (port sheet) sheet window)
       (when map
-        (xlib:map-window window))))
+        (xlib:map-window window)
```
scymtym: Let me know if you find this kind of feedback annoying. I think the commit message would be more correct like this:

> This issue is not visible with frames because they do draw on the medium *which* flushes clx's buffer. When we only adopt a standalone sheet to a graft *the* problem becomes apparent - *the* window is not mapped until *the* next xlib operation *causes* xlib *to* send buffered requests.

Core/clim-core/panes.lisp
```diff
@@ -913,10 +913,11 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
       (error 'sheet-supports-only-one-child :sheet pane))
     (sheet-adopt-child pane (first contents))))
 
-(defmethod compose-space ((pane single-child-composite-pane) &key width height)
+(defmethod compose-space ((pane single-child-composite-pane)
```
scymtym: Commit message: "If they were not supplied then width and height were *defaulted* NIL *which* lead to an error in make-space-requirements."

Otherwise: :+1: :+1: :smiley: :smiley:

Core/clim-basic/ports.lisp
```diff
@@ -84,28 +87,20 @@
...
-  ;; XXX: also should we allow reading objects from foreign application frames?
-  ;; This was the case on Genera and is requested by users from time to time...
-  #+ (or)
-  (unless (eq *application-frame* (pane-frame focus))
-    (error "frame mismatch in ~S" '(setf port-keyboard-input-focus)))
-  (setf (port-frame-keyboard-input-focus port *application-frame*) focus))
-
-(defgeneric port-frame-keyboard-input-focus (port frame))
-(defgeneric (setf port-frame-keyboard-input-focus) (focus port frame))
+(defgeneric note-input-focus-changed (sheet state)
```
scymtym: Please explain `state` in the documentation string.

Core/clim-basic/ports.lisp
```diff
@@ -84,28 +87,20 @@
...
+  (:documentation "Called when a sheet receives or loses the keyboard
+input focus. This is a McCLIM extension.")
+  (:method (sheet state)
+    (declare (ignore sheet state))))
+
+(defmethod (setf port-keyboard-input-focus) :around (focus port)
```
scymtym: Maybe change `focus` to `sheet` or `focus-sheet`?

Core/clim-basic/ports.lisp
```diff
@@ -84,28 +87,20 @@
...
+  (let ((old-focus (port-keyboard-input-focus port))
+        (result (call-next-method))     ; clim-null cheats
```
scymtym: Can `clim-null` stop cheating by either returning `nil` or actually storing the focused sheet?

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+        (dispatch-event-copy grabbed-sheet event))
+      (return-from distribute-event))
+    ;; Synthetize boundary change events and update the port.
+    (synthetize-enter-exit-events old-pointer-sheet new-pointer-sheet event)
+    (setf (port-pointer-sheet port) new-pointer-sheet)
+    ;; Set the pointer cursor.
+    (when-let ((cursor-sheet (or pressed-sheet new-pointer-sheet)))
+      (let ((old-pointer-cursor (port-lookup-current-pointer-cursor port (event-sheet event)))
+            (new-pointer-cursor (sheet-pointer-cursor new-pointer-sheet)))
+	(unless (eql old-pointer-cursor new-pointer-cursor)
```
scymtym: Indentation uses tabs in these two lines.

Core/clim-basic/ports.lisp
```diff
@@ -275,21 +272,33 @@
...
     ;; Distribute event to the innermost child.
     (dispatch-event-copy new-pointer-sheet event)))
 
+(defmethod distribute-event ((port basic-port) (event keyboard-event))
+  (if-let ((focused (port-keyboard-input-focus port)))
+    (dispatch-event-copy focused event)
+    (let ((sheet (event-sheet event)))
+      ;; SETF the focus to avoid further changes when we hover the
+      ;; pointer over different sheet.
+      (setf (port-keyboard-input-focus port) sheet)
```
scymtym: Maybe we should have a protocol to compute the initial focus and focus changes as a result of events. I'm suggesting this because I see two potential problems:
* It is not clear which sheet will get the keyboard focus before the user has clicked one and there is no way customize that
* There is no way to implement other focus policies ("focus follows mouse" would be an alternative model)

Examples/hierarchy-tool.lisp
```diff
@@ -17,10 +17,19 @@
                                          (make-rectangle* (+ x1 10) (+ y1 10)
                                                           (- x2 10) (- y2 10))))
       (draw-rectangle* pane x1 y1 x2 y2 :filled nil)
-      (draw-circle* pane (/ (- x2 x1) 2) (/ (- y2 y1) 2) 15
+      (let ((cx (/ (- x2 x1) 2))
```
scymtym: It would be cool (and useful) to show which sheet currently has the input focus.

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+;;; delivered. The right thing is not obvious:
+;;;
+;;; - we may assume that event-sheet is set correctly by port
+;;; - we may find the innermost sheet's child and deliver to it
+;;; - we may deliver event to sheet's graft and let it dispatch
+;;;
+;;; Third option would require a default handle-event method to call
+;;; handle-event on its child under the cursor. Such top-down approach
+;;; is appealing if we consider that scrolling pane could filter mouse
+;;; wheel buttons without gross handle-event :around hacks. For now we
```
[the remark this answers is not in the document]
dkochmanski: I've talked about code from the past (I forgot that we've changed that).

```
(defmethod handle-event ((sheet mouse-wheel-scroll-mixin)
                         (event pointer-scroll-event))
  (multiple-value-bind (viewport sheet) (find-viewport-for-scroll sheet)
    (when viewport
      (scroll-sheet sheet
                    (pointer-event-delta-x event)
                    (pointer-event-delta-y event)))))
```

used to be an around method over pointer-button-event. Should I remove the obsolete part of the comment?

Core/clim-basic/ports.lisp
dkochmanski: it is the other way around. if you use default clx mechanisms you will lose focus, in this changed behavior which I've introduced focus will stay. Should I phrase the commit message differently?

Backends/CLX/input.lisp
dkochmanski: previous version generated release event for scrolling (because it was not pressed), so we have queued a spurious event anyway. without fix in push-button gadget in the other place that lead to activating the button callback when we were scrolling. if at all it should be:
```
(cond ((member ...)
            (when press
            (queue scroll))
       (press (queue press))
       (release (queue release)))
```

whether the idea itself was wrong I have no opinion.

Core/clim-basic/events.lisp
```diff
@@ -130,17 +130,16 @@
    (graft-y :reader pointer-event-native-graft-y) ))
 
 (defmacro get-pointer-position ((sheet event) &body body)
-   (with-gensyms (event-var sheet-var x-var y-var)
+   (with-gensyms (event-var sheet-var)
```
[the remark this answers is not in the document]
dkochmanski: done

Core/clim-basic/ports.lisp
```diff
@@ -76,8 +76,13 @@
                         :reader port-text-style-mappings)
    (pointer-sheet :initform nil :accessor port-pointer-sheet
 		  :documentation "The sheet the pointer is over, if any")
+   ;; A difference between grabbed-sheet and pressed-sheet is that the
```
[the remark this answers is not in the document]
dkochmanski: done

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+  ;; - Pressed sheet receives replicated motion events
+  ;; - Pressing/releasing the button assigns pressed-sheet
+  ;; - Pointer motion may result in synthetized boundary events
+  ;; - Events are delivered to the innermost child of the sheet
+  (let ((grabbed-sheet (port-grabbed-sheet port))
+        (pressed-sheet (port-pressed-sheet port))
+        (old-pointer-sheet (port-pointer-sheet port))
+        (new-pointer-sheet (get-pointer-event-sheet event)))
+    (when grabbed-sheet
+      (unless (typep event '(or pointer-enter-event pointer-exit-event))
```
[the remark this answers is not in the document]
dkochmanski: done

Core/clim-basic/ports.lisp
dkochmanski: it wouldn't then make a shallow object copy so I don't think it would be a good thing. also I do not understand all implications of passing initargs to allocate-instance so I'd rather keep it as it is.

Core/clim-basic/ports.lisp
dkochmanski: I don't see how the full sentence would look like. "is not implemented locally"? or "is not implemented by port"? both statements are false: we do implement it locally and we overwrite whatever port does wrt grabbing/pressing etc.

Core/clim-basic/ports.lisp
dkochmanski: fixed in a way that we do not distribute event if it is a release button for one of the scrollings (so I've preserved the intent after all)

Backends/CLX/input.lisp
dkochmanski: fixed in a way that we do not distribute event if it is a release button for one of the scrollings (so I've preserved the intent after all)

Core/clim-core/panes.lisp
dkochmanski: fixed

Core/clim-basic/ports.lisp
dkochmanski: done

Core/clim-basic/ports.lisp
dkochmanski: fixed

Examples/hierarchy-tool.lisp
dkochmanski: done (with few other improvements)

n.b I've added enter/exit event balance counter and it clearly indicates we have a bug. note that this is not a regression, though I'll try to fix it.

Core/clim-core/gadgets.lisp
dkochmanski: done (they should not)

Core/clim-basic/ports.lisp
dkochmanski: done

Core/clim-basic/ports.lisp
dkochmanski: done

Core/clim-basic/ports.lisp
dkochmanski: done

Core/clim-basic/ports.lisp
dkochmanski: done

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+;;;
+;;; Third option would require a default handle-event method to call
+;;; handle-event on its child under the cursor. Such top-down approach
+;;; is appealing if we consider that scrolling pane could filter mouse
+;;; wheel buttons without gross handle-event :around hacks. For now we
+;;; implement the second option with the innermost child. In general
+;;; case we need z-ordering for both strategies. -- jd 2019-08-21
+(defun get-pointer-event-sheet (event &aux (sheet (event-sheet event)))
+  (get-pointer-position (sheet event)
+    (loop
```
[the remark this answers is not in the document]
dkochmanski: done

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
+;;; - we may find the innermost sheet's child and deliver to it
+;;; - we may deliver event to sheet's graft and let it dispatch
+;;;
+;;; Third option would require a default handle-event method to call
+;;; handle-event on its child under the cursor. Such top-down approach
+;;; is appealing if we consider that scrolling pane could filter mouse
+;;; wheel buttons without gross handle-event :around hacks. For now we
+;;; implement the second option with the innermost child. In general
+;;; case we need z-ordering for both strategies. -- jd 2019-08-21
+(defun get-pointer-event-sheet (event &aux (sheet (event-sheet event)))
```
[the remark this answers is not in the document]
dkochmanski: done

Core/clim-basic/ports.lisp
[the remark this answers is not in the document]
dkochmanski: removed part which is not up to date

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,119 @@
...
-          (dispatch-event sheet event)
-          (let ((new-event (shallow-copy-object event)))
-            (get-pointer-position (sheet event)
-              (setf (slot-value new-event 'x) x
-                    (slot-value new-event 'y) y
-                    (slot-value new-event 'sheet) sheet))
-            (change-class new-event 'pointer-motion-event)
-            (dispatch-event sheet new-event))))
-    (return-from distribute-event))
+;;; This function determines the sheet to which pointer should be
```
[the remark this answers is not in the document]
dkochmanski: done

Backends/CLX/port.lisp
dkochmanski: fixed

Core/clim-basic/ports.lisp
dkochmanski: it could, but null port has around method for distribute-event which ignores all events. I'd rather not change this behavior in this particular pull request. That means in particular that setf'ing port-keyboard-input-focus does not make any sheet focused, hence port-keyboard-input-focus justifully returns NIL (and part of setf protocol is to return assigned variable). So this "cheat" is arguably the right thing to do. Should I remove the ocmment?

Core/clim-basic/ports.lisp
dkochmanski: Customization should be made in a distribute-event method specialization (basic-port represents "default" behavior). I.e if we have wanted to implement "focus follows mouse" for CLX port (putting aside for now the fac that we'd also have clx's default behavior for pressed sheets and grabbed sheets), it would be enough to put in port.lisp

```
(defmethod distribute-event ((port clx-port) event)
  (dispatch-event (event-sheet event) event))
```

of course this method can accomodate any number of customizations with regard to changing what is focused or pressed. I don't think we want to make it configurable at runtime (yet?), so I'd rather address this problem when there is a need for it.

That's regarding the second point.

Regarding the first point I do agree, this PR does not maintain frame focus (so I'll need to fix it before we merge the branch). If you click on the window "decoration bar" focus will remain in the old sheet even if it is in a different frame.

As of defaults they are not much of a problem in case of frames, because their loop calls with-input-focus which sets the focused sheet to the frame's query-io (overwriting whatever we would set as a "default" strategy. When I ponder a little with frame "saved" focus I'll leave another comment here with a proposed solution.

Core/clim-basic/ports.lisp
dkochmanski: done

Core/clim-basic/ports.lisp
dkochmanski: This is how it currently looks like:
```
(defmethod (setf port-keyboard-input-focus) :around (sheet port)
  (let ((old-sheet (port-keyboard-input-focus port))
        ;; Result may not reflect the actual focus.
        (result (call-next-method))
        (new-sheet (port-keyboard-input-focus port)))
    (prog1 result
      (unless (eq old-sheet new-sheet)
        (note-input-focus-changed old-sheet nil)
        (note-input-focus-changed new-sheet t)))))
```

dkochmanski:
added wip tag until we regain sheet focus on window exposure (right now keyboard focus stays the same until we click on a sheet or with-input-focus is called).

Core/clim-basic/ports.lisp
```diff
@@ -101,7 +101,15 @@ is T when focused and NIL otherwise. This is a McCLIM extension.")
...
       (unless (eq old-sheet new-sheet)
         (note-input-focus-changed old-sheet nil)
-        (note-input-focus-changed new-sheet t)))))
+        (note-input-focus-changed new-sheet t)
+        (when-let ((top-sheet (get-top-level-sheet new-sheet)))
+          (setf (focused-sheet top-sheet) new-sheet))))))
+
+(defmethod (setf port-keyboard-input-focus) ((sheet top-level-sheet-mixin) port)
+  (if-let ((focus (focused-sheet sheet)))
+    (call-next-method focus port)
```
scymtym: I had to change this to
```lisp
(defmethod (setf port-keyboard-input-focus) ((sheet top-level-sheet-mixin) port)
  (if-let ((focus (focused-sheet sheet)))
    (if (eq focus sheet)
        (call-next-method)
        (setf (port-keyboard-input-focus port) focus))
    (call-next-method)))
```
Otherwise `call-next-method` would be called with a different set of applicable methods.

dkochmanski: you are right, thank you. fixed

dkochmanski:
I've managed to fix the issue with duplicated boundary events. I'm satisfied with the current state of this branch (I believe I haven't left any peer review remark without a comment waiting for answer or solved it on the spot).

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +181,184 @@
...
+;;; the mirror. port-pointer-sheet is NIL after this operation.
+;;;
+;;; port-enter-event dispatches the event for the mirror and then to
+;;; each parent of the innermost sheet under the pointer from the
+;;; highest unmirrored ancestor. port-pointer-sheet is assigned to the
+;;; innermost child.
+;;;
+;;; Other pointer events result in finding a common ancestor of the
+;;; old port-pointer-sheet and the innermost sheet under the
+;;; pointer. First we synthetize exit events and then enter events.
```
dkochmanski: after the thought, I want to test synthetize-boundary-events a little more and if possible simplify the code.

scymtym:
I got this while randomly clicking the color buttons in clim-fig:
```
debugger invoked on a SB-PCL::NO-APPLICABLE-METHOD-ERROR in thread
#<THREAD "#<MCCLIM-TRUETYPE::CLX-TTF-PORT :HOST \"\" :DISPLAY-ID 1 {10085DC763}>'s event process." RUNNING
   {1008C233C3}>:
    #<STANDARD-GENERIC-FUNCTION CLIM:DISPATCH-EVENT (10)>
  when called with arguments
    (#<CLIM-CLX:CLX-GRAFT {1008BBAE63}>
     #<CLIM:POINTER-EXIT-EVENT {100225ACD3}>).
See also:
  The ANSI Standard, Section 7.6.6
```

Another thing I noticed is that some of the accepting values demos seemed to behave strangely: when accepting multiple strings, clicking the text fields and writing text into them made some of the cursors disappear. In the "Accepting With Gadgets" case, clicking some of the gadgets caused `C-]` to no longer work (I don't know whether that is a regression).

Core/clim-basic/ports.lisp
dkochmanski: OK, I think this one should be fine now.

dkochmanski:
re clicking gadgets: this is expected behavior, when you click a gadget it takes over keyboard focus, so when you press C-] it is delivered as a keyboard event to said gadget (not to the input sheet). 

In other words clicking on a gadget makes it take focus. N.b we can now implement keyboard gestures on gadgets, i.e arrows to increase/decrease slider values.

regarding multiple stirngs: is this still happening? if so, could you describe what you are doing? I think I can't reproduce this.

Core/clim-core/gadgets.lisp
```diff
@@ -1164,9 +1164,9 @@ and must never be nil.")
 
 (defmethod handle-event ((pane push-button-pane) (event pointer-button-release-event))
   (with-slots (armed pressedp) pane
-    (setf pressedp nil)
-    (when armed
-      (activate-callback pane (gadget-client pane) (gadget-id pane))
+    (when (and armed pressedp)
```
scymtym: Fixing the issue mentioned in the commit message is good, of course. 

This commit has another side-effect: since `pressedp` now stays true until after `activate-callback` returns, so the button is rendered as pressed while `activate-callback` is executing. Pressing the button again can cause the pressed state to be reset when one of the `activate-callback`s returns, leading to an inconsistency. For this new behavior to work, we would have to count active "presses" for the button, I think.

scymtym: So how about this?

```diff
modified   Core/clim-core/gadgets.lisp
@@ -1128,7 +1128,7 @@ and must never be nil.")
                              activate/deactivate-repaint-mixin
                              enter/exit-arms/disarms-mixin
                              sheet-leaf-mixin)
-  ((pressedp          :initform nil)
+  ((pressed-count     :initform 0)
    (show-as-default-p :type boolean
                       :initform nil
                       :initarg :show-as-default-p
@@ -1158,26 +1158,26 @@ and must never be nil.")
                          :height 2*border-thickness)))
 
 (defmethod handle-event ((pane push-button-pane) (event pointer-button-press-event))
-  (with-slots (pressedp) pane
-    (setf pressedp t)
+  (with-slots (pressed-count) pane
+    (incf pressed-count)
     (dispatch-repaint pane +everywhere+)))
 
 (defmethod handle-event ((pane push-button-pane) (event pointer-button-release-event))
-  (with-slots (armed pressedp) pane
-    (when (and armed pressedp)
-      (activate-callback pane (gadget-client pane) (gadget-id pane)))
-    (when pressedp
-      (setf pressedp nil)
-      (dispatch-repaint pane +everywhere+))))
+  (with-slots (armed pressed-count) pane
+    (when (plusp pressed-count)
+      (when armed
+        (activate-callback pane (gadget-client pane) (gadget-id pane)))
+      (when (zerop (decf pressed-count))
+        (dispatch-repaint pane +everywhere+)))))
 
 (defmethod handle-repaint ((pane push-button-pane) region)
   (declare (ignore region))
-  (with-slots (armed pressedp) pane
+  (with-slots (pressed-count) pane
     (with-bounding-rectangle* (x1 y1 x2 y2) (sheet-region pane)
       (draw-rectangle* pane x1 y1 x2 y2 :ink (effective-gadget-background pane))
       (draw-bordered-rectangle* pane x1 y1 x2 y2
                                 :border-width 1
-                                :style (if (and pressedp armed) :inset :outset))
+                                :style (if (plusp pressed-count) :inset :outset))
       (let* ((x-spacing (pane-x-spacing pane))
              (y-spacing (pane-y-spacing pane))
              (border-thickness *3d-border-thickness*)
```

dkochmanski: couldn't we get away with:
```
(when pressedp
  (setf pressedp nil)
  (when armed (activate-callback))
  (dispatch-repaint))
```
?

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,117 @@
...
+      (gpes-rec (event-sheet event) x y))))
+
+;;; Function is responsible for making a copy of an immutable event
+;;; and adjusting its coordinates to be in the target-sheet
+;;; coordinates. Optionally it may change event's class.
+(defun dispatch-event-copy (target-sheet event &optional new-class
+                            &aux (sheet (event-sheet event)))
+  (if (and (eql target-sheet sheet)
+           (or (null new-class)
+               (eql new-class (class-of sheet))))
```
scymtym: Shouldn't this be `(class-of event)`?

Core/clim-core/gadgets.lisp
scymtym: Yes, rearranging the body seems best. The "pressed count" idea would have been nice, but it actually doesn't work with the way command loops nest.

dkochmanski: fix pushed

Core/clim-basic/ports.lisp
dkochmanski: yes, fix pushed

dkochmanski:
replaced by #833

PR #830, opened 2019-09-02
pointer-tracking: re-adding some franz-specific functions

### LINKED ISSUE #817: pointer-input-rectangle and other functions removed
With commit 9f36c9e4 functions `pointer-input-rectangle`,  `pointer-input-rectangle*` and `pointer-place-rubber-band-line*` were been removed.

I used some of them in some of my project so I like re-adding them.

In the commit message is written:

> If someone feels like re-adding them I think that such
> utilities should be put in climf package (clim-franz-compat) and kept
> in Vendor/ toplevel directory.

That's means `McCLIM/Vendor/Franz` directory or `McCLIM/Franz` or `McCLIM/Extensions/Franz/` or `McCLIM/Extensions/Vendor/Franz/`

admich:
With commit 9f36c9e4 the functions pointer-input-rectangle,
pointer-input-rectangle* and pointer-place-rubber-band-line* were
removed. These are functions documented in the Franz Manual but not in
CLIM Specifications.

Here I re-add them but now the functions are exported in clime package.

This Fix #817 issue.

### DIFF
--- a/Extensions/Franz/franz.lisp
+++ b/Extensions/Franz/franz.lisp
@@ -1,0 +1,149 @@
+(in-package :climi)
+
+(defun dragging-drawing (stream drawer &key (finish-on-release t)
+                         (pointer (port-pointer (port stream)))
+                         multiple-window)
+  "Draws something simple in response to pointer events for
+`pointer' and returns the coordinates of the pointer when the
+function finishes. The function finishes when mouse button one is
+no longer held down if `finish-on-release' is true; if it is
+false, it finishes when the mouse is clicked. `Drawer' should
+draw something on `stream', and is called with tree arguments:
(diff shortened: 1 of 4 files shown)


dkochmanski:
lgtm, thanks!

admich:
@crunsk, I suppose that you have cloned the McCLIM repository in the ~/quicklisp/local-project directory, and than you fetch the new commits. 

I'm not an expert but I have found that it is better to clone the repositories in ~/common-lisp directory. Quicklisp find  them and also update the available systems each time it is  started.

kc-:
@crunsk : Quicklisp has its own registry -- refresh it by doing `(ql:register-local-projects)` if you run into this problem again.

crunsk:
Ok, thanks admich and kc-!

PR #831, opened 2019-09-02
Fix #801: while filtering completions, use pathname-type from file.

### LINKED ISSUE #801: Filename completion regression
1) In clim-listener, run `Show File` command by typing `,show file↵`.
2) An accepting-values dialog will appear, right click to get possible completions.
3) In the Completion dialog, file names now don't have extensions so after selecting any file and submitting accepting-values dialog, You will get a "File not found" error.

kc-:


nsrahmad:
Looks good, Thank you.

PR #833, opened 2019-09-06
distribute-event: rewrite method for basic-port

dkochmanski:
We reintroduce a clean-slate implementation of the distribute-event
method for basic-port. Differences from previous implementation:

- No auxlliary methods (so port may simply specialize on port)
- Corrected coordinate fixup (get-pointer-position was buggy)
- Comments throughout the code explaining what is being done
- Shorter function body
- Pressed sheet implementation (i.e for scrolling)
- Grabbed sheet implementation
- Correctly synthetized enter/leave events for unmirrored sheets

Pressed sheet is activated when we press the pointer button until we
release it. This sheet receives duplicated motion and release
events.

Grabbed sheet hijacks all events received from the port (except
boundary events which are ignored). Sheet grabbing may be achieved
programmatically with a macro WITH-POINTER-GRABBED.

Enter and leave events must be synthetized for sheets without
mirrors. Carefully commented implementation is provided which avoids
generating spurious duplicated events (unlike the old one). A
clear (event hierarchy -wise) distinction is made between ungrab
enter/leave events and "ordinary" ones.


Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+                          (return-from common-ancestor nil))
+                         ((sheet-ancestor-p sheet-b sheet-a)
+                          (return-from common-ancestor sheet-a))
+                         (t (setf sheet-a (sheet-parent sheet-a)))))))
+         (let* ((old-sheet (port-pointer-sheet port))
+                (new-sheet (compute-pointer-event-sheet event))
+                (common-ancestor (common-ancestor old-sheet new-sheet)))
+           ;; distribute exit events (innermost first)
+           (do ((sheet old-sheet (sheet-parent sheet)))
+               ((or (null sheet)
```
scymtym: Is `(null sheet)` needed? If `common-ancestor` is non-`nil`, it is encountered before `sheet` becomes `nil`. If `common-ancestor` is `nil`, `(null sheet)` is the same as `(eq sheet common-ancestor)`.

Also, it is mildly confusing that basically the same operation is written using `loop` in line 266 and using `do` here.

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+                    (eq sheet common-ancestor)
+                    (sheet-direct-mirror sheet)))
+             (synthetize-exit sheet))
+           ;; distribute enter events (innermost last)
+           (do ((sheet new-sheet (sheet-parent sheet))
+                (sheets nil))
+               ((or (null sheet)
+                    (eq sheet common-ancestor)
+                    (sheet-direct-mirror sheet))
+                (dolist (sheet sheets)
```
scymtym: I would much prefer `(loop with sheets … finally (map nil #'synthetize-enter sheets))`.

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+;;; port-enter-event dispatches the event for the mirror and then to
+;;; each parent of the innermost sheet under the pointer from the
+;;; highest unmirrored ancestor. port-pointer-sheet is assigned to the
+;;; innermost child.
+;;;
+;;; Other pointer events result in finding a common ancestor of the
+;;; old port-pointer-sheet and the innermost sheet under the
+;;; pointer. First we synthetize exit events and then enter events.
+(defun synthetize-boundary-events (port event)
+  (labels ((synthetize-enter (sheet)
```
scymtym: My spell-checker says "synthesize". Are there multiple spellings?

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+          (set-sheet-pointer-cursor port (event-sheet event) new-pointer-cursor))))
+    ;; Handle some events specially.
+    (typecase event
+      ;; Pressing pointer button over a sheet makes a sheet
+      ;; pressed. Pressed sheet is assigned only when there is
+      ;; currently none.
+      (pointer-button-press-event
+       (when (null pressed-sheet)
+         (setf (port-pressed-sheet port) new-pointer-sheet))
+       (dispatch-event-copy new-pointer-sheet event 'window-manager-focus-event))
```
scymtym: Is `window-manager-focus-event` needed for this pull request? If not, please move it to the one concerned with input focus changes.

Core/clim-basic/ports.lisp
```diff
@@ -76,8 +76,13 @@
                         :reader port-text-style-mappings)
    (pointer-sheet :initform nil :accessor port-pointer-sheet
 		  :documentation "The sheet the pointer is over, if any")
+   ;; The difference between grabbed-sheet and pressed-sheet is that
```
scymtym: Commit message:
- "(so port may simply specialize on port)" Not sure what that means
* "Corrected coordinate fixup (get-pointer-position was buggy)" Is this right? I don't see changes to `get-pointer-position`
* "- Pressed sheet implementation (i.e for scrolling)\- Grabbed sheet implementation" Those are not new, but restored (they were removed in 23b66e37).
* "synthetized" -> "synthesized" according to my spell checker
* "Pressed sheet is activated when …" -> "The `pressed-sheet` slot is bound to a sheet when …"
* "with a macro WITH-POINTER-GRABBED" -> "with the macro WITH-POINTER-GRABBED"

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+                      (untransform-position (sheet-transformation child) x y)
+                    (map-within-mirror function child child-x child-y))))
+              sheet x y)))
+    ;; FIXME clx generates uneven number of ungrab-enter/ungrab-leave
+    ;; events what may disrupt the enter/exit event balance for a
+    ;; sheet. Not distributing such events makes overall UX worse, for
+    ;; instance user needs first to click on a window to press an
+    ;; armed push-button. What's the right thing? -- jd 2019-08-28
+    (typecase event
+      ;; (pointer-ungrab-enter-event (warn "ungrab enter event"))
```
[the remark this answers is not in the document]
dkochmanski: It is here to signify that we may have "spurious" enter events (as indicated in the comment above)

Core/clim-basic/ports.lisp
dkochmanski: I think it was needed, but I'll think about it later. do is used to reflect the similarity between enter/leave synthesize (but in reverse order) (not that in the first we distribute thing eagerly and in the other we first push and then we use the order on the list).

Core/clim-basic/ports.lisp
dkochmanski: for me do is much clearer here.

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
   (with-slots (properties) port
     (setf (getf properties indicator) value)))
 
+;;; This function determines the sheet to which the pointer event
+;;; should be delivered. The right thing is not obvious:
```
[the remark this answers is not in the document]
dkochmanski: port may be aware that some sheets are not mirrored and look for the correct sheet on its own (and then we should not touch it). another port could only deliver events to sheets which are mirrored. yet another may have all sheets mirrored. I mean that it is not specified what is the right thing and this comment is to highlight our decision for `clim:basic-port`. Different strategies could be specialized on `clim:port` subclass. Word `may` reflects this uncertainity.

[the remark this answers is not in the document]
dkochmanski: by our decision I mean: ports work directly with mirrored sheets.

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+;;; - we may assume that event-sheet is set correctly by port
+;;; - we may find the innermost sheet's child and deliver to it
+;;; - we may deliver event to sheet's graft and let it dispatch
+;;;
+;;; Third option would require a default handle-event method to call
+;;; handle-event on its child under the cursor. For now we implement
+;;; the second option with the innermost child. In general case we
+;;; need z-ordering for both strategies. -- jd 2019-08-21
+(defun compute-pointer-event-sheet (event &aux (sheet (event-sheet event)))
+  (labels ((gpes-rec (sheet x y)
```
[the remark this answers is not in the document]
dkochmanski: it was for readibility, but I may change it, maybe it is only more readable to me.

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+           (or (null new-class)
+               (eql new-class (class-of event))))
+      (dispatch-event sheet event)
+      (let ((new-event (shallow-copy-object event)))
+        (get-pointer-position (target-sheet new-event)
+          (setf (slot-value new-event 'x) x
+                (slot-value new-event 'y) y
+                (slot-value new-event 'sheet) target-sheet))
+        (when new-class
+          (change-class new-event new-class))
```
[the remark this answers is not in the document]
dkochmanski: it is not guaranteed that this other class will have the same slots as the original one. change-class reflects the purpose much better (imho).

Core/clim-basic/ports.lisp
dkochmanski: it may be Polish influence (syntetyzuj), I'll change that.

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+;;; highest unmirrored ancestor. port-pointer-sheet is assigned to the
+;;; innermost child.
+;;;
+;;; Other pointer events result in finding a common ancestor of the
+;;; old port-pointer-sheet and the innermost sheet under the
+;;; pointer. First we synthetize exit events and then enter events.
+(defun synthetize-boundary-events (port event)
+  (labels ((synthetize-enter (sheet)
+             (dispatch-event-copy sheet event 'pointer-enter-event)
+             (setf (port-pointer-sheet port) sheet))
```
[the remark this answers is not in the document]
dkochmanski: I'm not sure, I'll think about it

Core/clim-basic/ports.lisp
dkochmanski: "(so port may simply specialize on port)" Not sure what that means

So `a port` may subclass `clim:port` and specialize the protocol methods.

Core/clim-basic/ports.lisp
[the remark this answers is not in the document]
scymtym: Using the event logger I mentioned, I have convinced myself that this is not correct: consider a child that is smaller than its parent (both sheets don't have a direct mirror) and the pointer moves from the child to the parent. Then `new-sheet` is the parent, `old-sheet` is the child and `common-ancestor` is the parent. This results in a single exit event which also sets `port-pointer-sheet` to `nil`. No enter events are generated (and rightfully so). Thus the overall effect is that `port-pointer-sheet` is set to `nil` instead of the parent.

Core/clim-basic/ports.lisp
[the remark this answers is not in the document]
scymtym: I don't agree with the purpose part. `change-class` has to "diff" the slots of the old and the new class, introspecting the slot list a second time, not to mention potential type checks and initarg checks -- this is very expensive. Why not allocate an instance of the required class and copy only the relevant slots right from the start?

This would be enough:
```lisp
(defun shallow-copy-object (original &optional (new-class (class-of original)))
  (let ((copy (allocate-instance new-class)))
    (dolist (slot (mapcar #'c2mop:slot-definition-name
                          (c2mop:class-slots new-class)))
      (when (and (slot-exists-p original slot)
                 (slot-boundp original slot))
        (setf (slot-value copy slot)
              (slot-value original slot))))
    copy))
```

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+              for over-sheet = child then (sheet-parent over-sheet)
+              do (synthetize-exit over-sheet)
+              until (eql over-sheet sheet)))))
+      (otherwise
+       (flet ((common-ancestor (sheet-a sheet-b)
+                (loop
+                   (cond ((or (null sheet-a) (null sheet-b)
+                              (sheet-direct-mirror sheet-a)
+                              (sheet-direct-mirror sheet-b))
+                          (return-from common-ancestor nil))
```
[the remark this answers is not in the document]
dkochmanski: I think that return-from better signals the intention

Core/clim-basic/ports.lisp
dkochmanski: fixed

Core/clim-basic/ports.lisp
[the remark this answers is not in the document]
dkochmanski: fixed

Core/clim-basic/ports.lisp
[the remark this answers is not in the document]
dkochmanski: fixed (shallow-copy-object was modified in a separate commit before the distribute-event changes)

Core/clim-basic/ports.lisp
```diff
@@ -171,107 +176,183 @@
...
+
+;;; Boundary-events generated by port apply only to mirrored
+;;; sheets. For sheets without a mirror we must synthetize
+;;; boundary-events. Here is an outline what it does:
+;;;
+;;; pointer-exit-event first exits all unmirrored sheets from the last
+;;; port-pointer-sheet upwards and finally it dispatches the event for
+;;; the mirror. port-pointer-sheet is NIL after this operation.
+;;;
+;;; port-enter-event dispatches the event for the mirror and then to
```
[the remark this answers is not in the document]
dkochmanski: fixed

Core/clim-basic/ports.lisp
dkochmanski: fixed

Core/clim-basic/ports.lisp
dkochmanski: fixed

Core/clim-basic/ports.lisp
scymtym: Missed two in the commit message.

dkochmanski:
On this branch line wrapping does not work. Steps to reproduce:
1. open `demodemo`
2. select `seos baseline and wrapping`
3. try to wrap a line

- expected: line wraps
- observed: nothing happens

this is not happening on master branch

dkochmanski:
(it doesn't work when selected from menu, when we wrap by default all works as expected)

scymtym:
> On this branch line wrapping does not work. Steps to reproduce:

I believe this happens because the root menu gets its button release event (due to grabbing) before the leaf menu.

dkochmanski:
superseded by the other pr

PR #834, opened 2019-09-06
ports: implement click to focus policy

dkochmanski:
This is the second part of the split PR #822. Contains change from #833 since it depends on it. Should be merged after its dependency.


dkochmanski:
Replaced by #851

PR #836, opened 2019-09-09
Clouseau: Suppress ill-defined and unhelpful structure sharing indication

scymtym:
* Objects for which multiple occurrences in terms of EQ are not meaningful no longer participate in structure sharing indication: numbers, characters, `NIL`

* The "Specializer Usage" section of classes is presented without structure sharing indication since the class being inspected re-occurs in each item by definition.


kc-:
Have you had time to think about the suggestion with respect to "number of occurrences"? I'll attach a screenshot of an example of a situation I'm talking about.
shared-structure--asdf-example

scymtym:
> Have you had time to think about the suggestion with respect to "number of occurrences"?

I focused on making a decision based on the object type and forgot about that one, thank you. 

Would you prefer to limit the number of arrows (assuming we keep using arrows for now) to a certain number or draw no arrows at all if there would be too many?

scymtym:
> I'll attach a screenshot of an example of a situation I'm talking about.

Unrelatedly, there should be a "RIGHT ARROW" character between the slot and its value. Could your font be missing that character?

kc-:
> Would you prefer to limit the number of arrows (assuming we keep using arrows for now) to a certain number or draw no arrows at all if there would be too many?

I'd suggest drawing no arrows at all, and maybe instead using something like a [shared N times] overlay/hint/ drawn at the cursor, for example. My reasoning is that limiting the number of errors might give a false sense of how often some object is reused, if users come to rely on that information.

kc-:
>  Could your font be missing that character?
I'll look into this.

scymtym:
> I'd suggest drawing no arrows at all, and maybe instead using something like a [shared N times] overlay/hint/ drawn at the cursor, for example. My reasoning is that limiting the number of errors might give a false sense of how often some object is reused, if users come to rely on that information.

I implemented this suggestion as well as the possibility to limit the number of drawn arrows:

* The new parameter `*occurrence-highlight-limit*` controls how many occurrences are highlighted using Bezier arrows.

* The new parameter `*occurrence-highlight-disable-threshold*` controls at which number of occurrences highlighting using Bezier arrows is disabled entirely.

Setting `*occurrence-highlight-disable-threshold*` to a lower or equal value results in the behavior you suggested.

Whenever these constraints cause occurrences to not be highlighted, the number of omitted occurrences is noted in a text box.

PR #841, opened 2019-09-14
Fixes for RasterImage and Postscript backends, command output destinations

scymtym:
* Sort out dependency problems of the Postscript backend (The backend had forward reference a bunch of things because `clim-core` depended on it in the output destination code in `commands.lisp`)

* Integrate the RasterImage and Postscript backends with the output destination mechanism in way that registers the backend output destination when the backend is loaded.

* Add embryonic unit tests for the RasterImage and Postscript backends

* Fix a few problems in the RasterImage backend and add code to compute the width and height of the produced raster images automatically.

* Improve and document the listener to take advantage of the output destination improvements.


Backends/PostScript/output-destination.lisp
```diff
@@ -0,0 +1,37 @@
...
+    (continuation (destination postscript-destination))
+  (call-next-method (lambda ()
+                      (with-output-to-postscript-stream
+                          (ps-stream *standard-output*)
+                        (let ((*standard-output* ps-stream))
+                          (funcall continuation))))
+                    destination))
+
+(pushnew '("Postscript File" postscript-destination)
+         climi::*output-destination-types* :test #'equal)
```
dkochmanski: If we move it from core `file-destination` and `invoke-with-standard-output` should be exported from climb package (and at least mentioned as a comment in `developer-manual.texi` as something worth documenting).

Also instead of `pushnew` we should have a function: `register-destination-type` (also in climb package and also at least mentioned in the manual in a comment).

dkochmanski: I wonder if we should also make available in climb `output-destination` and `stream-destination` classes.

Backends/PostScript/output-destination.lisp
```diff
@@ -0,0 +1,37 @@
...
+  ())
+
+(defmethod climi::invoke-with-standard-output
+    (continuation (destination postscript-destination))
+  (call-next-method (lambda ()
+                      (with-output-to-postscript-stream
+                          (ps-stream *standard-output*)
+                        (let ((*standard-output* ps-stream))
+                          (funcall continuation))))
+                    destination))
```
dkochmanski: maybe: `(with-output-to-ps-stream (*standard-output* *standard-output*) …)`

Backends/PostScript/test/package.lisp
```diff
@@ -0,0 +1,31 @@
...
+;;; This library is distributed in the hope that it will be useful,
+;;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;;; Library General Public License for more details.
+;;;
+;;; You should have received a copy of the GNU Library General Public
+;;; License along with this library; if not, write to the
+;;; Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+;;; Boston, MA  02111-1307  USA.
+
```
dkochmanski: This and the test added in ` rasterimage: add a basic smoke test ` should be unified into one system under `Tests/`. The test suite invocation should accept the output destination argument for running. Then the test op would be (if we must use uiop)
`(uiop:symbol-call '#:output-destination.test '#:run-tests "Postscript File" :file "foo.ps")` and similar for raster. Similar tests should be defined for PDF backend.

Backends/RasterImage/test/smoke.lisp
```diff
@@ -0,0 +1,30 @@
...
+
+(test smoke
+  "Smoke test for the raster image backend."
+
+  (finishes
+    (mcclim-raster-image:with-output-to-raster-image-file
+        (stream "raster-image-test.png")
+      (clim:draw-rectangle* stream 30 30 60 60 :ink clim:+blue+)
+      (clim:draw-rectangle* stream 0 0 30 30 :ink clim:+red+)
+      (clim:draw-rectangle* stream 60 60 90 90 :ink clim:+yellow+))))
```
dkochmanski: ditto

Backends/RasterImage/port.lisp
```diff
@@ -30,7 +30,18 @@
...
+(defmacro with-raster-image-port ((port-var server width height) &body body)
+  `(invoke-with-raster-image-port
+    (lambda (,port-var) ,@body) ,server ,width ,height))
+
+(defun invoke-with-raster-image-port (cont server width height)
+  (let* ((path (list server :width width :height height))
+         (port (find-port :server-path path)))
+    (unwind-protect
+         (funcall cont port)
+      (destroy-port port))))
```
dkochmanski: this should be a general utility since the same semantics are used for postscript and pdf as well. i.e `(with-temporary-port ((port-var path) &body body)`

Backends/RasterImage/output-to-image.lisp
```diff
@@ -44,6 +44,10 @@
 
 (defmacro with-output-to-rgba-image ((stream-var image &rest options)
                                      &body body)
+  `(with-output-to-iamge (,stream-var ,image ,@options) ,@body))
```
dkochmanski: `with-output-to-image` (typo)

dkochmanski: either way macro `with-output-to-rgba-image` should be removed. symbol is not exported neither it is used anymore.

Core/clim-core/commands.lisp
```diff
@@ -1059,49 +1056,54 @@ examine the type of the command menu item to see if it is
...
 (define-presentation-method accept
     ((type stream-destination) stream (view textual-view)
-     &key)
-  (let ((dest (eval (accept 'form
-			    :stream stream
-			    :view view
-			    :default *standard-output*))))
+                               &key
+                               (default '*standard-output*)
+                               (prompt "stream form"))
```
dkochmanski: maybe `stream-var` or `stream-symbol` or `stream-name`?

Backends/RasterImage/output-destination.lisp
```diff
@@ -0,0 +1,58 @@
...
+                                  :width (width destination)
+                                  :height (height destination)
+                                  :border-width (border-width destination)
+                                  :recording-p nil)
+                        (let ((*standard-output* stream))
+                          (funcall continuation))))
+                    destination))
+
+(pushnew '("Raster Image File" raster-image-destination)
+         climi::*output-destination-types* :test #'equal)
```
dkochmanski: see comment in PS output-destination commit.

also the same additions should be put in PDF backend.

Documentation/Manual/Texinfo/chap-listener.texi
```diff
@@ -189,3 +190,22 @@ the pretty names used at the interactor map to a function name which
...
+be invoked with the @cl{:output-destination} keyword, for example
+
+@lisp
+CLIM-USER> ,Show Class Superclasses (class) standard-class
+  (keywords) :output-destination
+  (output-destination) Postscript File
+  (destination file) /tmp/output.ps
+@end lisp
+
+will redirect the command's output to the file @file{/tmp/output.ps}.
```
dkochmanski: it is worth to think about extending developer-manual with description how to register new output destinations. at least providing lengthy comment with TODO.

Backends/PostScript/clim-postscript.asd
```diff
@@ -1,17 +1,17 @@
...
 ;;; "included" part of McCLIM. Hence the defsystem is here.
-;;; 
+;;;
 ;;; Move the whole system to the Modules/ directory - dk
 
 (defsystem #:clim-postscript
-  :depends-on (#:clim-basic #:clim-postscript-font)
+  :depends-on (#:clim-basic
+               #:mcclim-backend-common
+               #:clim-postscript-font)
```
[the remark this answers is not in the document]
scymtym: The new dependency is on the command output destination stuff in `Core/clim-core/commands.lisp`. How would you prefer to express this? `:depends-on (#:clim-core #:clim-postscript-font)`?

Backends/PostScript/output-destination.lisp
scymtym: Done except for the "If we move it from core …" part which i don't understand.

Backends/PostScript/output-destination.lisp
scymtym: Done.

Documentation/Manual/Texinfo/chap-listener.texi
scymtym: Done.

Backends/RasterImage/output-destination.lisp
scymtym: Done and done.

Backends/RasterImage/port.lisp
scymtym: Done.

Backends/PostScript/test/package.lisp
scymtym: All output destination tests now use the `print-test-page` function so the repeated code issue is gone. Tests for the PDF backend have been added.

Backends/RasterImage/output-to-image.lisp
scymtym: Done.

Core/clim-core/commands.lisp
scymtym: The argument is evaluated so it really is a form.

Backends/PostScript/clim-postscript.asd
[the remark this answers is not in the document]
dkochmanski: I think it still depends also on clim-basic, so that would be

`:depends-on (#:clim-basic #:clim-core #:clim-postscript-font)`

otoh I don't really like the fact that postscript backend depends on core. I'd prefer to have all backends depend only on clim-basic (so they may be used without loading clim-core); main motive is to allow future speration for the low level abstraction (silica/dragonstone) and clim.

I understand that this is about output-destination in commands, but is there a chance to have some generic mechanism on top of which command mechanism is built?

Backends/PostScript/output-destination.lisp
dkochmanski: Given we move this snippet from modules in `Core/` directory and we still depend on internal symbols from `CLIMI` package, then these symbols should be promoted to be exported from `CLIMB` package and used that way. We don't want backends to depend on `CLIMI` internal symbols.

scymtym: In the latest version, the symbols are exported from the `climb` package.

Backends/PostScript/clim-postscript.asd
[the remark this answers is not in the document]
scymtym: We could move the code for `invoke-with-standard-output`, `{stream,file}-destination`, `register-output-destination` and `*output-destination-types*` to the `mcclim-backend-common` system, i.e. a new file `Backends/common/output-destination.lisp`.

Backends/common/ports.lisp
```diff
@@ -193,3 +193,16 @@
               (or (try-translators tr1 nil)
                   (try-translators (find-presentation-translators selected-type requested-type table-2)
                                    tr1)))))))))
+
+;;; Port macros
+
+(defmacro with-port ((port-var server &rest args &key &allow-other-keys)
+                     &body body)
+  `(invoke-with-port (lambda (,port-var) ,@body) ,server ,@args))
```
dkochmanski: I see that this is used in backends via `climi::with-port`, in this case we should export the symbol from climb package to avoid such access.

Backends/RasterImage/output-destination.lisp
```diff
@@ -0,0 +1,58 @@
...
+;;;; License along with this library; if not, write to the
+;;;; Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+;;;; Boston, MA  02111-1307  USA.
+
+(cl:in-package #:mcclim-raster-image)
+
+(defclass raster-image-destination (file-destination)
+  ((%file-format :initarg :file-format
+                 :reader file-format)
+   (climi::element-type :allocation :class
```
dkochmanski: ditto, this symbol if meant for backends should be exported from climb package

Backends/RasterImage/output-to-image.lisp
```diff
@@ -61,53 +61,64 @@
...
                                                 (border-width 0) (recording-p t))
-  (let ((port (find-port :server-path (list server
-                                            :width width :height height)))
-        (result nil))
-    (let* ((top-level-sheet (make-raster-top-level-sheet port format))
-           (vbox (make-instance 'vbox-pane :port port))
-           (border-pane (make-instance 'climi::border-pane
-                                       :port port
-                                       :border-width border-width)))
+  (climi::with-port (port server :width 1 :height 1)
```
dkochmanski: ditto

Backends/common/ports.lisp
scymtym: Done.

Backends/RasterImage/output-destination.lisp
scymtym: Replaced `:allocation :class` with methods on `destination-element-type` which is already exported.

Backends/RasterImage/output-to-image.lisp
scymtym: Done.

dkochmanski:
lgtm, thank you!

PR #845, opened 2019-09-23
Automatable tests and test utilities

scymtym:
* Use `fiveam` to allow automated test execution and, more importantly, test result collection.

* Add a system `mcclim/test-util` for test-related utilities. The first utility `print-test-page` will be used to smoke-test output-only backends.


dkochmanski:
I've looked into commits and I found nothing wrong with them. Merging, thanks!

n.b I've noticed a pattern (not only in this pr) to replace defpackage and in-package with cl:defpackage and cl:in-package -- is it justified? I understand the intention: we don't want someone to substitute operators with their own symbols, but why?

scymtym:
> I've looked into commits and I found nothing wrong with them. Merging, thanks! 

Thanks.

> n.b I've noticed a pattern (not only in this pr) to replace  defpackage and in-package with cl:defpackage and cl:in-package -- is it  justified? I understand the intention: we don't want someone to  substitute operators with their own symbols, but why?

I haven't thought about this in a long time but here goes: In theory, the value of `*package*` is arbitrary when the file is read (for example if the file is loaded after the user or some other system changed `*package*` to `my-logistics-tools`). Just `(in-package …)` would then be read as `(my-logistics-tools::in-package …)` which could be about actual packages instead of CL packages.

Note that `defpackage` has the same potential issue. For `defpackage` one common way to guard against this is
```lisp
(in-package :cl-user)
(defpackage …)
```

In my view, the simplest solution is always writing `(cl:defpackage …)` and `(cl:in-package …)`.

One could argue (and Fare did) that ASDF should control these syntax-related aspects (also including the readtable, `*read-default-float-format*`, etc.) but as far as I know, it currently doesn't.

PR #853, opened 2019-10-02
slider-pane: check if the slider is properly armed for release event

### LINKED ISSUE #852: button release results in spurious slider event firing
In the latest HEAD (69371014), if I select a scroll bar, hold the mouse button down, move over to a slider and release the mouse button, I get an event firing on the slider, which isn't what I would expect. Sorry for not having a clean test case. If the broken behavior isn't obvious, I can try to generate one.

dkochmanski:
Fixes #852.


slyrus:
Looks good to me (and fixes the problem). Thanks!

PR #856, opened 2019-10-05
Small fixes for the listener and scigraph

scymtym:
These fixes silence around three warnings in SBCL.


dkochmanski:
Lgtm, thanks!

PR #857, opened 2019-10-05
Cache the bit depth for drawables.

### LINKED ISSUE #637: significant efficiency problem with draw-pattern*
There has been a 100-fold decrease in the performance `draw-pattern*` over the past 12 months - between versions of 10/2017 and 10/2018. 

The problem is obvious if you run the following code, which repeatedly draws a 2x2 pattern to an application pane:

```
(in-package :clim-user)

(defparameter +pattern+
  (make-pattern
    #2A((0 1)
        (1 0))
    (list +white+ +black+)))

(defun display-pattern (frame stream)
  (declare (ignore frame))
  (window-clear stream)
  (time
    (loop for x from 0 to 250 by 10
      do
      (loop for y from 0 to 250 by 10
        do
        (draw-pattern* stream +pattern+ x y)))))
 
(define-application-frame drawtest ()
  ()
  (:menu-bar t)
  (:panes
    (p1
      (make-pane 'application-pane
        :width 300 :height 300
        :display-function #'display-pattern)))
  (:layouts
    (default
      (spacing (:thickness 20) p1))))

(defun run-drawtest ()
  (run-frame-top-level (make-application-frame 'drawtest)))

(define-drawtest-command (quit :name t :menu t) ()
  (with-application-frame (frame) (frame-exit frame)))

(run-drawtest)
```

Here's a summary of the timing results:

```
SBCL 1.3.18, macOS
mcclim-20181018-git
clx-20181018-git
  11.217 seconds of real time
  8.339215 seconds of total run time (6.543405 user, 1.795810 system)
  11,073,952 bytes consed

SBCL 1.3.18, macOS
mcclim-20171023-git
clx-20171023-git
  0.078 seconds of real time
  0.062724 seconds of total run time (0.055204 user, 0.007520 system)
  13,019,632 bytes consed
```

i.e. the runtime has increased from less than 0.1 second to more than 10 seconds. In my application, this slow-down significantly degrades user experience.

I don't know what could have caused this. Changes in McCLIM or in CLX? The definition of `draw-pattern*` is certainly very different to what it used to be. I would very much appreciate a fix to get back to somewhere near the old performance.

kc-:
This fixes #637 inside McCLIM instead of relying on a cache inside CLX.

### DIFF
--- a/Backends/CLX/medium.lisp
+++ b/Backends/CLX/medium.lisp
@@ -406,10 +406,19 @@ translated, so they begin at different position than [0,0])."))
       (push #'(lambda () (xlib:free-pixmap mm)) ^cleanup)
       mm)))
 
+
+;;; The purpose of this is to reduce local network traffic for the case of many
+;;; calls to compute-rgb-image, for example when drawing a pattern.
+;;; For more details, see also: https://github.com/sharplispers/clx/pull/146
+(defun cached-drawable-depth (drawable)
+  (or (getf (xlib:drawable-plist drawable) :clim-cache)
+      (setf (getf (xlib:drawable-plist drawable) :clim-cache)
+            (xlib:drawable-depth drawable))))
+
 (defun compute-rgb-image (drawable image)
   (let* ((width (pattern-width image))
          (height (pattern-height image))
-         (depth (xlib:drawable-depth drawable))
+         (depth (cached-drawable-depth drawable))
          (idata (climi::pattern-array image)))
     (let* ((pm (xlib:create-pixmap :drawable drawable
                                    :width width


Backends/CLX/medium.lisp
```diff
@@ -406,10 +406,19 @@ translated, so they begin at different position than [0,0])."))
       (push #'(lambda () (xlib:free-pixmap mm)) ^cleanup)
       mm)))
 
+;;; For more details, see also: https://github.com/sharplispers/clx/pull/146
+(defun cached-drawable-depth (drawable)
+  "The purpose of this is to reduce local network traffic for the case of many
+calls to compute-rgb-image, for example when drawing a pattern."
```
dkochmanski: make this docstring a comment and rename the function to `ensure-drawable-depth`

Backends/CLX/medium.lisp
```diff
@@ -406,10 +406,19 @@ translated, so they begin at different position than [0,0])."))
...
+  (symbol-macrolet ((plist (xlib:drawable-plist drawable)))
```
dkochmanski: there is no real need for symbol-macrolet (form is short and used only twice). Please remove it.

kc-: Okay -- I'll note for the record though that in my opinion, it is much more readable with this macrolet.

Backends/CLX/medium.lisp
kc-: I moved the docstring. We discussed the naming on IRC.

PR #865, opened 2019-10-09
Refactor regions

dkochmanski:


dkochmanski:
Most important changes:
- all unbounded region methods are put in a small macro
- region-set methods are kept separately
- region-contains-position-p is implemented for protocol classes (using their protocols)
- missing region-difference "general" methods is introduced
- "general" specializations are made on "bounded-rectangle" (and not region)

dkochmanski:
I've also moved utilities to a separate file

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -1,43 +1,65 @@
...
+(declaim (inline segment-contains-point-p))
+(defun segment-contains-point-p (x1 y1 x2 y2 x y)
+  (and (coordinate-in x1 x x2)
+       (coordinate-in y1 y y2)
+       (line-contains-point-p x1 y1 x2 y2 x y)))
+
+;;; CLIM "native" coordinate system is left-handed (y grows down).
+;;; Angles are specified to grow in the counter-clockwise direction
+;;; (disregarding of the coordinate system), so we need to inverse the
+;;; y coordinate in a call to atan.
```
scymtym: need to *invert* the y coordinate in *the* call to atan.

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -1,43 +1,65 @@
 (in-package #:climi)
 
+;;; This file contains implementation of the predicate protocol. To
+;;; avoid circular dependencies and other issues we set a few rules:
+;;;
+;;; - don't use functions which methods are defined later in the file
```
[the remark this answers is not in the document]
dkochmanski: methods are sorted in groups: region-contains-position-p, region-intersects-region-p, region-contains-region-p and region-equals. this sentence means, that region-intersects-region-p should not call region-contains-region-p.

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -1,43 +1,65 @@
 (in-package #:climi)
 
+;;; This file contains implementation of the predicate protocol. To
```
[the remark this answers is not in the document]
dkochmanski: done

Core/clim-basic/geometry/coordinates.lisp
```diff
@@ -65,6 +74,15 @@
...
   (<= x y))
 
+(declaim (inline coordinate-in))
+(defun coordinate-in (c1 x c2)
+  (or (<= c1 x c2)
+      (>= c1 x c2)))
+
+(declaim (inline coordinate-in*))
+(defun coordinate-in* (c1 x c2)
+  (<= c1 x c2))
```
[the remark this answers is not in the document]
dkochmanski: done

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: done

dkochmanski:
I've pushed wip-refactor-regions branch, I don't think I have enough energy right now to execute things I have planned, so for now I'll settle with this PR regarding geometry module.

dkochmanski:
rebased against current master and resolved conflicts

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -210,6 +210,14 @@
 (defmethod region-equal ((a bounding-rectangle) (b bounding-rectangle))
```
scymtym: Type in commit message: "reigon-equal"

dkochmanski: fixed

Core/clim-basic/geometry/regions.lisp
```diff
@@ -68,8 +68,47 @@
...
+    +nowhere+ a))
+
+(defmethod region-contains-position-p ((self everywhere-region) x y)
+  (declare (ignore x y))
+  t)
+
+(defmethod region-contains-position-p ((self nowhere-region) x y)
+  (declare (ignore x y))
+  nil)
+
```
scymtym: This whole block should move a few lines down and so that it follows the definitions of `+everywhere+` and `+nowhere+`. Otherwise a compile-time warning may be signaled due to `+nowhere+` not yet being defined.

dkochmanski: fixed. also changed defvar with defconstant. in clim-lisp-patch it is replaced with defvar, but it should be replaced with alexandria's define-constant, but that's another story.

Core/clim-basic/geometry/regions.lisp
```diff
@@ -465,12 +465,30 @@
...
+     (let ((cx (coordinate cx))
+           (cy (coordinate cy))
+           (rdx1 (coordinate rdx1))
+           (rdy1 (coordinate rdy1))
+           (rdx2 (coordinate rdx2))
+           (rdy2 (coordinate rdy2)))
+       (make-3-point-transformation* 0 0 1 0 0 1
+                                     cx cy
+                                     (+ cx rdx1) (+ cy rdy1)
+                                     (+ cx rdx2) (+ cy rdy2))))))
```
scymtym: Why is the method needed? I can't find any subclasses of `ellipse` or `elliptical-arc` besides `standard-ellipse` and `standard-elliptical-arc` respectively and those two don't need the method.

Core/clim-basic/geometry/transforms.lisp
```diff
@@ -702,7 +702,7 @@ real numbers, and default to 0."
 (defun atan* (x y)
   ;; atan like we need it.
   ;; For us, phi=0 along the x axis and angles are always between 0 and 2pi.
-  (let ((r (atan y x)))
+  (let ((r (atan (float y pi) x)))
```
scymtym: Why is this needed?

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -83,21 +134,80 @@
...
+  )
+
 

 ;;   REGION-CONTAINS-REGION-P region1 region2
 ;;
-;;        Returns t if all points in the region region2 are members of the
-;;        region region1; otherwise, it returns nil.
+;;        Returns T if all points in the region are members of the
+;;        region B; otherwise, it returns nil.
```
scymtym: I think the previous version of the comment made more sense since it referred to the parameters.

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -83,21 +134,80 @@
...
+
+(defmethod region-intersects-region-p ((a standard-rectangle) (b standard-rectangle))
+  (declare (ignorable a b))
+  ;; for rectangles, the bounding rectangle test is correct, so if we
+  ;; wind up here, we just can return T.
+  t
+  ;;(multiple-value-bind (x1 y1 x2 y2) (rectangle-edges* a)
+  ;;  (multiple-value-bind (u1 v1 u2 v2) (rectangle-edges* b)
+  ;;    (and (<= u1 x2) (<= x1 u2)
+  ;;         (<= v1 y2) (<= y1 v2))))
```
scymtym: Remove this.

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -1,43 +1,62 @@
...
-  (block nil
-    (map-over-polygon-segments
-     (lambda (x1 y1 x2 y2)
-       (when (line-contains-point-p* x1 y1 x2 y2 x y)
-         (return t)))
-     self)
-    nil))
+

+(defmethod region-contains-position-p ((self everywhere-region) x y)
```
scymtym: Maybe call the parameter `region` instead of `self` to match the generic function?

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1204 @@
...
+        (coordinate= (* (- x2 x1) (- y3 y2))
+                     (* (- x3 x2) (- y2 y1)))))))
+
+

+;;; line utilities
+(declaim (inline line-contains-point-p))
+(defun line-contains-point-p (x1 y1 x2 y2 px py)
+  (= (* (- py y1) (- x2 x1))
+               (* (- px x1) (- y2 y1))))
```
scymtym: Indentation. But also, why does this not use `coordinate=`?

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -1,43 +1,62 @@
...
-(defmethod region-contains-position-p ((self standard-line) x y)
-  (multiple-value-bind (x1 y1) (line-start-point* self)
-    (multiple-value-bind (x2 y2) (line-end-point* self)
-      (line-contains-point-p* x1 y1 x2 y2 x y))))
-
-(defmethod region-contains-position-p ((self standard-rectangle) x y)
-  (with-standard-rectangle (x1 y1 x2 y2)
-      self
-    (and (<= x1 (coordinate x) x2)
-         (<= y1 (coordinate y) y2))))
```
scymtym: Maybe keep this specialization (in addition to the new method on `rectangle`) since it is potentially much more efficient.

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -83,21 +134,80 @@
...
+                              (+ 1 (square polar-dx) (square polar-dy))))))))
+
+

+;;   REGION-INTERSECTS-REGION-P region1 region2
+;;
+;;        Returns nil if region-intersection of the two regions region1 and
+;;        region2 would be +nowhere+; otherwise, it returns t.
+;;
+;;        aka region1 und region2 sind nicht disjunkt  aka AB /= 0
```
scymtym: Translation: "that means region1 and region2 are not disjoint, that means A ∩ B ≠ ∅"

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -83,21 +134,80 @@
...
+    (and (coordinate-between* x1 x x2)
+         (coordinate-between* y1 y y2))))
+
+(defmethod region-contains-position-p ((self ellipse) x y)
+  (nest
+   (multiple-value-bind (cx cy) (ellipse-center-point* self))
+   (let ((alpha (or (ellipse-start-angle self) 0))
+         (omega (or (ellipse-end-angle self) (* 2 pi)))
+         (polar->screen (polar->screen self)))
+     (and (angle-contains-point-p alpha omega (- x cx) (- y cy))
```
scymtym: This does lots of unnecessary work if start and end angle are null. Conversely, if the angle test fails, the `polar->screen` call is wasted. I suggest making local functions `angle-contains-p` for the angle test and `position-contains-p` for the position test and organizing the body like
```lisp
(and (or (not start-angle)
         (angle-contains-p start-angle (or end-angle 2pi)))
     (position-contains-p))
```

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -66,12 +77,52 @@
...
+                ;; but rotated elliptoids are naively rendered in clx.
+                (multiple-value-bind (polar-dx polar-dy)
+                    (untransform-distance polar->screen 1 1))
+                (coordinate-between*
+                 (- 1 (+ (square polar-dx) (square polar-dy)))
+                 (+ (square polar-x) (square polar-y))
+                 (+ 1 (+ (square polar-dx) (square polar-dy)))))))))
+
+(defmethod region-contains-position-p ((region polygon) x y)
+  (and (region-contains-position-p (bounding-rectangle region) x y)
```
scymtym: Would it make sense to do this generally in an `:around` method for 2-dimensional regions like for `region-intersects-region-p`?

scymtym: Maybe specialized on `area` or the cached-bounding-rectangle thing.

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -66,12 +77,52 @@
...
+          (nest (multiple-value-bind (polar-x polar-y)
+                    (untransform-position polar->screen x y))
+                ;; FIXME we don't need to factor the additonal epsilon
+                ;; but rotated elliptoids are naively rendered in clx.
+                (multiple-value-bind (polar-dx polar-dy)
+                    (untransform-distance polar->screen 1 1))
+                (coordinate-between*
+                 (- 1 (+ (square polar-dx) (square polar-dy)))
+                 (+ (square polar-x) (square polar-y))
+                 (+ 1 (+ (square polar-dx) (square polar-dy)))))))))
```
scymtym: Don't compute `(+ (square polar-dx) (square polar-dy))` twice.

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -66,12 +77,52 @@
...
-       (not (region-contains-position-p (standard-region-difference-b self) x y))))
+(defmethod region-contains-position-p ((region standard-region-difference) x y)
+  (let ((region-a (standard-region-difference-a region))
+        (region-b (standard-region-difference-b region)))
+    (and (region-contains-position-p region-a x y)
+         (not (region-contains-position-p region-b x y)))))
+
+(defmethod region-contains-position-p ((self point) x y)
+  (and (coordinate= (point-x self) x)
+       (coordinate= (point-y self) y)))
```
scymtym: Please use `point-position` -- `point-x` **and** `point-y` call it anyway which is crazy expensive.

Core/clim-basic/geometry/regions.lisp
dkochmanski: quoting a self-imposed rule:
```
;;; - when possible provide a method for a protocol class, and when
;;;   beneficial provide (also) a method for a standard class
```

in principle this method works on any object which implements the ellipse protocol (only protocol functions are invoked). Thanks to that our definitions may be specialized on ellipse/elliptical-arc and it will work for any ellipse (not only standard-ellipse/elliptical-arc).

Core/clim-basic/geometry/transforms.lisp
dkochmanski: because we compare the result with pi (i.e in correct-angle and angle-contains-angle-p); I recall I had a very tangible problem with that but I don't remember what was it now.

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: this method is removed in 65c66671 (and the argument name is carried from the previous implementation)

Core/clim-basic/geometry/region-utilities.lisp
dkochmanski: I don't know/remember, indentation shows that it was indeed coordinate= before. I take a guess that that must have been a typo. fixed

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: fixed.

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: fixed (I've left `aka` since it doesn't need translation)

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: fixed (I've used slightly different code). if start-angle is not null then end-angle is not null neither (specified behavior) so I've skipped (or end-angle 2pi).

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: it does make sense but not as things are now. I have plans to cache bounding rectangle for all regions but as things are now it would be very inefficient i.e for ellipses.

region-intersects-region-p around method is there to provide a negative test, because next method (region-intersection) is not defined for all pairs of geometrical objects. In other words it is a hack which makes us fail less often with no-applicable-method exception or with incorrect result from blind creation of standard-region-intersection object when we do not know what to do.

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: fixed. also fixed to check for null start-angle (like with ellipse). clearly forgot about that part.

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: fixed

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: fixed

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: fixed

dkochmanski:
addressed all concerns and force-pushed.

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1225 @@
...
+
+;;; A band list is represented by
+
+;;;  ((x_0 . a_0) (x_1 . a_1) ... (x_n . nil))
+
+;;; The a_i are the relevant interval sums for x in [x_i, x_(i+1)].
+
+;;; The empty band could have been representated as
+;;;  ((x . nil))  x arbitrary
+;;; But to get a cononic representation, I'll choose simply NIL.
```
nsrahmad: cononic -> canonic

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1225 @@
...
+
+(defun %ellipse-position->angle (ellipse x y)
+  (multiple-value-bind (xc yc) (ellipse-center-point* ellipse)
+    ;; remember, that y-axis is reverted
+    (coordinate (atan* (- x xc) (- (- y yc))))))
+
+;;; -- Ellipse simplified representation ---------------------------------------
+
+(defun ellipse-coefficients (ell)
+  ;; Returns the coefficients of the equation specifing the ellipse as in
```
nsrahmad: specifing -> specifying

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1225 @@
...
+           (y (sin base-angle)))
+      (multiple-value-bind (mxx mxy myx myy tx ty) (get-transformation tr)
+        (values (+ (* mxx x) (* mxy y) tx)
+                (+ (* myx x) (* myy y) ty))))))
+
+(defun ellipse-simplified-representation (el)
+  ;; returns H (horizontal radius), V (vertical radius) and rotation angle in
+  ;; screen coordinates. `ellipse-normal-radii*' returns vectors with correct
+  ;; direction, but radius length is shorter than in reality (verified with
+  ;; experimentation, not analitically), so we compute radius from phi.
```
nsrahmad: verified experimentally, not "analitically"-> analytically

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1225 @@
...
+                 (expt (* h v) 2)))
+           (dc (sqrt (- (* b b) (* 4 a c))))
+           (y1 (/ (- (- b) dc)
+                  (* 2 a)))
+           (y2 (/ (+ (- b) dc)
+                  (* 2 a))))
+      (values (+ cx x) (+ cy y1) (+ cx x) (+ cy y2)))))
+
+(defun intersection-line/ellipse (el lx1 ly1 lx2 ly2)
+  "Returns coordinates where ellipse intersects with arbitral line (except vertical)."
```
nsrahmad: arbitral line -> arbitrary line

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1225 @@
...
+      (reparameterize-ellipse radius1-dx radius1-dy radius2-dx radius2-dy)
+    (ellipse-point (- lambda0 theta) center-x center-y a b theta)))
+
+(defun ellipse-derivative (eta a b theta)
+  "Given an ellipse having two radii of length A and B, with angle
+THETA between the radii from the center of the ellipse, returns two
+values, the x and y coordinates of the derivative of the parametric
+curve of ellipse at the parametric angle eta. Note that this eta is
+not the angle ANGLE, as in ellipse-derivative*, but rather is computed
+parametricly from theta. See the paper from Luc Maisonobe for
```
nsrahmad: parametricly -> parametrically

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -1,77 +1,73 @@
 (in-package #:climi)
 
+;;; This file contains implementation of the region predicate protocol
+;;; for several region classes. A few rules to follow:
```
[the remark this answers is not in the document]
dkochmanski: "a few rules" is correct, it has slightly different meaning than "few rules" [1].

https://www.diffen.com/difference/A_Few_vs_Few

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1225 @@
...
+                    (isum-op as bs boole-op in-a in-b x0))))))))
+
+;;; -- Bands -----------------------------------------------------------------
+
+
+;;; A band list is represented by
+
+;;;  ((x_0 . a_0) (x_1 . a_1) ... (x_n . nil))
+
+;;; The a_i are the relevant interval sums for x in [x_i, x_(i+1)].
```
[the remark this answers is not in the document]
dkochmanski: not `a_i are relevant interval sums`? by a_i I mean a[0], a[1] etc..

dkochmanski:
thank you, I've made a single cosmetic commit to fix typos (I've incorporated all remarks except commented two).

nsrahmad:
Looks good. I have nothing else to add and will leave the merging to @scymtym .

dkochmanski:
thank you. I've found a regression inroduced in `geometry: separate region-set...`, looking at it now.

dkochmanski:
ah, that was a rebase conflict fix issue (region-composition was not loaded), fixed.

dkochmanski:
is there anything elseto change in this PR?

Core/clim-basic/geometry/region-predicates.lisp
scymtym: OK.

Core/clim-basic/geometry/region-predicates.lisp
scymtym: OK, I guess we can improve this later.

Core/clim-basic/geometry/transforms.lisp
scymtym: Please add a FIXME to either remove the `float` call or document why it is needed, then.

Core/clim-basic/geometry/region-predicates.lisp
```diff
@@ -66,12 +77,55 @@
...
-(defmethod region-contains-position-p ((self standard-region-difference) x y)
-  (and (region-contains-position-p (standard-region-difference-a self) x y)
-       (not (region-contains-position-p (standard-region-difference-b self) x y))))
+(defmethod region-contains-position-p ((region standard-region-difference) x y)
+  (let ((region-a (standard-region-difference-a region))
+        (region-b (standard-region-difference-b region)))
+    (and (region-contains-position-p region-a x y)
+         (not (region-contains-position-p region-b x y)))))
+
+(defmethod region-contains-position-p ((self point) x y)
```
scymtym: Please use `region` instead of `self` for the newly added methods.

Core/clim-basic/geometry/regions.lisp
```diff
@@ -68,11 +68,50 @@
...
+  ;; +nowhere+ (that would complicate the geometry module). If we
+  ;; decide otherwise don't use standard-region-difference because it
+  ;; is a subclass of a bounded-rectangle so it can't represent
+  ;; unbounded region.  -- jd 2019-09-10
+  (def-method region-difference
+    +nowhere+ a (error "Unsupported unbounded region operation.")
+    a a a
+    +nowhere+ a))
+
+(defmethod region-contains-position-p ((self everywhere-region) x y)
```
scymtym: Please call the parameter `region`.

Core/clim-basic/geometry/regions.lisp
```diff
@@ -68,11 +68,50 @@
...
+  (declare (ignore x y))
+  t)
+
+(defmethod region-contains-position-p ((self nowhere-region) x y)
+  (declare (ignore x y))
+  nil)
 
-(defvar +everywhere+ (make-instance 'everywhere-region))
-(defvar +nowhere+ (make-instance 'nowhere-region))
+;;; coordinate is defined in coordinates.lisp
```
scymtym: I don't think this comment is needed.

Core/clim-basic/geometry/region-composition.lisp
```diff
@@ -77,7 +63,7 @@
     :regions (append (standard-region-set-regions a)
                      (standard-region-set-regions b))))
 
-(defmethod region-union ((a region) (b region))
+(defmethod region-union ((a bounding-rectangle) (b bounding-rectangle))
```
scymtym: The commit message says

>"General" methods specialized on region have been changed to specialize on a bounding-rectangle to avoid ambiguity.

Can you give an example? The method bodies don't seem to require the region to be a `bounding-rectangle`.

Core/clim-basic/geometry/region-set-composition.lisp
```diff
@@ -0,0 +1,192 @@
...
+(defun rectangle-set->polygon-union (rs)
+  (let ((res nil))
+    (map-over-region-set-regions (lambda (r) (push r res)) rs)
+    (make-instance 'standard-region-union :regions res)))
+
+

+;; REGION-UNION
+
+#+ (or) ;; defined in region-composition.lisp
```
scymtym: Can this be removed? Since the method has no region set specializer, I wouldn't expect it to be defined in this file.

Core/clim-basic/geometry/region-set-composition.lisp
```diff
@@ -0,0 +1,192 @@
...
+  (make-instance 'standard-region-union :regions (list a b)))
+
+(defmethod region-union ((a bounding-rectangle) (b standard-region-difference))
+  (make-instance 'standard-region-union :regions (list a b)))
+
+

+;;; REGION-INTERSECTION
+
+#+ (or) ;; defined in region-composition.lisp
```
scymtym: Can this be removed? Since the method has no region set specializer, I wouldn't expect it to be defined in this file.

Core/clim-basic/clim-basic.asd
```diff
@@ -24,7 +24,9 @@
                  (:file "region-bounding-rectangles")
                  (:file "region-transformations")
                  (:file "region-predicates")
-                 (:file "region-composition")))
+                 (:file "region-composition")
+                 (:file "region-set-predicates")
+                 (:file "region-set-composition")))
```
scymtym: Maybe also extract the region classes and associated utilities (`map-over-bands` and friends) into `region-sets.lisp`? `regions.lisp` is almost 2000 lines and would benefit from it, I think.

Core/clim-basic/geometry/region-set-predicates.lisp
```diff
@@ -0,0 +1,42 @@
...
+  (not (region-equal +nowhere+ (region-intersection a b))))
+
+#+ (or) ;; defined in region-predicates.lisp
+(defmethod region-contains-region-p ((a bounding-rectangle) (b bounding-rectangle))
+  (or (eq a b)
+      (region-equal +nowhere+ (region-difference b a))))
+
+#+ (or) ;; defined in region-predicates.lisp
+(defmethod region-equal ((a bounding-rectangle) (b bounding-rectangle))
+  (region-equal +nowhere+ (region-exclusive-or a b)))
```
scymtym: Can this be removed? Since the method has no region set specializer, I wouldn't expect it to be defined in this file.

Core/clim-basic/clim-basic.asd
```diff
@@ -24,7 +24,9 @@
                  (:file "region-bounding-rectangles")
                  (:file "region-transformations")
                  (:file "region-predicates")
-                 (:file "region-composition")))
+                 (:file "region-composition")
```
scymtym: The commit message says

> add missing "general" methods for region-difference (clearly missing)

Please list the methods in the commit message since it is really hard to tell from the diff.

Tests/geometry/regions.lisp
```diff
@@ -300,18 +300,19 @@
          (some-el (list el1 el2 el3 el4))
```
scymtym: > Tests for belinging with a difference of 1 will likely fail because we factor an additional epsilon for ellipse tests. Without that clx rendering of rotated ellipses doesn't work correctly.

"belinging" seems like a typo.

But also, please add a FIXME and comment like this to the test in question.

Core/clim-basic/geometry/region-composition.lisp
```diff
@@ -239,25 +239,13 @@
...
-  (multiple-value-bind (x1 y1) (line-start-point* a)
-    (multiple-value-bind (x2 y2) (line-end-point* a)
-      (schnitt-line/polygon x1 y1 x2 y2 b))))
-
-(defmethod region-intersection ((b standard-rectangle) (a standard-line))
-  (multiple-value-bind (x1 y1) (line-start-point* a)
-    (multiple-value-bind (x2 y2) (line-end-point* a)
-      (schnitt-line/polygon x1 y1 x2 y2 b))))
+(defmethod region-intersection ((a line) (b polygon))
+  (region-intersection b a))
```
scymtym: I like this cleanup, but the second method adds a generic function call. Would a local macro for generating the body (one as is and the second with `a` and `b` swapped) make sense?

Core/clim-basic/geometry/region-composition.lisp
```diff
@@ -325,15 +298,10 @@
...
-  (multiple-value-bind (x1 y1) (line-start-point* a)
-    (multiple-value-bind (x2 y2) (line-end-point* a)
-      (differenz-line/polygon x1 y1 x2 y2 b))))
-
-(defmethod region-difference ((a standard-line) (b standard-rectangle))
+(defmethod region-difference ((a line) (b polygon))
   (multiple-value-bind (x1 y1) (line-start-point* a)
     (multiple-value-bind (x2 y2) (line-end-point* a)
-      (differenz-line/polygon x1 y1 x2 y2 b))))
+      (difference-segment/polygon x1 y1 x2 y2 b))))
```
scymtym: This would better fit into the "remove duplicated methods" commit.

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1204 @@
...
+          (> angle (* pi 2)))
+      (mod angle (* pi 2))
+      angle))
+
+(defun find-angle* (x1 y1 x2 y2)
+  "Returns the angle between two vectors described by x1, y1 and x2,
+y2."
+  (let ((theta (- (phase (complex y2 x2))
+                  (phase (complex y1 x1)))))
+    (normalize-angle theta)))
```
scymtym: These should probably be in the "point and angle utilities" section.

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1204 @@
...
+  ;; For us, phi=0 along the x axis and angles are always between 0 and 2pi.
+  (let ((r (atan (float y pi) x)))
+    (if (< r 0) (+ r (* 2 pi)) r)))
+
+(declaim (inline angle-contains-angle-p))
+(defun angle-contains-angle-p (alpha omega delta)
+  (if (< alpha omega)
+      (coordinate-between* alpha delta omega)
+      (or (coordinate<= alpha delta)
+          (coordinate<= delta omega))))
```
scymtym: I found the old name and lambda list ` %angle-between-p (angle start-angle end-angle)` much clearer.

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -0,0 +1,1204 @@
...
+                                       (setf res (nconc sol res))
+                                       (return))))
+                                 (setf x1 x-start)
+                                 (setq n 0))))))
+                 (setf x (min 1d0 (max -1d0 x1)))        ;Is this allowed?
+                 (incf n)))
+             res)))))
+
+(defun horner-schema (polynom x)
+  ;; Evaluates the polynomial `polynom' by means of horner's method at the
```
scymtym: I would call the parameter `polynomial` (`polynom` is German) and change the comment to `;; Evaluate POLYNOMIAL …`.

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -338,57 +329,87 @@
...
+  ;; through the segment (x1 y1 x2 y2) and the polygon. Each of these
+  ;; points may be potentially a vertice of a segment which is the
+  ;; intersection of the polygon and the line. Function argument is a
+  ;; parameter indicating where on the segment S the intersection
+  ;; point is positioned (where 0d0 is [x1,y1] and 1d0 is [x2,y2]). If
+  ;; parameter is not (<= 0d0 param 1d0) position falls outside the
+  ;; segment (but is still on the line). -- jd 2019-10-10
+  (flet ((call-fun (point)
+           (let ((px (point-x point))
+                 (py (point-y point)))
```
scymtym: * Use `multiple-value-bind` with `point-position` to get both at once (and save three generic function calls).
* Declare `call-fun` inline. 
* Use `(let ((fun (alexandria:ensure-function fun))) BODY)` so `fun` is coerced only once and can then be assumed to be a function (rather than a function designator).

Core/clim-basic/geometry/region-utilities.lisp
```diff
@@ -338,57 +329,87 @@
...
+          ;; The line goes directly through PN (handled later)
+          ((line-contains-point-p x1 y1 x2 y2 (point-x pn) (point-y pn))
+           nil)
+          ;; The line doesn't go throuh PO nor PN points. It may cross
+          ;; a segment PO-PN or fall outside.
+          (t
+           (let* ((x3 (point-x po))
+                  (y3 (point-y po))
+                  (x4 (point-x pn))
+                  (y4 (point-y pn))
```
scymtym: Use `point-position`.

Core/clim-basic/geometry/bounding-rectangle.lisp
```diff
@@ -31,7 +31,7 @@
 ;;; 4.1.2 Bounding Rectangle Convenience Functions
 
 (defmacro with-bounding-rectangle* ((min-x min-y max-x max-y) region &body body)
-  ;; What is the purpose of this macro; IHMO m.-v.-b. looks as nice as with-b.-.r. .
+  ;; What is the purpose of this macro; IMHO m.-v.-b. looks as nice as with-b.-.r. .
```
scymtym: The macro is in the spec, so the comment can be deleted, I think.

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: do you want me to change also already existing methods? also, should I change transforms too to replace self with transformation?

Core/clim-basic/geometry/region-composition.lisp
dkochmanski: all unbounded regions are handled in regions.lisp (see macro def-method). If we specialize "just" on region, then the right method made specially for unbound regions may be replaced by a another one. For instance:

```
(DEFMETHOD REGION-DIFFERENCE ((A REGION) (B EVERYWHERE-REGION))
     (DECLARE (IGNORABLE A B))
     +NOWHERE+)

(defmethod region-difference ((a rectangle) (b region))
  (make-instance 'standard-region-difference a b))
```

problem does not occur if we specialize all method on bounded-rectangles. The point is that for all unbounded regions have methods defined and we do not want to replace them.

Core/clim-basic/geometry/region-set-composition.lisp
dkochmanski: fixed

Core/clim-basic/geometry/region-set-composition.lisp
dkochmanski: fixed

Core/clim-basic/geometry/region-set-predicates.lisp
dkochmanski: fixedfixed

Tests/geometry/regions.lisp
dkochmanski: fixed

Core/clim-basic/geometry/region-composition.lisp
dkochmanski: it is certainly possible, but I think that it will obfuscate code for little gain (i.e it is unlikely it will be a bottleneck in practice).

Core/clim-basic/geometry/region-composition.lisp
dkochmanski: how so? we do not remove here any duplicated method, we simply rename the utility

Core/clim-basic/geometry/region-utilities.lisp
dkochmanski: fixed

Core/clim-basic/geometry/region-utilities.lisp
dkochmanski: this rename is more consistent with angle-contains-point-p so I'd rather keep it this way (I find it better despite coming up first with %angle-between-p)

Core/clim-basic/geometry/region-utilities.lisp
dkochmanski: fixed

Core/clim-basic/geometry/region-utilities.lisp
dkochmanski: fixed (here and in other places too)

Core/clim-basic/geometry/bounding-rectangle.lisp
dkochmanski: deleted

Core/clim-basic/clim-basic.asd
dkochmanski: fixed the commit message with a brief explanation why it is "clearly" missing.

Core/clim-basic/clim-basic.asd
dkochmanski: map-over-bands is already moved to regin-utilities.lisp (and regions has 774 lines)

Core/clim-basic/geometry/transforms.lisp
dkochmanski: commented - please see last commit

(and for amusement: https://irclog.tymoon.eu/freenode/%23clim?around=1574156524#1574156524)

dkochmanski: done (please see last commit)

dkochmanski:
I think I've answered all remarks. When I was unsure I've left a comment and kept the discussion open.

Core/clim-basic/geometry/region-predicates.lisp
scymtym: I would change it in any code I "touched".

Regarding the question whether the change makes sense: I think `self` is bad because

1. It is likely a result of the "classes own methods" line of thought
2. In the generic function definition the parameter is called `region` which accurately describes the role of the parameter in the generic function and protocol. I see no reason to deviate from that naming in methods.

Core/clim-basic/geometry/regions.lisp
scymtym: See previous comment.

Core/clim-basic/geometry/region-composition.lisp
scymtym: I see. Could you add this explanation to `region-composition.lisp` as a comment and the commit message?

Core/clim-basic/clim-basic.asd
scymtym: OK.

Core/clim-basic/geometry/region-composition.lisp
scymtym: We seem to have different tolerances for introducing performance regressions when cleaning up. My experience so far has been that we have to re-optimize sooner or later after introducing performance regressions to facilitate cleanups (e.g. `clx-fb`). I have also run into performance issues with the geometry module several times already, so I'm more worried than I would normally be.

In this case, reducing four methods to two still seems like a good simplification (with or without macrology).

Core/clim-basic/geometry/region-composition.lisp
scymtym: Hm, I think I did something wrong in the review interface.

This was supposed to reference "- remove duplicated methods calling difference-segment/polygon" in the "geometry: move math utilities to a separate file" commit.

Core/clim-basic/geometry/region-utilities.lisp
scymtym: Sorry, but I think it is really unclear. Firstly because it is not clear how an angle can "contain" another angle, secondly because `alpha omega delta` do not indicate which role the respective parameter plays.

I can understand that you would like something using the `X-contains-Y-p` shape. How about circle-segment contains angle or arc contains angle with parameters `angle` and e.g. `arc-start-angle` and `arc-end-angle`?

Core/clim-basic/geometry/region-composition.lisp
dkochmanski: please look at the last commit

Core/clim-basic/geometry/region-composition.lisp
dkochmanski: I'm still unsure what should be done.

Core/clim-basic/geometry/region-utilities.lisp
dkochmanski: pushed as a separate commit (last one)

Core/clim-basic/geometry/region-composition.lisp
dkochmanski: I've added it to file regions.lisp (and to commit message), regions lisp is more appropriate place because that's where this requirement comes from and it spans not only region-composition, but also region-predicates and region-set files.

Core/clim-basic/geometry/regions.lisp
dkochmanski: done

Core/clim-basic/geometry/region-predicates.lisp
dkochmanski: done

Issue #605, opened 2018-11-11
Creating empty labels


thrashdin:
For some reason creating label without :text and running the app reboots my user session. I was trying to set the text of the label, and discovered this. Funny thing is that in case some :text "text" is passed, it also reboots my user session, but only if I launch it once, then close and then launch it again.
    ` (screen :label :text "input")`

Also https://stackoverflow.com/questions/53224545/mcclim-text-labels-access

I have Debian, everything as latest as it can be, and i weekly dist-upgrade. If some update will fix the problem, I'll add it here.

dkochmanski:
Hey, we already had this issue, this should be fixed in clx (to not send no-op requests with empty strings). Curiously enough some x11 servers crash and some don't.

Sokolov57:
Is there any workaround then? Smith to replace labels without looking odd, and with same or similar functionality.

nsrahmad:
>Is there any workaround then? Smth to replace labels without looking odd, and with same or similar functionality.

Assuming the functionality mentioned here is ability to display text that could change. Looking around in `clim-examples`, I found `clim-fig` uses `text-field` for this purpose. Note the 
status at the bottom:
![clim-fig-text-field-as-status](https://user-images.githubusercontent.com/10996250/66548748-7c385d80-eb5f-11e9-943b-a8953e029a9a.png)
That does look odd. A label would be better for that. Unfortunately, `label-pane` in CLIM is not specified to have a function that changes the label. However in McCLIM (after #866 is merged), `clim-extensions:label-pane-label` can be used to get a label and `(setf (clime:label-pane-label))` to change the label.

`clim-fig` with a label:
![clim-fig-after](https://user-images.githubusercontent.com/10996250/66549925-008be000-eb62-11e9-898e-7a666e98516a.png)

PR #866, opened 2019-10-10
Label pane

nsrahmad:
See #605 for context.

An example to test:
```lisp
(eval-when (:compile-toplevel :load-toplevel :execute)
  (unless (find-package "CLIM")
    (ql:quickload 'mcclim)))

(in-package :clim-user)

(define-application-frame label-test ()
  ()
  (:panes
   (label :label-pane
	  :label "A Label")
   (int :interactor :width 600))
  (:layouts
   (default (vertically () label int))))

(define-label-test-command (com-change-label :name t :menu t)
    ((new-label string))
  (setf (clime:label-pane-label (find-pane-named *application-frame* 'label))
	new-label))

(defun run-label-test ()
  (run-frame-top-level (make-application-frame 'label-test)))
```
A sample run looks like:

anim

### DIFF
--- a/Core/clim-core/panes.lisp
+++ b/Core/clim-core/panes.lisp
@@ -2302,7 +2302,7 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
 (defclass label-pane (multiple-child-composite-pane)
   ((label :type string
           :initarg :label
-          :accessor label-pane-label
+          :accessor clime:label-pane-label
           :initform "")
    (alignment :type (member :bottom :top)
               :initform :top
@@ -2313,6 +2313,9 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
    :align-y    :center
    :text-style (make-text-style :sans-serif nil nil))
   (:documentation ""))
+
+(defmethod (setf clime:label-pane-label) :after (new-value (pane label-pane))
+  (repaint-sheet pane (sheet-region pane)))
 
 (defmacro labelling ((&rest options) &body contents)
   `(make-pane 'label-pane ,@options :contents (list ,@contents)))
@@ -2342,7 +2345,7 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
 
 (defmethod compose-space ((pane label-pane) &key width height)
   (declare (ignore width height))
-  (let* ((w (text-size pane (label-pane-label pane)))
+  (let* ((w (text-size pane (clime:label-pane-label pane)))
          (a (text-style-ascent (pane-text-style pane) pane))
          (d (text-style-descent (pane-text-style pane) pane))
          (m0 2)
@@ -2393,7 +2396,7 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
   (let ((m0 2)
         (a (text-style-ascent (pane-text-style pane) pane))
         (d (text-style-descent (pane-text-style pane) pane))
-        (tw (text-size pane (label-pane-label pane))))
+        (tw (text-size pane (clime:label-pane-label pane))))
     (with-bounding-rectangle* (x1 y1 x2 y2) (sheet-region pane)
       (multiple-value-bind (iright itop ileft ibottom
                                    bright btop bleft bbottom)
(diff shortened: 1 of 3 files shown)


Examples/clim-fig.lisp
```diff
@@ -427,7 +427,8 @@
       (T
        (erase-output-record latest-undo-entry *standard-output*)
        (push latest-undo-entry (clim-fig-redo-list *application-frame*))
-       (activate-gadget (find-pane-named *application-frame* 'clear))))
+       (activate-gadget (find-pane-named *application-frame* 'clear))
+       (activate-gadget (find-pane-named *application-frame* 'redo))))
```
scymtym: I have changes that synchronize the active/inactive state of these buttons with the corresponding commands (including menus). Should I submit those as a pull request?

nsrahmad: Sure, go ahead.

scymtym:
Using a label instead of a text field in clim-fig is also a significant performance improvement.

Core/clim-core/panes.lisp
```diff
@@ -2314,6 +2314,12 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
    :text-style (make-text-style :sans-serif nil nil))
   (:documentation ""))
 
+(defmethod (setf clime:label-pane-label) :after (new-value (pane label-pane))
+  (let ((region (sheet-region pane)))
+    (multiple-value-bind (x1 y1 x2 y2) (rectangle-edges* region)
+      (draw-rectangle* pane x1 y1 x2 y2 :ink (pane-background pane)))
+    (repaint-sheet pane region)))
```
scymtym: I would prefer just `(repaint-sheet pane (sheet-region pane))` here, but adding `(draw-rectangle* pane x1 (- ty a) x2 (+ ty d) :ink (pane-background pane))` before the `draw-text*` call in `handle-repaint`. That way, the label will always be drawn onto a freshly cleared background, no matter what caused the repaint. E.g. repainting the pane without changing the label would no longer lead to the label text appearing almost bold caused by overdrawing the anti-aliased glyph. This effect is visible in the "A" part of the label in the following screenshot:

screenshot

dkochmanski:
Note that we have "always-repaint-background" mixin

On October 12, 2019 5:27:53 PM GMT+02:00, Jan Moringen <notifications@github.com> wrote:
>scymtym requested changes on this pull request.
>
>I would like to generalize the repainting improvement for `label-pane`.
>
>> @@ -2314,6 +2314,12 @@ SCROLLER-PANE appear on the ergonomic left
>hand side, or leave set to
>    :text-style (make-text-style :sans-serif nil nil))
>   (:documentation ""))
> 
>+(defmethod (setf clime:label-pane-label) :after (new-value (pane
>label-pane))
>+  (let ((region (sheet-region pane)))
>+    (multiple-value-bind (x1 y1 x2 y2) (rectangle-edges* region)
>+      (draw-rectangle* pane x1 y1 x2 y2 :ink (pane-background pane)))
>+    (repaint-sheet pane region)))
>
>I would prefer just `(repaint-sheet pane (sheet-region pane))` here,
>but adding `(draw-rectangle* pane x1 (- ty a) x2 (+ ty d) :ink
>(pane-background pane))` before the `draw-text*` call in
>`handle-repaint`. That way, the label will always be drawn onto a
>freshly cleared background, no matter what caused the repaint. E.g.
>repainting the pane without changing the label would no longer lead to
>the label text appearing almost bold caused by overdrawing the
>anti-aliased glyph. This effect is visible in the "A" part of the label
>in the following screenshot:
>
>screenshot

-- Wysłane za pomocą K-9 Mail.

scymtym:
> Note that we have "always-repaint-background" mixin

That repaints the whole sheet region (and is currently only used in one place for some reason). The suggested change only repaints the region that could potentially be covered by the label.

Core/clim-core/panes.lisp
nsrahmad: Done. This is indeed better solution.

scymtym:
Thank you.

scymtym:
> Sure, go ahead.

#869

PR #868, opened 2019-10-11
[wip] drag-and-drop: improve user feedback

dkochmanski:
This is important if we have multiple translators with different modifiers.


dkochmanski:
where do we stand with this?

scymtym:
> where do we stand with this?

I believe I was supposed to try adding a clause for key presses in the `pointer-tracking` form of presentation drag and drop which I didn't manage to do yet.

scymtym:
> I believe I was supposed to try adding a clause for key presses in the `pointer-tracking` form of presentation drag and drop which I didn't manage to do yet.

I did this now and got to a reasonable proof-of-concept with the file manager demo. Do you want to take a look and comment on the changes I had to make to get this working?

Core/clim-core/presentation-drag-and-drop.lisp
```diff
@@ -129,60 +125,70 @@
...
             (setq last-event event
                   last-presentation nil)
-            (maybe-funcall feedback-fn frame from-presentation
-                           window initial-x initial-y x y :highlight)
-            (document-drag-n-drop translator nil
-                                  context-type frame event window
-                                  x y))
+            (update-feedback-and-documentation translator nil event window x y))
+          (:keyboard (&key event)
+            (let ((new-event (shallow-copy-object last-event)))
```
dkochmanski: this lands me in the debugger when I have shift pressed when moving Trash (because we try to make a shallow-copy of NIL).

Examples/file-manager.lisp
```diff
@@ -0,0 +1,260 @@
...
+          :reader   root
+          :initform (make-root
+                     (make-directory "Trash")
+                     (make-directory "Directory 1")
+                     (make-directory "Directory 2"
+                                     (make-file "baz")
+                                     (make-file "fez"))
+                     (make-file "whoop"))))
+  (:panes
+   (explananation :label :label "Drag files and directories. Modifier keys change move, copy, etc.")
```
dkochmanski: more elaborate explanation (which modifier does what). Also "Modifier keys change > drop action to < …"

Core/clim-core/presentation-drag-and-drop.lisp
dkochmanski: it seems to not be easily reproducible because now when I try it I select the text via text-selection protocol

Examples/file-manager.lisp
```diff
@@ -0,0 +1,260 @@
...
+              (height     (bounding-rectangle-height from-presentation))
+              (text       (string-downcase mode))
+              (text-width (text-size stream text)))
+         (repaint-sheet stream (make-rectangle* (+ x1 offset-x -1)
+                                                (+ y1 offset-y -1)
+                                                (+ x1 offset-x width 2 4 text-width)
+                                                (+ y1 offset-y height 2 4))))))))
+
+(defun drag-file-feedback/copy (frame from-presentation stream x0 y0 x1 y1 state)
+  (drag-file-feedback frame from-presentation stream x0 y0 x1 y1 state :copy))
```
dkochmanski: when I copy directory I get a file with the same name (but not a directory)

dkochmanski: either that or symlink-directory-on-copy or refusing to copy a directory (with a notification, that it has files and copy is not recursive).

Core/clim-core/presentation-drag-and-drop.lisp
dkochmanski: press pointer over a file (do not move it), and after that press the modifier

Examples/file-manager.lisp
dkochmanski: recursive copy directory to one of its ancestors blows the stack

Core/clim-core/presentation-drag-and-drop.lisp
dkochmanski: this is still a problem. why is there a shallow copy and reinitialize-instance? can't we simply use the new event in the invocation to update-for-presentation?

Backends/CLX/keysyms-common.lisp
```diff
@@ -121,7 +123,7 @@
 ;;; This dictionary maps CLX keysym names to the power-of-two
 ;;; constants that the CLIM II specification requires.
 (defconstant +clim-modifiers+
-  '(((:meta-left :meta-right) #.+meta-key+)
+  '(((:meta-left :meta-right :alt-left :alt-right) #.+meta-key+)
```
dkochmanski: I agree that they should be treated like other modifiers. That said Alt and Meta are semantically different things, so alt-left and alt-right should resolve to +alt-key+ (that is - we should map to this key -- n.b such constant is already defined).

Core/clim-basic/extended-streams/pointer-tracking.lisp
```diff
@@ -143,28 +143,31 @@
...
+        last-x last-y)
+    (flet ((pointer-event-position (event)
+             (let ((sheet (event-sheet event)))
+               (get-pointer-position (sheet event)
+                 (if (not transformp)
+                     (values x y)
+                     (with-sheet-medium (medium sheet)
+                       (transform-position
+                        (medium-transformation medium) x y)))))))
+      (letf (((port-keyboard-input-focus (port sheet)) sheet))
```
dkochmanski: we already do steal the event thanks to the `event-read`, there is no need to change the input focus. notice, that we do pass the keyboard event as it is without changing the event-sheet in case of pointer events; so we already do allow mismatches in the handler implementations (they are not handle-event methods after all).

Core/clim-basic/extended-streams/pointer-tracking.lisp
```diff
@@ -143,28 +143,31 @@
...
-              (handle-event (event-sheet event) event)
-              (multiple-value-bind (x y)
-                  (when (typep event 'pointer-event)
-                    (pointer-event-position event))
-                (track-event state event x y))))))
+(defun invoke-tracking-pointer (state)
+  (let ((sheet (tracked-sheet state))
+        (multiple-window (multiple-window state))
+        (transformp (transformp state))
+        last-x last-y)
```
dkochmanski: keyboard events do have x and y coordinates, were these coordinates problematic for some reason?

Examples/file-manager.lisp
scymtym: Done (thanks to multi-line labels).

Core/clim-core/presentation-drag-and-drop.lisp
scymtym: Fixed by synthesizing a pointer event at the beginning of the pointer tracking loop.

Backends/CLX/keysyms-common.lisp
scymtym: Added a comment as discussed.

Core/clim-basic/extended-streams/pointer-tracking.lisp
scymtym: Are you suggesting to "steal" all `keyboard-event`s irregardless `event-sheet` and `multiple-window`?

scymtym: Passing the unmodified `keyboard-event` to the handler does not work because its event sheet is determined by the keyboard input focus of the port. The problem is that feedback functions try to draw onto the sheet of the event which is the wrong sheet in this case.

dkochmanski:
```
(if (and (not multiple-window)
                   (not (eql sheet (event-sheet event))))
              ;; Event is not intercepted.
              (handle-event (event-sheet event) event)
              (multiple-value-bind (x y)
                  (when (typep event 'pointer-event)
                    (pointer-event-position event))
                (track-event state event x y)))
```

first arm of this if says it all: if multiple-window is NIL and event-sheet does not match, then we do not steal it. we track events only on a sheet on which tracking-pointer is invoked. If multiple-window is T, we always steal events and we never modify it (i.e do not copy it nor doesn't change the target sheet).

In case of tracking pointer (when multiple windows are involved), we need to be vary that we track the pointer "for sheet" on "multiple windows". Event contains the event window (and it is often essential to be able to tell, what is the event window i.e to draw something under the pointer!), so this should not change. If you want to access the sheet "for which" we track, you need to access it from the tracking-pointer-state parameter with a method `tracked-sheet` (internal interface). Does it make sense?

dkochmanski: so: window = event's sheet, sheet = tracked sheet. otherwise imo it would be close to impossible to implement sensibly multiple-window.

dkochmanski:
Please see commit https://github.com/McCLIM/McCLIM/pull/943/commits/32fa1b71223dbc0a20f547f6ebede8ff841d90e8 , it solves the issue with modifiers during dnd without changes in ec90101 and d344f3e

dkochmanski:
given 32fa1b7, is there a need for ec90101 and d344f3e?

scymtym:
> given 32fa1b7, is there a need for ec90101 and d344f3e?

Answering on a pedantic level, that commit is on a different branch, so yes, there is still a need for some solution in this branch.

More generally, I still don't know which is better: translating non-pointer events to pointer events or adding handling of non-pointer events in drag and drop processing.

Apart from that, 32fa1b7 has a bug that causes synthesized events to refer to the wrong sheet.

I have an updated version of this branch, but I don't have time to work on this at the moment.

dkochmanski:
I'm pushing to master the original commit from this PR (e4ad1a11).

scymtym:
> I'm pushing to master the original commit from this PR (e4ad1a1).

That should be fine.

Core/clim-basic/extended-streams/pointer-tracking.lisp
```diff
@@ -143,28 +146,45 @@
...
+          (track-pointer-event event)))
+      (loop for event = (event-read tracked-sheet)
+            do (cond ((and (not multiple-window)
+                           (not (eql tracked-sheet (event-sheet event))))
+                      ;; Event is not intercepted.
+                      (handle-event (event-sheet event) event))
+                     ((typep event 'pointer-event)
+                      (track-pointer-event event))
+                     (t
+                      (track-event state event nil nil)))))))
```
dkochmanski: should we maybe pass here (optionally transformed) `device-event-native-x/y`?

in tracking-event `:keyboard` clause only `gesture` is mentioned (in the spec), but we extend it to allow also specifying event, x, y etc arguments. for device-event the reader device-event-native-* is not exported, so asking programmer to query such events for x/y would be unwise, otoh having the pointer position as arguments may make the programmer life much easier.

Examples/file-manager.lisp
```diff
@@ -0,0 +1,349 @@
...
+;;;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;;;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;;;; Library General Public License for more details.
+;;;;
+;;;; You should have received a copy of the GNU Library General Public
+;;;; License along with this library; if not, write to the
+;;;; Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+;;;; Boston, MA  02111-1307  USA.
+
+(defpackage #:mcclim.examples.file-manager
```
dkochmanski: while I agree that mcclim.examples.file-manager is a compelling package name, I think that we should standardize the package names in our examples to something derived from a clim-demo as something what is already there for a long time. I see that we have some defpackage's in Examples, so these should be changed too.

if you have time for that, please rename packages in Examples to clim-demo.demo-name, if you don't, please only rename the package for file-manager demo to `clim-demo.file-manager` and I'll make a PR which changes package names in other demos too later.

Examples/file-manager.lisp
```diff
@@ -0,0 +1,349 @@
...
+  (:panes
+   (explananation :label :label      *explanation*
+                         :background +beige+)
+   (files         :application :display-function (lambda (frame pane)
+                                                   (let ((root (root frame)))
+                                                     (present root (presentation-type-of root)
+                                                              :stream     pane
+                                                              :single-box t)))
+                               :default-view     (make-instance 'file-manager-view))
+   (interactor    :interactor))
```
dkochmanski: please remove column indent, it looks odd with multiline arguments.

Core/clim-basic/extended-streams/pointer-tracking.lisp
scymtym: I would prefer to leave that question for another pull request (passing `nil nil` for non-pointer events is not new, but it wasn't as obvious in the previous spelling of this section).

Examples/file-manager.lisp
scymtym: Changed.

Examples/file-manager.lisp
scymtym: Changed.

dkochmanski:
lgtm, thanks

PR #871, opened 2019-10-14
Sort out dependency issues between design.lisp and regions.lisp

scymtym:
Before this change, the compile and load order of design.lisp and regions.lisp was not explicitly specified and loading design.lisp first lead to a warning due to `+identity-transformation+` not being defined yet.

Generally speaking, the geometry module (including `regions.lisp`) is more fundamental and should be compiled and loaded first. The only obstacle were `+{no,every}where+`, `{no,every}where-mixin` which are defined in `design.lisp` but initialized in `regions.lisp`. To allow compiling and loading `regions.lisp` first (and avoid the definition and initialization of `+{no,every}where+` in distinct places), `+{no,every}where+` and `{no,every}where-mixin` are now defined in `regions.lisp`.

### DIFF
--- a/Core/clim-basic/geometry/regions.lisp
+++ b/Core/clim-basic/geometry/regions.lisp
@@ -53,13 +53,26 @@
 
 (in-package :clim-internals)
 
-(defclass nowhere-region (region nowhere-mixin) ())
+;;;; Design <-> Region Equivalences
+
+;;; As Gilbert points in his notes, transparent ink is in every
+;;; respect interchangable with the nowhere region, and likewise
+;;; foreground ink is interchangable with the everywhere region.
+;;; By defining the following mixins and adding them to the
+;;; appropriate ink/region class pairs, we can reduce the number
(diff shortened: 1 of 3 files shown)


Core/clim-basic/geometry/regions.lisp
```diff
@@ -53,13 +53,26 @@
...
 
-(defclass nowhere-region (region nowhere-mixin) ())
+;;;; Design <-> Region Equivalences
+
+;;; As Gilbert points in his notes, transparent ink is in every
+;;; respect interchangable with the nowhere region, and likewise
+;;; foreground ink is interchangable with the everywhere region.
+;;; By defining the following mixins and adding them to the
+;;; appropriate ink/region class pairs, we can reduce the number
+;;; of methods necessary.
```
dkochmanski: "of methods necessary in design module".

scymtym: Fixed.

dkochmanski:
merged, thanks

PR #872, opened 2019-10-15
Fixes for the TEXT-STYLE-TO-FONT removal

scymtym:
This fixes minor problems introduced in b2e1930b.

* The `text-style-mapping` can be specialized to `standard-text-style` instead of checking the type in the body. It can also be written more concisely using `alexandria:ensure-gethash`.

* One of the `text-style-mapping` calls in the freetype backend lacked the `clim:` package prefix.

* Indentation fixes.

* Test `stream-character-width` in the `print-test-page` test function.

### DIFF
--- a/Backends/CLX/fonts.lisp
+++ b/Backends/CLX/fonts.lisp
@@ -57,12 +57,13 @@
                     (open-font display
                                (format nil "-~a-~a-*-*-~d-*-*-*-*-*-~a"
                                        family-name face-name size encoding))))
-;;; xxx: this part is a bit problematic - we either list all fonts
-;;; with any possible encoding (what leads to the situation, when our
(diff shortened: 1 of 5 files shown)


dkochmanski:
lgtm, thanks!

admich:
Now with `text-style-mapping` specialized to `standard-text-style` instead of checking the type in the body if you call the function like this:
`(text-style-mapping (find-port) '())`
the font is not cached because the method is not called: `nil` is not `standard-text-style`. 
In the previous version the type checked inside the body was the type of the parsed style.

scymtym:
> In the previous version the type checked inside the body was the type of the parsed style.

Right, sorry. I missed that because of the `&aux` rebinding.

PR #873, opened 2019-10-16
Fix initialization of POSTSCRIPT-STREAM

scymtym:
* Fix initialization of `postscript-stream` (fixes failing unit test)

* Move other Postscript-specific unit tests into the Postscript backend system.


dkochmanski:
lgtm, n.b comment about revisiting inheritance just above the first change is interesting. it may be worth investigating when someone gets into updating-output machinery.

PR #885, opened 2019-10-24
move the windowing substrate to a separate module

dkochmanski:
This function had some problems:

- for temporary-medium-sheet-output-mixin we've replaced the medium
  even if some was already bound (spec violation)

- specialization on basic-medium was bogus (spec mentions only the
  medium protocol class)

I've also removed some comments expressing lack of understending of
the macro purpose. Its purpose is to associate the medium to a sheet
without any (nested call won't associate a new medium because there is
one already). It makes sense for mediums which creation is expensive.

This change is a requirement before factoring out a "windowing" module
like we did with the "geometry" module (it removes a circular
dependency between drawing primitives and windowing primitives).


dkochmanski:
superseded by #886.

PR #886, opened 2019-10-25
clim-basic: group files by module

dkochmanski:
Some files in modules should be split further. Also we still have some minor forward-referenced calls. Artificial separation between clim-basic and clim-core imposed by drei dependency should be get rid of and all parts should be put in modules in a single system definition.


Core/clim-basic/clim-basic.asd
```diff
@@ -38,7 +38,7 @@
                  (:file "grafts"  :depends-on ("sheets" "ports"))
                  (:file "repaint" :depends-on ("sheets" "ports" "grafts" "events" "output"))))
    (:file "design" :depends-on ("decls" "protocol-classes" "utils" "geometry"))
-   (:file "X11-colors" :depends-on ("decls" "protocol-classes" "multiprocessing" "design" "geometry"))
+   (:file "colors" :depends-on ("design"))
```
scymtym: This change in filename should be reflected in `Tools/generated-named-colors.lisp`. In fact, that tool should be changed to generate a comment mentioning that the file has been generated and by what it has been generated.

Core/clim-basic/output.lisp
```diff
@@ -90,49 +90,29 @@
...
+  (:argument-precedence-order sheet medium continuation)
+  (:method (continuation medium sheet)
+    (funcall continuation medium))
+  (:method (continuation medium (sheet permanent-medium-sheet-output-mixin))
+    (funcall continuation (sheet-medium sheet)))
+  (:method (continuation medium (sheet temporary-medium-sheet-output-mixin))
+    (if-let ((sheet-medium (sheet-medium sheet)))
+      (funcall continuation sheet-medium)
+      (if (null medium)
+          (let ((new-medium (allocate-medium (port sheet) sheet)))
```
scymtym: Keep the port in a local variable.

Core/clim-basic/repaint.lisp
```diff
@@ -93,8 +93,8 @@
     (everywhere-region
```
scymtym: "referencign" is a typo in the commit message.

"because sheet may not have any associated at the time being." -> "because the sheet may not have an associated medium at that time."

Core/clim-basic/clim-basic.asd
```diff
@@ -1,3 +1,5 @@
+(in-package #:asdf-user)
```
scymtym: I suggest doing "ASDF cleanup" in a separate commit for all defined systems after discussing which style we want to use.

Core/clim-basic/clim-basic.asd
dkochmanski: I've added comment in generated file. I've also squashed rename with creation of the module. it is a minor name change so there is no point in having a separate commit.

Core/clim-basic/windowing/sheets.lisp
```diff
@@ -411,14 +411,12 @@
   (with-slots (device-region) sheet
```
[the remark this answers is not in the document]
dkochmanski: fixed

Core/clim-basic/windowing/repaint.lisp
```diff
@@ -93,6 +93,7 @@
     (everywhere-region
      (call-next-method))
     (otherwise
+     (call-next-method)
```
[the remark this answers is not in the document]
dkochmanski: whoops, I **forgot** to remove it, sorry about that. fixed.

Core/clim-basic/repaint.lisp
```diff
@@ -232,13 +232,13 @@
...
            (native-sheet-region (effective-repaint-region parent sheet region)))
-          (with-sheet-medium (medium parent)
-            (with-drawing-options (medium :clipping-region native-sheet-region
-                                          :ink (pane-background sheet)
-                                          :transformation +identity-transformation+)
-              (with-bounding-rectangle* (left top right bottom)
-                  native-sheet-region
-                (medium-draw-rectangle* medium left top right bottom t)))))))
+      (with-sheet-medium (medium parent)
+        (letf (((medium-clipping-region medium) native-sheet-region)
```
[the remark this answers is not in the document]
dkochmanski: It does, it is changed to avoid a forward-dependency. Nevertheless the intention here is to clear the area, not to draw something, so it is semantically correct. Whenever specialization on always-repaint-background-mixin belongs to repaint.lisp is another question (arguably that should be handled by the medium itself).

Core/clim-basic/repaint.lisp
dkochmanski: fixed

Core/clim-basic/output.lisp
```diff
@@ -90,49 +90,29 @@
...
+            (unwind-protect
+                 (progn
+                   (engraft-medium new-medium (port sheet) sheet)
+                   (setf (%sheet-medium sheet) new-medium)
+                   (funcall continuation new-medium))
+              (setf (%sheet-medium sheet) nil)
+              (degraft-medium new-medium (port sheet) sheet)
+              (deallocate-medium (port sheet) new-medium)))
```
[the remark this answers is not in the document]
dkochmanski: I'll do that then.

Core/clim-basic/output.lisp
dkochmanski: fixed (I've put it in a separate commit; it was like that before)

Issue #880, opened 2019-10-22
text-margin initarg doesn't work with standard-page-layout initialize-instance


dkochmanski:
Steps to reproduce:

```common-lisp
(clim:define-application-frame mcp/desktop ()
  ()
  (:command-table (mcp/desktop :inherit-from nil))
  (:geometry :width 800 :height 600)
  (:pane :interactor ;; :text-margins '(:left 15
                     ;;                 :top  8)
         :text-margin 30))

(clim:run-frame-top-level (clim:make-application-frame 'mcp/desktop))
```

expected: right margin is set to `(:absolute ,text-margin)` (see https://github.com/McCLIM/McCLIM/issues/680 for discussion why this is not quite right and why we chose that)

observed: invalid margin-spec condition

PR #887, opened 2019-10-25
Casual improvements

dkochmanski:
Supersedes #882. This branch is based on #886 so it should be reviewed after merging it.

### DIFF
--- a/Core/clim-basic/extended-streams/text-formatting.lisp
+++ b/Core/clim-basic/extended-streams/text-formatting.lisp
@@ -22,7 +22,7 @@
   (ignore-errors ; destructuring-bind may error; that yields invalid spec
     (destructuring-bind (&key left top right bottom) margins
       (flet ((margin-spec-p (m)
-               (and (member (first m) '(:absolute :relative))
+               (and (member (first m) '(:relative :absolute))
                     (not (null (second m))))))
         (every #'margin-spec-p (list left top right bottom))))))
 
@@ -31,8 +31,8 @@
 
 (defclass standard-page-layout ()
   ((%page-region :reader stream-page-region :writer (setf %page-region))
-   (margins :initform '(:left   (:absolute 0)
-                        :top    (:absolute 0)
+   (margins :initform '(:left   (:relative 0)
+                        :top    (:relative 0)
                         :right  (:relative 0)
                         :bottom (:relative 0))
             :accessor stream-text-margins :type margin-spec)))
@@ -40,14 +40,16 @@
 (defmethod initialize-instance :after ((instance standard-page-layout)
                                        &key text-margins text-margin)
   (macrolet ((thunk (edge default)
-               `(unless (getf text-margins ,edge)
-                  (setf (getf text-margins ,edge) ,default))))
-    (thunk :left   '(:absolute 0))
-    (thunk :top    '(:absolute 0))
+               `(let ((spec (getf text-margins ,edge)))
+                  (typecase spec
+                    (null (setf (getf text-margins ,edge) ,default))
+                    (atom (setf (getf text-margins ,edge) `(:relative ,spec)))))))
+    (thunk :left   `(:relative 0))
+    (thunk :top    `(:relative 0))
     (thunk :right  (if text-margin
                        `(:absolute ,text-margin)
                        `(:relative 0)))
-    (thunk :bottom '(:relative 0))
+    (thunk :bottom `(:relative 0))
     (setf (stream-text-margins instance) text-margins)))
 
 (defgeneric stream-cursor-initial-position (stream)
@@ -74,10 +76,10 @@
           (y2 (* 43 (text-style-height (stream-text-style stream) stream))))
       (setf sheet-region (make-rectangle* 0 0 x2 y2))))
   (with-bounding-rectangle* (x1 y1 x2 y2) sheet-region
-    (macrolet ((thunk (margin edge sign direction)
+    (macrolet ((thunk (margin edge sign orientation)
                  `(if (eql (first ,margin) :absolute)
-                      (parse-space stream (second ,margin) ,direction)
-                      (,sign ,edge (parse-space stream (second ,margin) ,direction)))))
+                      (parse-space stream (second ,margin) ,orientation)
+                      (,sign ,edge (parse-space stream (second ,margin) ,orientation)))))
       (destructuring-bind (&key left top right bottom) (stream-text-margins stream)
         (setf (%page-region stream)
               (make-rectangle* (thunk left   x1 + :horizontal)
@@ -93,9 +95,10 @@
     (new-margins (stream standard-page-layout)
      &aux (old-margins (stream-text-margins stream)))
   (macrolet ((thunk (edge)
-               `(unless (getf new-margins ,edge)
-                  (setf (getf new-margins ,edge)
-                        (getf old-margins ,edge)))))
+               `(let ((spec (getf new-margins ,edge)))
+                  (typecase spec
+                    (null (setf (getf new-margins ,edge) (getf old-margins ,edge)))
+                    (atom (setf (getf new-margins ,edge) `(:relative ,spec)))))))
     (thunk :left)
     (thunk :top)
     (thunk :right)
--- a/Core/clim-basic/extended-streams/text-formatting.lisp
+++ b/Core/clim-basic/extended-streams/text-formatting.lisp
@@ -31,11 +31,11 @@
 
 (defclass standard-page-layout ()
   ((%page-region :reader stream-page-region :writer (setf %page-region))
-   (margins :initform '(:left   (:relative 0)
-                        :top    (:relative 0)
-                        :right  (:relative 0)
-                        :bottom (:relative 0))
-            :accessor stream-text-margins :type margin-spec)))
+   (margins :accessor stream-text-margins :type margin-spec))
+  (:default-initargs :text-margins '(:left   (:relative 0)
+                                     :top    (:relative 0)
+                                     :right  (:relative 0)
+                                     :bottom (:relative 0))))
 
 (defmethod initialize-instance :after ((instance standard-page-layout)
                                        &key text-margins text-margin)
@@ -50,7 +50,7 @@
                        `(:absolute ,text-margin)
                        `(:relative 0)))
     (thunk :bottom `(:relative 0))
-    (setf (stream-text-margins instance) text-margins)))
+    (setf (slot-value instance 'margins) text-margins)))
 
 (defgeneric stream-cursor-initial-position (stream)
   (:documentation "Returns two values: x and y initial position for a cursor on page.")
--- a/Core/clim-basic/extended-streams/text-formatting.lisp
+++ b/Core/clim-basic/extended-streams/text-formatting.lisp
@@ -44,7 +44,9 @@
                   (setf (getf text-margins ,edge) ,default))))
     (thunk :left   '(:absolute 0))
     (thunk :top    '(:absolute 0))
-    (thunk :right  (or text-margin '(:relative 0)))
+    (thunk :right  (if text-margin
+                       `(:absolute ,text-margin)
+                       `(:relative 0)))
     (thunk :bottom '(:relative 0))
     (setf (stream-text-margins instance) text-margins)))
 
--- a/Core/clim-basic/extended-streams/text-formatting.lisp
+++ b/Core/clim-basic/extended-streams/text-formatting.lisp
@@ -20,14 +20,28 @@
 
 (defun valid-margin-spec-p (margins)
   (ignore-errors ; destructuring-bind may error; that yields invalid spec
-    (destructuring-bind (&key left top right bottom) margins
-      (flet ((margin-spec-p (m)
-               (and (member (first m) '(:relative :absolute))
-                    (not (null (second m))))))
-        (every #'margin-spec-p (list left top right bottom))))))
+   (destructuring-bind (&key left top right bottom) margins
+     (flet ((margin-spec-p (margin)
+              (destructuring-bind (anchor value) margin
+                (and (member anchor '(:relative :absolute))
+                     (realp value)))))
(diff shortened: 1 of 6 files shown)


Core/clim-basic/extended-streams/text-formatting.lisp
```diff
@@ -44,7 +44,9 @@
                   (setf (getf text-margins ,edge) ,default))))
```
scymtym: Commit message: "stream-text-margin" -> "stream-text-margins"

Core/clim-basic/extended-streams/text-formatting.lisp
```diff
@@ -44,7 +44,9 @@
...
     (thunk :left   '(:absolute 0))
     (thunk :top    '(:absolute 0))
-    (thunk :right  (or text-margin '(:relative 0)))
+    (thunk :right  (if text-margin
+                       `(:absolute ,text-margin)
+                       `(:relative 0)))
     (thunk :bottom '(:relative 0))
```
scymtym: I suggest extracting this logic into a function `(normalize-text-margins text-margins text-margin)` or similar and adding a unit test.

Core/clim-core/bordered-output.lisp
```diff
@@ -456,8 +456,29 @@
                           (draw-line* stream left bottom right bottom
                                       :ink ink
                                       :line-style line-style)))
-                       (updating-output-record  nil)
-                       (compound-output-record  (fn child))))))
+                       (updating-output-record nil)
+                       (compound-output-record (fn child))))))
+      (fn record))))
+
+(define-border-type :crossout (stream record
```
scymtym: This is often called "strikethrough". Is there a specific reason for going with "crossout"?

Core/clim-core/bordered-output.lisp
```diff
@@ -456,8 +456,29 @@
                           (draw-line* stream left bottom right bottom
```
scymtym: In `examples`, there is a bordered output test and a dedicated underlining test (within drawing tests?). One or both should test this feature.

Core/clim-basic/extended-streams/text-formatting.lisp
```diff
@@ -22,7 +22,7 @@
   (ignore-errors ; destructuring-bind may error; that yields invalid spec
     (destructuring-bind (&key left top right bottom) margins
       (flet ((margin-spec-p (m)
-               (and (member (first m) '(:absolute :relative))
+               (and (member (first m) '(:relative :absolute))
```
scymtym: Can you add a `destructuring-bind` to check the shape of `m` and document the meaning of its parts?

Core/clim-basic/extended-streams/text-formatting.lisp
```diff
@@ -93,9 +95,10 @@
     (new-margins (stream standard-page-layout)
      &aux (old-margins (stream-text-margins stream)))
   (macrolet ((thunk (edge)
-               `(unless (getf new-margins ,edge)
-                  (setf (getf new-margins ,edge)
-                        (getf old-margins ,edge)))))
+               `(let ((spec (getf new-margins ,edge)))
+                  (typecase spec
+                    (null (setf (getf new-margins ,edge) (getf old-margins ,edge)))
```
scymtym: Pull the common `setf (getf new-margins ,edge)` part out of the `typecase`.

Core/clim-basic/extended-streams/text-formatting.lisp
```diff
@@ -31,23 +31,25 @@
...
 
 (defmethod initialize-instance :after ((instance standard-page-layout)
                                        &key text-margins text-margin)
   (macrolet ((thunk (edge default)
-               `(unless (getf text-margins ,edge)
-                  (setf (getf text-margins ,edge) ,default))))
-    (thunk :left   '(:absolute 0))
-    (thunk :top    '(:absolute 0))
+               `(let ((spec (getf text-margins ,edge)))
+                  (typecase spec
```
scymtym: Since this is repeated below, add a `normalize-margin-spec` or similar helper function.

Core/clim-basic/extended-streams/text-formatting.lisp
```diff
@@ -31,23 +31,25 @@
 
 (defclass standard-page-layout ()
   ((%page-region :reader stream-page-region :writer (setf %page-region))
-   (margins :initform '(:left   (:absolute 0)
-                        :top    (:absolute 0)
+   (margins :initform '(:left   (:relative 0)
+                        :top    (:relative 0)
```
scymtym: Why is this changed? This does not seem purely syntactic.

Core/clim-basic/extended-streams/text-formatting.lisp
```diff
@@ -50,7 +50,7 @@
                        `(:absolute ,text-margin)
                        `(:relative 0)))
     (thunk :bottom `(:relative 0))
-    (setf (stream-text-margins instance) text-margins)))
+    (setf (slot-value instance 'margins) text-margins)))
```
scymtym: Why not use the accessor?

Core/clim-core/bordered-output.lisp
dkochmanski: I've checked names and I had two choices with the same meaning: `crossout` and `strikethrough`, I've picked the shorter one.

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: left/top is always located at 0/0, I've changed it to `:relative` to match bottom/right defaults (so the person who reads the code doesn't have to wonder why they are different).

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: this commit message is correct, it addresses #880 where `:text-margin` - which is the initarg for a slot accessed with `stream-text-margin`, didn't work quite correct with standard-page-layout (and stream-text-margins).

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: it doesn't change the value when spec is a list which is not empty. pulling it out would be incorrect.

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: it is slightly different (below macrolet merges old and new specs while this one provides defaults)

Core/clim-basic/extended-streams/text-formatting.lisp
scymtym: I thought with a `normalize-margin-spec` function (mentioned in another comment below) which would be the identity on (well-formed) conses, this could be `(setf (getf text-margins ,edge) (normalize-margin-spec (getf text-margins ,edge)))`.

Now that I think about it some more, why not make a function that normalizes (and checks?) a whole "margin spec plist"? It could be used here as well as in `(setf stream-text-margins) :around` and also tested.

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: because `(setf stream-text-margins)` has an around method which tries to merge new margins with old ones (and that calls `stream-text-margins` on an the yet-unbound slot)

Core/clim-basic/extended-streams/text-formatting.lisp
scymtym: As far as I can tell, a function `normalize-margin-spec-plist plist defaults` called as

* `(normalize-margin-spec-plist text-margins '(:left (:absolute 0) :top (:absolute 0) …))`

* `(normalize-margin-spec-plist text-margins old-margins)`

would do the job.

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: this logic with default value is used only in this particular method (there is something similar but not quite the same in setf stream-text-margins) -- is there much of a value in extracting it?

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: given it looks like this:
```
(defun valid-margin-spec-p (margins)
  (ignore-errors ; destructuring-bind may error; that yields invalid spec
   (destructuring-bind (&key left top right bottom) margins
     (flet ((margin-spec-p (margin)
              (destructuring-bind (anchor value) margin
                (and (member anchor '(:relative :absolute))
                     (realp value)))))
       (every #'margin-spec-p (list left top right bottom))))))
```

how would you like me to document it?

(notice, that I've changed (not (null value)) to (realp value))

scymtym: `destructuring-bind (anchor value) margin` is exactly what I meant by documenting the shape using `destructuring-bind`.

Core/clim-core/bordered-output.lisp
dkochmanski: done

dkochmanski:
I've modified valid-margin-spec-p and added examples for crossout. I've also rebased against master.

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: done

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: ditto

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: done

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: done

Tests/text-formatting.lisp
```diff
@@ -0,0 +1,70 @@
...
+
+(test normalize-margin-spec
+  (flet ((equals (spec-1 spec-2)
+           (and (= (length spec-1) (length spec-2))
+                (loop for (edge value) on spec-1 by #'cddr
+                      unless (equal value (getf spec-2 edge))
+                        do
+                           (return nil)
+                      finally
+                         (return t)))))
```
scymtym: `always (equal value (getf spec-2 edge))`

Tests/text-formatting.lisp
```diff
@@ -0,0 +1,70 @@
...
+    (is (equals (climi::normalize-margin-spec nil
+                                              '(:left (:relative 0)
+                                                :top (:relative 0)
+                                                :right (:relative 0)
+                                                :bottom (:relative 0)))
+                '(:left (:relative 0)
+                  :top (:relative 0)
+                  :right (:relative 0)
+                  :bottom (:relative 0))))
```
scymtym: Fiveam wants the expected value as the first argument. Putting the expected value first generates the report
```
(climi::normalize-margin-spec …)

  evaluated to

(:top …)
```

instead of

```
  (:top …)

evaluated to

  (:top …)
```

Tests/text-formatting.lisp
```diff
@@ -0,0 +1,70 @@
...
+           :foo)
+         default))
+      (signals error
+        (climi::normalize-margin-spec
+         `(,(5am:gen-one-element :left :right :top :bottom)
+           (:relative 1 2))
+         default))
+      (signals error
+        (climi::normalize-margin-spec `(:left* (:relative 42))
+                                      default)))))
```
scymtym: * The `5am` package prefix is not needed since we `:use` the package
* `gen-one-element` returns a generator function, so ``(,(gen-one-element …) …)` splices the function, not a random value
* `gen-one-element` and other generators are intended to be used in conjunction with `for-all`

I suggest testing the validation behavior like this:
```lisp
(defun gen-invalid-margin-spec-element ()
  (let ((anchor (gen-one-element :relative :absolute)))
   (lambda ()
     (case (random 3)
       (0 :foo)
       (1 (list (funcall anchor) :foo))
       (2 (list (funcall anchor) 1 2))))))

(defun gen-invalid-margin-spec ()
  (let ((edge (gen-one-element :left :right :top :bottom))
        (element (gen-invalid-margin-spec-element)))
    (lambda ()
      (list (funcall edge) (funcall element)))))

(test normalize-margin-spec.validation

  (for-all ((spec (gen-invalid-margin-spec)))
    (let ((default '(:left (:relative 0)
                     :top (:relative 0)
                     :right (:relative 0)
                     :bottom (:relative 0))))
      (signals error (climi::normalize-margin-spec spec default)))))
```

Core/clim-basic/extended-streams/text-formatting.lisp
```diff
@@ -40,17 +53,13 @@
...
-                       `(:relative 0)))
-    (thunk :bottom `(:relative 0))
+  (let ((right-margin (if text-margin
+                          `(:absolute ,text-margin)
+                          `(:relative 0))))
+    (setf text-margins (normalize-margin-spec text-margins `(:left   (:relative 0)
+                                                             :top    (:relative 0)
+                                                             :right  ,right-margin
+                                                             :bottom (:relative 0))))
     (setf (slot-value instance 'margins) text-margins)))
```
scymtym: You could just `(setf (slot-value instance 'margins) (normalize …))` now.

Tests/text-formatting.lisp
dkochmanski: done

Tests/text-formatting.lisp
dkochmanski: done

Tests/text-formatting.lisp
dkochmanski: done

Core/clim-basic/extended-streams/text-formatting.lisp
dkochmanski: done

PR #896, opened 2019-11-07
Implement %invalidate-cached-device-regions

### LINKED ISSUE #895: Unimplemented method when writing text to a pixmap
When using `DRAW-TEXT*` to write text to a pixmap, an error occurs due to a missing implementation of the method `%INVALIDATE-CACHED-DEVICE-REGIONS` for pixmaps.

lokedhs:
This provides a default implementation for
%invalidate-cached-device-regions so that the application will not
crash even if a sheet does not specialise this method.

Fixes #895


dkochmanski:
I'm aware that there was some deliberation where to put the method, but since there is no review remarks I'm accepting it as a good enough fallback.

PR #898, opened 2019-11-09
utils: parse-space: use function instead of a macro

### LINKED ISSUE #893: warning when compiling parse-space
I see the following warning when I compile parse-space Core/clim-basic/utils.lisp:

```

; file: /home/sly/projects/McCLIM/Core/clim-basic/utils.lisp
; in: DEFUN CLIM-INTERNALS::PARSE-SPACE
;     (STREAM)
; 
; caught WARNING:
;   The function STREAM is undefined, and its name is reserved by ANSI CL so that
;   even if it were defined later, the code doing so would not be portable.

;     (CLIM:WITH-OUTPUT-TO-OUTPUT-RECORD (STREAM)
;       (FUNCALL CLIM-INTERNALS::SPECIFICATION))
; 
; caught STYLE-WARNING:
;   undefined function: CLIM:WITH-OUTPUT-TO-OUTPUT-RECORD
; 
```

dkochmanski:
Macro with-output-to-output-record is not defined until later (so when
it is compiled it is assumed to be a function and a warning is
signaled). Use invoke-with-output-to-output-record which is already
declared. Fixes #893.

### DIFF
--- a/Core/clim-basic/utils.lisp
+++ b/Core/clim-basic/utils.lisp
@@ -498,8 +498,12 @@ STREAM in the direction DIRECTION."
                              (ecase direction
                                (:horizontal width)
                                (:vertical height))))
-    (function (let ((record (with-output-to-output-record (stream)
-                              (funcall specification))))
+    (function (let ((record
+                      (invoke-with-output-to-output-record
+                       stream (lambda (s o)
+                                (declare (ignore s o))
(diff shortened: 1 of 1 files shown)


slyrus:
See proposed alternative fix in https://github.com/McCLIM/McCLIM/pull/900

PR #904, opened 2019-11-15
display string in pointer-documentation-pane

admich:
Old version of FRAME-DISPLAY-POINTER-DOCUMENTATION-STRING didn't
trigger the redisplay of the pointer-documentation-pane so the string
were not visible (e.g documentation in some scigraph commands).

----

#

### DIFF
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -1366,7 +1366,38 @@ have a `pointer-documentation-pane' as pointer documentation,
 
 (defun frame-display-pointer-documentation-string (frame string)
   (with-output-to-pointer-documentation (stream frame)
-    (write-string string stream)))
+      (write-string string stream))
+  (let ((*pointer-documentation-output* (frame-pointer-documentation-output frame)))
+    ;; To see the string it is necessary to trigger the redisplay of
+    ;; pointer-documentation-pane with FRAME-UPDATE-POINTER-DOCUMENTATION.
+    ;; As INPUT-CONTEXT we pass NIL. FRAME-COMPUTE-POINTER-DOCUMENTATION-STATE and
+    ;; FRAME-PRINT-POINTER-DOCUMENTATION specialize on that.
+    ;; We pass the STRING as EVENT argument in this way
+    ;; FRAME-COMPUTE-POINTER-DOCUMENTATION-STATE calculate a new state
+    ;; value cached for icremental-redisplay machinery.  -- admich 2019-11-15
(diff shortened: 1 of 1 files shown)


Core/clim-core/frames.lisp
```diff
@@ -1366,7 +1366,41 @@ have a `pointer-documentation-pane' as pointer documentation,
 
 (defun frame-display-pointer-documentation-string (frame string)
   (with-output-to-pointer-documentation (stream frame)
-    (write-string string stream)))
+      (write-string string stream))
+  (let ((*pointer-documentation-output* (frame-pointer-documentation-output frame)))
+    ;; To see the string it is necessary to trigger the redisplay of
+    ;; pointer-documentation-pane with FRAME-UPDATE-POINTER-DOCUMENTATION.
+    ;; I don't know why we need to trigger the update function twice to see
```
dkochmanski: I think we need to first understand why it is needed to be called twice (probably that will reveal an issue somewhere else).

scymtym: > I think we need to first understand why it is needed to be called twice (probably that will reveal an issue somewhere else).

I think I found a missing `force-output` around the pointer documentation code when working on the fb backend. Maybe that's the problem?

dkochmanski: why not `finish-output`?

admich: Actually the two calls of `FRAME-UPDATE-POINTER-DOCUMENTATION` are necessary only if you call `FRAME-DISPLAY-POINTER-DOCUMENTATION-STRING` from SLIME. If you call the function from a CLIM application one call is enough also if you send a string on a different application-frame. So we can:
1. continue to investigate why from `slime` two calls are necessary.
2. remove one of the two calls (`FRAME-DISPLAY-POINTER-DOCUMENTATION-STRING` is supposed to be called from CLIM not from SLIME).
3. leave the two calls

dkochmanski: currently stream access from outside clim loop is prone to race conditions and it is a known problem, so I think it should be just a single call. there is a ticket for stream concurrent access so there is no need for a separate issue either.

admich:
Removed the double call.

admich:
I removed the double call of `FRAME-UPDATE-POINTER-DOCUMENTATION`. Are there other problems with this PR?

nsrahmad:
@admich No problems on your side. I think @dkochmanski got busy with ECL stuff and forgot about this.

I don't see any more problem's so I am merging it.

Issue #654, opened 2019-01-02
accepting-values key bindings


nsrahmad:
From the wiki, Accepting values key-bindings are:
>  
- OK :keystroke `(#\[ :control)`
- Exit :keystroke `(#\] :control)`
- Abort :keystroke `(#\z :control)`
- Next :keystroke `(#\n :meta)`
- Previous :keystroke `(#\p :meta)`

When looking at the source (in `Core/clim-core/dialog.lisp`) I found:
- There never was a command named `OK` and `(#\[ :control)` was never bound.
- `Exit` command actually submits the dialog, and should be bound to `#\[ :control)` as `OK` action. Currently, It is bound to `(#\] :control)`.
- `Abort` command which is currently `(#\z :control)` actually cancels the dialog, and should be bound to `(#\] :control)` as `Cancel` action.

lokedhs:
These keybindings are quite non-intuitive, even if the proposed corrections are made. I'd like to see Meta-RET to submit a form.

duuqnd:
I'd like to ask about what the difference between Exit and Abort is supposed to be. I can't quite see how they would be different. Would it be acceptable to have them both do the same thing?

nsrahmad:
@duuqnd I think I made a mistake interpreting what those keys meant. On looking up in Symbolics guide it cleared up. There are only two command needed here which are Exit (which submits the dialog) and Abort (which aborts the dialog).

So "OK" is just a label for Exit and "Cancel" is label for Abort. In Symbolics CLIM, corresponding labels are "Exit" and "Abort".

PR #907, opened 2019-11-29
Fix accepting-values keybindings

duuqnd:
#654 describes problems with the keybindings for accepting-values, and so I've made an attempt at fixing one of them.

No command for OK existed, so I created one. It is bound to `C-[` and `M-RET`. `M-RET` was added based on @lokedhs's suggestion that it is more intuitive.

Exit works like you'd expect OK to. I haven't changed this since the OK button uses the command to confirm, and some people might have gotten used to pressing `C-]` to confirm.

This is my first contribution to McCLIM, so please let me know if I've done something wrong.

### DIFF
--- a/Core/clim-core/dialog.lisp
+++ b/Core/clim-core/dialog.lisp
@@ -440,19 +440,23 @@ highlighting, etc." ))
   (apply #'prompt-for-accept-1 stream type :display-default nil args))
 
 (define-command (com-query-exit :command-table accept-values
-                                :keystroke (#\] :control)
+                                :keystroke (#\[ :control)
                                 :name nil
                                 :provide-output-destination-keyword nil)
     ()
   (signal 'av-exit))
 
+(add-keystroke-to-command-table 'accept-values '(#\Return :meta) :command #'com-query-exit)
+
 (define-command (com-query-abort :command-table accept-values
-                                 :keystroke (#\z :control)
+                                 :keystroke (#\] :control)
                                  :name nil
                                  :provide-output-destination-keyword nil)
     ()
   (and (find-restart 'abort)
        (invoke-restart 'abort)))
+
+(add-keystroke-to-command-table 'accept-values '(#\g :control) :command #'com-query-abort)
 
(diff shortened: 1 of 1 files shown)


nsrahmad:
And Hey, Welcome.

In case you are wodering what to do now. Use `git reset --hard master` to discard you changes, then make the new commits with above changes. and use `git push` with `--force` flag to force update it.

duuqnd:
I've made performed changes requested. The reason why I originally did not want to change the keybinding of Exit was because some people might have gotten used to using it.

Core/clim-core/dialog.lisp
```diff
@@ -440,20 +440,24 @@ highlighting, etc." ))
...
 
 (define-command (com-query-exit :command-table accept-values
-                                :keystroke (#\] :control)
+                                :keystroke (#\[ :control)
                                 :name nil
                                 :provide-output-destination-keyword nil)
     ()
   (signal 'av-exit))
 
+(add-keystroke-to-command-table 'accept-values '(#\Return :meta) :command #'com-query-exit)
```
nsrahmad: Have You tried running and testing this? I think `:command` needs to be a symbol so `'com-query-exit` will work but `#'com-query-exit` wouldn't.

Core/clim-core/dialog.lisp
```diff
@@ -440,20 +440,24 @@ highlighting, etc." ))
...
 (define-command (com-query-abort :command-table accept-values
-                                 :keystroke (#\z :control)
+                                 :keystroke (#\] :control)
                                  :name nil
                                  :provide-output-destination-keyword nil)
     ()
   (and (find-restart 'abort)
        (invoke-restart 'abort)))
 
+(add-keystroke-to-command-table 'accept-values '(#\g :control) :command #'com-query-abort)
```
nsrahmad: Also consider adding `:errorp nil` to both `add-keystroke...` commands to avoid error when reloading McCLIM.

duuqnd:
Alright, that's done.

PR #909, opened 2019-11-29
What works for @armedbear currently

easye:
Initial foray into compatibilty: add a `fix-abcl.lisp`.

PR #911, opened 2019-11-30
abcl: patch for getenv

easye:
*This* turns out to be all that is needed for @armedbear to compile McCLIM master successfully.

Supersedes <https://github.com/McCLIM/McCLIM/pull/909>.


dkochmanski:
lgtm, thanks!

Issue #912, opened 2019-12-03
Frames without menu-bar have empty borders


nsrahmad:
Notice the 4 pixel wide border thing below the title-bar of frames.

![bug](https://user-images.githubusercontent.com/10996250/70030224-c32a5880-15ce-11ea-9d62-0696932b3bea.png)
----

Bug here is that in frame definition, defualt value of `menu-bar` is specified to be `T`. So if you don't want menu-bar, you need to add `(:menu-bar nil)` in frame definition.

Expected output: After adding `(:menu-bar nil)` in `Tables with borders` demo:

![bug](https://user-images.githubusercontent.com/10996250/70030555-77c47a00-15cf-11ea-85fe-e19dad5ba03a.png)

To solve this issue, you would need to add `(:menu-bar nil)` in all frame definitions in `clim-demo` which doesn't require menu-bar.

PR #914, opened 2019-12-04
Add (:menu-bar nil) to demos and tests that don't have menu bars.

duuqnd:
Relevant issue: #912

### DIFF
--- a/Examples/bordered-output-examples.lisp
+++ b/Examples/bordered-output-examples.lisp
@@ -131,6 +131,7 @@
                                     :outline-ink +red+ :background +white+
                                     :shadow +gray80+)))
 	   :reader shapes-of))
+  (:menu-bar nil)
   (:pane
    (scrolling (:width 600 :height 700)
      (make-pane :application-pane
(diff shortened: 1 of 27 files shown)


nsrahmad:
Hi, good work so far. This Fix misses tests in third column of `demodemo` such as `coordinate-swizzling.lisp`, `flipping-ink.lisp` etc.

duuqnd:
That was intentional. Those are labelled in the demo list as "regression tests", so I thought that a change that's merely cosmetic would not be worth adding to them. I could do it anyway if you think I should.

nsrahmad:
Please do it. It is harmless precisely because "it is merely cosmetic".

duuqnd:
Alright, I've added it the the regression tests too.

nsrahmad:
merged, Thank you.
I wanted to tell you about commit messages. See my edited commit message in https://github.com/McCLIM/McCLIM/commit/a98d3e884dbce35faa91bd36d529db1e92cfda0f 

Your commit message had a single line showing "What" but not "Why". GIT allows much larger text  after the single line message which should be used to describe "Why". (For this reason, I always commit in Magit). Also on Github, if you include a line like "Fixes #912", Github will close that issue automatically after merge.

duuqnd:
Thanks! I'll keep that in mind in the future.

PR #916, opened 2019-12-05
implement radio-box-selections and check-box-selections

### LINKED ISSUE #358: [launchpad #493674]  radio-box-selections check-box-selections
https://bugs.launchpad.net/mcclim/+bug/493674

```
Report from Paul Werkowski

 affects mcclim
 status confirmed
 importance low
 tag gadget radio-box check-box
 done

radio-box-selections is supposed to return a list of *all* the toggle-button
gadgets.
Currently it is implemented as

(defmethod radio-box-selections ((pane radio-box))
  (let ((v (radio-box-current-selection pane)))
    (and v (list v))))

check-box-selections is similar, but is not implemented at all.

Paul
```

pouar:
Fixes #358

### DIFF
--- a/Core/clim-core/gadgets.lisp
+++ b/Core/clim-core/gadgets.lisp
@@ -555,8 +555,9 @@ and must never be nil.")
 (defgeneric radio-box-selections (radio-box))
 
 (defmethod radio-box-selections ((pane radio-box))
-  (let ((v (radio-box-current-selection pane)))
-    (and v (list v))))
+  (loop for child in (sheet-children pane)
+        when (typep child 'toggle-button)
+        collect child))
 
 (defmethod value-changed-callback :before (value-gadget (client radio-box) gadget-id value)
   (declare (ignorable value-gadget gadget-id value))
@@ -591,6 +592,11 @@ and must never be nil.")
 
 (defmethod (setf check-box-current-selection) (new-value (check-box check-box))
   (setf (gadget-value check-box) new-value))
+
+(defmethod check-box-selections ((pane check-box))
+  (loop for child in (sheet-children pane)
+        when (typep child 'toggle-button)
+        collect child))
 
 (defmethod value-changed-callback :before (value-gadget (client check-box) gadget-id value)
   (declare (ignorable gadget-id))


pouar:
kinda wish McCLIM has Iterate as a dependency, but the standard Loop macro seems to work here

Core/clim-core/gadgets.lisp
```diff
@@ -555,8 +555,9 @@ and must never be nil.")
 (defgeneric radio-box-selections (radio-box))
```
scymtym: In the first line of the commit message, please describe concise what the commit does.

`fixes #NNN` can be the final line of the commit message.

Core/clim-core/gadgets.lisp
```diff
@@ -555,8 +555,9 @@ and must never be nil.")
...
 
 (defmethod radio-box-selections ((pane radio-box))
-  (let ((v (radio-box-current-selection pane)))
-    (and v (list v))))
+  (loop for child in (sheet-children pane)
+        when (and (typep child 'toggle-button) (gadget-value child))
+        collect child))
```
scymtym: This seems unnecessary since the specification says:

> `radio-box [Class]`
> The instantiable class that implements an abstract radio box, that is, a gadget that constrains a number of toggle buttons, **only one of which may be selected at any one time**. It is a subclass of value-gadget and oriented-gadget-mixin.

Reading more, I'm not sure this method is about the currently selected choice at all. Compare

> `radio-box-current-selection radio-box [Generic Function]`
> Returns (or sets) the **current** selection for the radio box.

and

> `radio-box-selections radio-box [Generic Function]`
> Returns a sequence of **all** of the selections in the radio box.

I now think this should just return all children (maybe filtering out things that are not `toggle-button`s like you did).

Core/clim-core/gadgets.lisp
```diff
@@ -592,6 +593,11 @@ and must never be nil.")
 (defmethod (setf check-box-current-selection) (new-value (check-box check-box))
   (setf (gadget-value check-box) new-value))
 
+(defmethod check-box-selections ((pane check-box))
+  (loop for child in (sheet-children pane)
+        when (and (typep child 'toggle-button) (gadget-value child))
+        collect child))
```
scymtym: Like above, I think this should just return all buttons.

Core/clim-core/gadgets.lisp
pouar: So returning all buttons regardless of gadget-value?

pouar: Decided to push a commit doing this while I wait for an answer. I can always reset back to the previous commit if I'm wrong.

pouar: afaik it's the master branch you want to avoid force pushing to

Core/clim-core/gadgets.lisp
pouar: this good?

Core/clim-core/gadgets.lisp
scymtym: > So returning all buttons regardless of gadget-value?

That's how I read it and @dkochmanski  agrees.

> afaik it's the master branch you want to avoid force pushing to

I guess it depends on the project but we absolutely rewrite history in pull request branches.

Core/clim-core/gadgets.lisp
scymtym: Yes. Though, we usually indicate the changed subsystem (using the term loosely here). For example: "gadgets: implement radio-box-selections …".

Can we ask you to tweak this once more and also squash the second commit into the first?

pouar: like this?

scymtym:
Applied in master, thank you.

PR #920, opened 2019-12-14
Listener: open pathnames with #\Space

admich:
If in a pathname there is a #\Space character it is necessary to
escape it with a backslash when the view command string is generated.

----

#


dkochmanski:
lgtm, thanks!

PR #921, opened 2019-12-15
Preparation for automatic frame instance update on frame class redefinition

scymtym:
This pull request introduces a mechanism for updating existing frame instances when their frame class is redefined. Documentation and a new example are also included.

### DIFF
--- a/Core/clim-basic/utils.lisp
+++ b/Core/clim-basic/utils.lisp
@@ -466,6 +466,24 @@ in KEYWORDS removed."
       (declare (dynamic-extent #',cont))
       (,fun ,@to-bind #',cont ,@to-pass))))
 
+;;; Macro writing
+
+;;; This is a utility which is intended to be used by macro writers to
+;;; help application programmers. It allows the CL implementation to
+;;; associate conditions signaled during macroexpansion with sub-forms
+;;; of the whole macro form. In practice this means tools like SLIME
+;;; will highlight that sub-form when reporting the error. This macro
+;;; is most useful in complex macros such as
+;;; `define-application-frame'.
+(defmacro with-current-source-form ((form &rest more-forms) &body body)
+  "Associate errors signaled while executing BODY with FORM.
+MORE-FORMS are used when FORM is unsuitable (for example some
+implementations cannot associate errors with atoms). MORE-FORMS should
+be forms containing FORM."
+  #-sbcl (declare (ignore form more-forms))
+  #+sbcl `(sb-ext:with-current-source-form (,form ,@more-forms) ,@body)
+  #-sbcl `(progn ,@body))
+
 ;;;; ----------------------------------------------------------------------
 
 ;;; FIXME valid space specification format is described in the section
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -751,7 +751,7 @@ documentation produced by presentations.")
   (setf (slot-value pane 'name) name)
   pane)
 
-(defun do-pane-creation-form (name form)
+(defun generate-pane-creation-form (name form)
   (destructuring-bind (pane &rest options) form
     (cond ((and (null options) (listp pane)) ; Single form which is a function call
            `(coerce-pane-name ,pane ',name))
@@ -766,8 +766,8 @@ documentation produced by presentations.")
           (t ; Non-standard pane designator fed to the `make-pane'
            `(make-pane ',pane :name ',name ,@options)))))
 
-(defun make-panes-generate-panes-form (class-name menu-bar panes layouts
-                                       pointer-documentation)
+(defun generate-generate-panes-form (class-name menu-bar panes layouts
+                                 pointer-documentation)
   (when pointer-documentation
     (setf panes (append panes
                         '((%pointer-documentation%
@@ -778,45 +778,38 @@ documentation produced by presentations.")
          (unless (frame-panes-for-layout frame)
            (setf (frame-panes-for-layout frame)
                  (list
-                  ,@(loop
-                       for (name . form) in panes
-                       collect
-                         `(cons ',name ,(do-pane-creation-form name form))))))
-         (let ,(loop
-                  for (name . form) in panes
-                  collect `(,name (alexandria:assoc-value
-                                   (frame-panes-for-layout frame)
-                                   ',name :test #'eq)))
+                  ,@(loop for (name . form) in panes
+                          collect `(cons ',name ,(generate-pane-creation-form
+                                                  name form))))))
+         (let ,(loop for (name . form) in panes
+                     collect `(,name (alexandria:assoc-value
+                                      (frame-panes-for-layout frame)
+                                      ',name :test #'eq)))
            ;; [BTS] added this, but is not sure that this is correct for
            ;; adding a menu-bar transparently, should also only be done
            ;; where the exterior window system does not support menus
-           ,(if (or menu-bar pointer-documentation)
-                `(setf (frame-panes frame)
-                       (ecase (frame-current-layout frame)
-                         ,@(mapcar (lambda (layout)
-                                     `(,(first layout)
-                                        (vertically ()
-                                          ,@(cond
-                                              ((eq menu-bar t)
-                                               `((setf (frame-menu-bar-pane frame)
-                                                       (clim-internals::make-menu-bar
-                                                        ',class-name))))
-                                              ((consp menu-bar)
-                                               `((clim-internals::make-menu-bar
-                                                  (make-command-table
-                                                   nil
-                                                   :menu ',menu-bar))))
-                                              (menu-bar
-                                               `((clim-internals::make-menu-bar
-                                                  ',menu-bar)))
-                                              (t nil))
-                                          ,@(rest layout)
-                                          ,@(when pointer-documentation
-                                              '(%pointer-documentation%)))))
-                                   layouts)))
-                `(setf (frame-panes frame)
-                       (ecase (frame-current-layout frame)
-                         ,@layouts)))))
+           (setf (frame-panes frame)
+                 (ecase (frame-current-layout frame)
+                   ,@(if (or menu-bar pointer-documentation)
+                         (mapcar (lambda (layout)
+                                   `(,(first layout)
+                                     (vertically ()
+                                       ,@(cond
+                                           ((eq menu-bar t)
+                                            `((setf (frame-menu-bar-pane frame)
+                                                    (make-menu-bar ',class-name))))
+                                           ((consp menu-bar)
+                                            `((make-menu-bar
+                                               (make-command-table
+                                                nil :menu ',menu-bar))))
+                                           (menu-bar
+                                            `((make-menu-bar ',menu-bar)))
+                                           (t nil))
+                                       ,@(rest layout)
+                                       ,@(when pointer-documentation
+                                           '(%pointer-documentation%)))))
+                                 layouts)
+                         layouts)))))
        ;; Update frame-current-panes and the special pane slots.
        (update-frame-pane-lists frame))))
 
@@ -930,7 +923,7 @@ documentation produced by presentations.")
           ,@user-default-initargs)
          ,@other-options)
 
-       ,(make-panes-generate-panes-form
+       ,(generate-generate-panes-form
          name menu-bar panes layouts pointer-documentation)
 
        ,@(when command-table
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -742,8 +742,7 @@ documentation produced by presentations.")
                        :name ',name
                        ,@(cdr form)))
        (:interactor `(make-clim-interactor-pane
-                      :name ',name ,@(cdr form)
-                      ,@(cdr form)))
+                      :name ',name ,@(cdr form)))
        (:pointer-documentation `(make-clim-pointer-documentation-pane
                                  :name ',name
                                  ,@(cdr form)))
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -252,6 +252,9 @@ documentation produced by presentations.")
   ;; Let client code know.
   (clime:note-frame-pretty-name-changed (frame-manager frame) frame new-value))
 
+(defmethod frame-all-layouts ((frame application-frame))
+  (mapcar #'car (frame-layouts frame)))
+
 (define-condition frame-layout-changed (condition)
   ((frame :initarg :frame :reader frame-layout-changed-frame)))
 
@@ -725,6 +728,25 @@ documentation produced by presentations.")
 ;
 ; FIXME
 
+(defun update-frame-pane-lists (frame)
+  (let ((all-panes     (frame-panes frame))
+        (named-panes   (mapcar #'cdr (frame-panes-for-layout frame)))
+        (current-panes '()))
+    ;; Find intersection of named panes and current layout panes.
+    (map-over-sheets (lambda (sheet)
+                       (when-let ((index (position sheet named-panes)))
+                         (push (cons sheet index) current-panes)))
+                     all-panes)
+    (setf current-panes (mapcar #'car (sort current-panes #'< :key #'cdr)))
+    ;; Populate current-pane list and special pane slots.
+    (let ((interactor            (find-pane-of-type current-panes 'interactor-pane))
+          (application           (find-pane-of-type current-panes 'application-pane))
+          (pointer-documentation (find-pane-of-type all-panes 'pointer-documentation-pane)))
+      (setf (frame-current-panes frame) current-panes
+            (frame-standard-output frame) (or application interactor)
+            (frame-standard-input frame) (or interactor (frame-standard-output frame))
+            (frame-pointer-documentation-output frame) pointer-documentation))))
+
 (defun coerce-pane-name (pane name)
   (setf (slot-value pane 'name) name)
   pane)
@@ -795,37 +817,8 @@ documentation produced by presentations.")
                 `(setf (frame-panes frame)
                        (ecase (frame-current-layout frame)
                          ,@layouts)))))
-
-       ;; XXX: this computation may be cached for each layout!
-       (let ((named-panes (mapcar #'cdr (frame-panes-for-layout frame)))
-             (panes nil))
-
-         ;; Find intersection of named panes and current layout panes
-         (map-over-sheets #'(lambda (p)
-                              (when (member p named-panes)
-                                (push p panes)))
-                          (frame-panes frame))
-
-         (setf (frame-current-panes frame)
-               (uiop:while-collecting (sorted-panes)
-                 (mapc #'(lambda (pane)
-                           (when (member pane panes)
-                             ;; collect pane
-                             (sorted-panes pane)
-                             ;; reduce search time
-                             (setf panes (delete pane panes))))
-                       named-panes)))
-
-         (setf (frame-standard-output frame)
-               (or (find-pane-of-type (frame-current-panes frame) 'application-pane)
-                   (find-pane-of-type (frame-current-panes frame) 'interactor-pane))
-
-               (frame-standard-input frame)
-               (or (find-pane-of-type (frame-current-panes frame) 'interactor-pane)
-                   (frame-standard-output frame))
-
-               (frame-pointer-documentation-output frame)
-               (find-pane-of-type (frame-panes frame) 'pointer-documentation-pane))))))
+       ;; Update frame-current-panes and the special pane slots.
+       (update-frame-pane-lists frame))))
 
 (defmacro define-application-frame (name superclasses slots &rest options)
   (when (null superclasses)
@@ -899,9 +892,6 @@ documentation produced by presentations.")
          ,@geometry
          ,@user-default-initargs)
         ,@others)
-
-      (defmethod frame-all-layouts ((frame ,name))
-        ',(mapcar #'car layouts))
 
       ,(make-panes-generate-panes-form name menu-bar panes layouts
                                        pointer-documentation)
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -820,96 +820,129 @@ documentation produced by presentations.")
        ;; Update frame-current-panes and the special pane slots.
        (update-frame-pane-lists frame))))
 
+(defun parse-define-application-frame-options (options)
+  (let ((infos '(;; CLIM
+                 (:pane                  * :conflicts (:panes :layouts))
+                 (:panes                 * :conflicts (:pane))
+                 (:layouts               * :conflicts (:pane))
+                 (:command-table         1)
+                 (:command-definer       1)
+                 (:menu-bar              ensure-1)
+                 (:disabled-commands     *)
+                 (:top-level             1)
+                 ;; :icon is the CLIM specification but we don't support it
+                 (:geometry              *)
+                 ;; :resize-frame is mentioned in a spec annotation but we don't support it
+                 ;; McCLIM extensions
+                 (:pointer-documentation 1)
+                 ;; Default initargs
+                 (:pretty-name           1)
+                 ;; Common Lisp
+                 (:default-initargs      *)))
+        (all-values '()))
+    (labels ((definedp (key)
+               (not (eq (getf all-values key 'undefined) 'undefined)))
+             (parse-option (key values)
+               (when-let ((info (find key infos :key #'first)))
+                 (destructuring-bind (name value-count &key conflicts) info
+                   (declare (ignore name))
+                   (cond ((when-let ((other (find-if #'definedp conflicts)))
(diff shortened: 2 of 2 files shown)


Core/clim-core/frames.lisp
```diff
@@ -725,6 +728,29 @@ documentation produced by presentations.")
...
+(defun update-frame-pane-lists (frame)
+  (let ((all-panes     (frame-panes frame))
+        (named-panes   (mapcar #'cdr (frame-panes-for-layout frame)))
+        (current-panes '()))
+    ;; Find intersection of named panes and current layout panes.
+    (map-over-sheets (lambda (sheet)
+                       (when-let ((index (position sheet named-panes)))
+                         (push (cons sheet index) current-panes)))
+                     all-panes)
+    (setf current-panes (map 'list #'car (sort current-panes #'< :key #'cdr)))
```
dkochmanski: mapcar #'car ?

Core/clim-core/frames.lisp
```diff
@@ -725,6 +728,29 @@ documentation produced by presentations.")
...
+    ;; Populate current-pane list and special pane slots.
+    (let ((interactor (find-pane-of-type current-panes 'interactor-pane)))
```
dkochmanski: is there a reason for preemptively find interactor pane, but not application and pointer-documentation panes? if not, please add (application (find-pane-of-type … 'application-pane)) and (pointer-doc …) and use them accordingly.

Core/clim-basic/utils.lisp
```diff
@@ -466,6 +466,17 @@ in KEYWORDS removed."
...
+
+(defmacro with-current-source-form ((form &rest more-forms) &body body)
+  "Associate errors signaled while executing BODY with FORM.
+MORE-FORMS are used when FORM is unsuitable (for example some
+implementations cannot associate errors with atoms). MORE-FORMS should
+be forms containing FORM."
+  #-sbcl (declare (ignore form more-forms))
+  #+sbcl `(sb-ext:with-current-source-form (,form ,@more-forms) ,@body)
+  #-sbcl `(progn ,@body))
+
```
dkochmanski: I'm not thrilled with an utility tailored specifically for a single implementation (with a stub for everything else). I gather it simplifies working with macros and I'll leave it be, but for runtime functionality we should discuss a policy of inclusion of implementation-specific magic. This remark may be considered as "resolved".

Core/clim-core/frames.lisp
```diff
@@ -830,96 +830,131 @@ documentation produced by presentations.")
...
+  (let ((infos '(;; CLIM
+                 (:pane                  * :conflicts (:panes :layouts))
+                 (:panes                 * :conflicts (:pane))
+                 (:layouts               * :conflicts (:pane))
+                 (:command-table         1)
+                 (:command-definer       1)
+                 (:menu-bar              ensure-1)
+                 (:disabled-commands     *)
+                 (:top-level             1)
+                 ;; icon
```
dkochmanski: what this comment stands for?

Core/clim-core/frames.lisp
```diff
@@ -830,96 +830,131 @@ documentation produced by presentations.")
...
+                 (:geometry              *)
+                 ;; resize-frame
```
dkochmanski: ditto

Core/clim-core/frames.lisp
```diff
@@ -830,96 +830,131 @@ documentation produced by presentations.")
...
+                              (parse-option :pretty-name (list user-pretty-name))))
+                          (setf (getf all-values :user-default-initargs)
+                                (alexandria:remove-from-plist values :pretty-name)))
+                         (t
+                          (setf (getf all-values key)
+                                (ecase value-count
+                                  (1        (first values))
+                                  (ensure-1 (if (listp values)
+                                                (first values)
+                                                values))
```
dkochmanski: alexandria:ensure-car ?

Core/clim-core/frames.lisp
```diff
@@ -755,46 +756,60 @@ documentation produced by presentations.")
...
+             (:interactor `(make-clim-interactor-pane
+                            :name ',name ,@(cdr form)))
+             (:pointer-documentation `(make-clim-pointer-documentation-pane
+                                       :name ',name
+                                       ,@(cdr form)))
+             (:command-menu `(make-clim-command-menu-pane
+                              :name ',name ,@(cdr form)))
+             (otherwise (generic-make-or-reinitialize-pane))))
+          (t ; Non-standard pane designator fed to the `make-pane'
+           (generic-make-or-reinitialize-pane)))))
```
dkochmanski: I've provided earlier a remark about replacing cond with if and merging `t` and `otherwise` clause in `do-pane-creation-form`, it applies here too.

Core/clim-core/frames.lisp
```diff
@@ -936,10 +960,15 @@ documentation produced by presentations.")
                                 ,@(cdr top-level)))
           ,@geometry
           ,@user-default-initargs)
+         ,@(cond (metaclass
+                  `((:metaclass ,metaclass)))
+                 (update-instances-on-redefinition
+                  `((:metaclass redefinition-updates-instances-class))))
```
dkochmanski: wouldn't it be more straightforward to document the metaclass and provide an example for it (instead of adding a custom option to define-application-pane)?

Core/clim-core/redefinition.lisp
```diff
@@ -0,0 +1,176 @@
...
+
+(defvar *delayed-redefinition-notifications*)
+
+(defun call-with-delayed-redefinition-notifications (thunk)
+  (if (boundp '*delayed-redefinition-notifications*)
+      (funcall thunk)
+      (let ((*delayed-redefinition-notifications* '()))
+        (funcall thunk)
+        (map nil #'handle-frame-class-redefinition
+             *delayed-redefinition-notifications*))))
```
dkochmanski: not that it matters much, but shouldn't it be `(multiple-value-prog1 (funcall thunk) (mapc …))` (or in the first arm of if `(prog1 nil (funcall thunk))`)

Core/clim-core/redefinition.lisp
```diff
@@ -0,0 +1,176 @@
...
+                                 :new-command-table (new-command-table)))))
+             (maybe-update (weak-instance)
+               ;; Update instances that haven't become garbage and
+               ;; haven't been CHANGE-CLASSed. Implicitly drop other
+               ;; instances.
+               (when-let ((live-instance (tg:weak-pointer-value weak-instance)))
+                 (when (typep live-instance class)
+                   (push weak-instance remaining)
+                   (update live-instance)))))
+      (map nil #'maybe-update (instances class)))
```
dkochmanski: mapc?

Core/clim-core/frames.lisp
```diff
@@ -755,46 +756,60 @@ documentation produced by presentations.")
...
-    (t `(make-pane ',(first form) :name ',name ,@(cdr form)))))
+(defun %generic-make-or-reinitialize-pane
+    (panes-for-layout abstract-class-name name &rest initargs)
+  (if-let ((pane (alexandria:assoc-value
+                  panes-for-layout name :test #'eq)))
+    (progn
+      (when-let ((parent (sheet-parent pane)))
+        (sheet-disown-child parent pane))
+      (apply #'reinitialize-instance pane initargs))
+    (apply #'make-pane abstract-class-name :name name initargs)))
```
dkochmanski: is there a reason why this function body is not inlined in generic-make-or-reinitialize-pane in flet later?

Core/clim-core/frames.lisp
```diff
@@ -755,46 +756,60 @@ documentation produced by presentations.")
...
        (with-look-and-feel-realization (fm frame)
-         (unless (frame-panes-for-layout frame)
+         ;; Make (or reinitialize) pane instances and establish
+         ;; lexical variables so layout forms can use them.
+         (let* (,@(when reinitializep
+                    `((old-panes (frame-panes-for-layout frame))))
+                ,@(loop for (name . form) in panes
+                        collect `(,name ,(make-ensure-pane-form
+                                          name form
+                                          (when reinitializep 'old-panes)))))
```
dkochmanski: `(and reinitializep 'old-panes)`

I think that passing the variable name symbol like this makes the code confusing.

Core/clim-core/frames.lisp
```diff
@@ -755,46 +756,60 @@ documentation produced by presentations.")
...
+    (panes-for-layout abstract-class-name name &rest initargs)
+  (if-let ((pane (alexandria:assoc-value
+                  panes-for-layout name :test #'eq)))
+    (progn
+      (when-let ((parent (sheet-parent pane)))
+        (sheet-disown-child parent pane))
+      (apply #'reinitialize-instance pane initargs))
+    (apply #'make-pane abstract-class-name :name name initargs)))
+
+(defun make-ensure-pane-form (name form panes-for-layout-var)
```
dkochmanski: do we have a convention for functions which are used as helpers for macros? `do-pane-*` doesn't seem like a good convention either, but `make-*` seems to indicate that it is a constructor.

Core/clim-core/frames.lisp
```diff
@@ -739,17 +739,13 @@ documentation produced by presentations.")
...
-                      ,@(cdr form)))
+                      :name ',name ,@(cdr form)))
        (:pointer-documentation `(make-clim-pointer-documentation-pane
-                                 :name ',name
-                                 ,@(cdr form)))
+                                 :name ',name ,@(cdr form)))
        (:command-menu `(make-clim-command-menu-pane
-                                :name ',name
-                                ,@(cdr form)))
+                                :name ',name ,@(cdr form)))
```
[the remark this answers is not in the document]
scymtym: Changed.

Core/clim-core/frames.lisp
```diff
@@ -739,17 +739,13 @@ documentation produced by presentations.")
...
        (otherwise `(make-pane ,(first form) :name ',name ,@(cdr form)))))
     ;; Non-standard pane designator fed to the `make-pane'
     (t `(make-pane ',(first form) :name ',name ,@(cdr form)))))
```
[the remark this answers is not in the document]
scymtym: Changed. There is now a separate commit for cosmetic changes in `do-pane-creation-form`.

Core/clim-core/frames.lisp
```diff
@@ -252,6 +252,9 @@ documentation produced by presentations.")
   ;; Let client code know.
   (clime:note-frame-pretty-name-changed (frame-manager frame) frame new-value))
 
+(defmethod frame-all-layouts ((frame application-frame))
+  (map 'list #'car (frame-layouts frame)))
```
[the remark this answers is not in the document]
scymtym: I use `map` when I'm treating the argument as a sequence, not specifically as a list (that is, in contrast to  `mapcan`, `mapl` and friends).

If we want to enforce a particular policy with these matter-of-taste-ish things (line breaking, helper functions and `&aux` come to mind), we should write a coding style document instead of discussing this in each pull request.

Core/clim-core/frames.lisp
scymtym: The `interactor` variable is used in multiple places which is not the case for the results of the other `find-pane-of-type` results

Core/clim-basic/utils.lisp
scymtym: > I'm not thrilled with an utility tailored specifically for a single implementation (with a stub for everything else). I gather it simplifies working with macros and I'll leave it be

The macro simplifies working with macros for the application programmer, not the macro writer. Specifically, it allows debugging invalid uses of macros more easily by allowing SLIME or other tools to highlight the sub-form of the whole macro form that is the cause of the problem. I will rewrite the commit message to state this more clearly.

In my opinion, more implementations should provide a similar functionality since it is very useful in programming interfaces with complex macros (other examples include `defclass` and `defpackage`). Therefore I hope it doesn't remain as implementation-specific. That said, I would like to try the feature in practice some more before writing a CDR. One piece of feedback I already got from your comment is that its purpose is not obvious without further explanation.

> but for runtime functionality we should discuss a policy of inclusion of implementation-specific magic.

I don't see how the distinction between macroexpansion and runtime is significant in this regard. Also, there is no magic involved. If the purpose of the change is unclear I simply have to explain it better.

> This remark may be considered as "resolved".

Sorry to put this bluntly, but then don't write it. I'm happy to answer questions regarding the change or acknowledge or address pointed out problems, but expressing dissatisfaction followed by "but it doesn't matter" seems inappropriate for a code review.

Core/clim-core/frames.lisp
scymtym: `:icon` is mentioned as a valid option in the spec entry for `define-application-frame` but we don't seem to support it.

Core/clim-core/frames.lisp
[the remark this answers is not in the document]
dkochmanski: I've asked for line break after the keyword, that is:
```
(:application
  `(make-clim-application-pane :name ',name ,@(cdr form))
```

etc

Core/clim-core/frames.lisp
scymtym: An annotation for the spec entry for `define-application-frame` states that this should probably be an accepted option, but we don't seem to support it.

scymtym: I added a comment explaining this to the commit.

Core/clim-core/frames.lisp
scymtym: I added a comment explaining this to the commit.

Core/clim-core/frames.lisp
scymtym: Changed.

Core/clim-core/frames.lisp
scymtym: Changed (across multiple commits, so a bit hard to keep track of).

Core/clim-core/frames.lisp
scymtym: From my point of the view, the metaclass is an implementation detail. Ideally, we would figure out a way to allow a custom metaclass and `:update-instances-on-redefinition` at the same time.

Core/clim-core/frames.lisp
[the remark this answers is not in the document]
dkochmanski: I'm asking, because you've changed the call which was used in a macro expansion of the same method. I.e is there a sudden reason that frame-layouts returns a vector (not always a list)? if not, then using map removes this semantic information from this code (that frame-layouts returns specifically a list).

In the past we were discussing "matter of taste" things in pull requests number of times; proposing a coding style document makes sense of course, but I don't think it makes this remark "sudden" in this pull request.

Core/clim-core/redefinition.lisp
scymtym: Yes, I think your suggestion makes it better. Changed. Although the only user, `define-application-frame` doesn't seem to have a specified return value.

Core/clim-core/frames.lisp
dkochmanski: We look for each pane anyway, so there is no harm of putting them in LET and it (matter-of-taste-speaking) makes the code more readable.

Core/clim-basic/utils.lisp
dkochmanski: macroexpansion time and runtime are different in a sense that I would ask you to remove the code if it had introduced a feature for applications  which works only on a single implementation (i.e if you had introduced automatic frame instance update which were working only on clisp).

Point taken, next time I'll either hit "cancel" button or ask for a change.

Core/clim-core/frames.lisp
scymtym: It is a significant chunk that would be repeated for each pane in the expansion. I see multiple arguments in favor of having a separate function:

* Less code in the expander function
* Less code in the expansion
  * Easier to debug
  * Quicker compilation
* (small advantage) Changes to the function can take effect without re-expanding all `define-application-frame` forms

Core/clim-core/frames.lisp
scymtym: `make-…-form` seems fine to me but I see your point. Would you prefer `generate-…-form`?

Core/clim-core/frames.lisp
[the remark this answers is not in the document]
scymtym: I didn't (force-)push the accumulated changes yet. I did what you describe.

Core/clim-core/frames.lisp
scymtym: Right, will do.

Core/clim-core/frames.lisp
dkochmanski: generate makes more sense, yes

scymtym: Changed.

Core/clim-core/frames.lisp
scymtym: Changed.

dkochmanski:
testing: when I redefine the frame class in the example frame-class-redefinition by changing the layout order (suggested in a readme) I have an error:

```
  #<STANDARD-GENERIC-FUNCTION (COMMON-LISP:SETF CLIM-INTERNALS::PRESENTATION-HISTORY-POINTER) (1)>
when called with arguments
  (NIL NIL).
   [Condition of type SB-PCL::NO-APPLICABLE-METHOD-ERROR]
See also:
  Common Lisp Hyperspec, 7.6.6 [:section]

Restarts:
 0: [RETRY] Retry calling the generic function.
 1: [ABORT] Return to application command loop.
 2: [ABORT] Return to application command loop.
 3: [RETRY] Retry SLIME REPL evaluation request.
 4: [*ABORT] Return to SLIME's top level.
 5: [ABORT] abort thread (#<THREAD "repl-thread" RUNNING {10041F9BC3}>)

Backtrace:
  0: ((:METHOD NO-APPLICABLE-METHOD (T)) #<STANDARD-GENERIC-FUNCTION (COMMON-LISP:SETF CLIM-INTERNALS::PRESENTATION-HISTORY-POINTER) (1)> NIL NIL) [fast-method]
  1: (SB-PCL::CALL-NO-APPLICABLE-METHOD #<STANDARD-GENERIC-FUNCTION (COMMON-LISP:SETF CLIM-INTERNALS::PRESENTATION-HISTORY-POINTER) (1)> (NIL NIL))
  2: ((FLET "CLEANUP-FUN-13" :IN CLIM:ACCEPT)) [cleanup]
  3: (CLIM:ACCEPT (CLIM:COMMAND :COMMAND-TABLE #<CLIM:STANDARD-COMMAND-TABLE CLIM-DEMO::FRAME-CLASS-REDEFINITION {1005B3E553}>) :STREAM #<CLIM-CLX::CLX-825-INTERACTOR-PANE CLIM-DEMO::INTERACTOR {10147BB783..
  4: (CLIM:READ-COMMAND #<CLIM:STANDARD-COMMAND-TABLE CLIM-DEMO::FRAME-CLASS-REDEFINITION {1005B3E553}> :STREAM #<CLIM-CLX::CLX-825-INTERACTOR-PANE CLIM-DEMO::INTERACTOR {10147BB783}> :COMMAND-PARSER NIL :..
  5: (CLIM:READ-COMMAND-USING-KEYSTROKES #<CLIM:STANDARD-COMMAND-TABLE CLIM-DEMO::FRAME-CLASS-REDEFINITION {1005B3E553}> NIL :STREAM #<CLIM-CLX::CLX-825-INTERACTOR-PANE CLIM-DEMO::INTERACTOR {10147BB783}> ..
  6: (CLIM:READ-COMMAND #<CLIM:STANDARD-COMMAND-TABLE CLIM-DEMO::FRAME-CLASS-REDEFINITION {1005B3E553}> :STREAM #<CLIM-CLX::CLX-825-INTERACTOR-PANE CLIM-DEMO::INTERACTOR {10147BB783}> :COMMAND-PARSER NIL :..
  7: ((:METHOD CLIM:READ-FRAME-COMMAND :AROUND (CLIM:APPLICATION-FRAME)) #<CLIM-DEMO::FRAME-CLASS-REDEFINITION {1014324C43}> :STREAM #<CLIM-CLX::CLX-825-INTERACTOR-PANE CLIM-DEMO::INTERACTOR {10147BB783}>)..
  8: ((:METHOD CLIM:DEFAULT-FRAME-TOP-LEVEL (CLIM:APPLICATION-FRAME)) #<CLIM-DEMO::FRAME-CLASS-REDEFINITION {1014324C43}> :COMMAND-PARSER CLIM:COMMAND-LINE-COMMAND-PARSER :COMMAND-UNPARSER CLIM:COMMAND-LIN..
  9: ((:METHOD CLIM:RUN-FRAME-TOP-LEVEL (CLIM:APPLICATION-FRAME)) #<CLIM-DEMO::FRAME-CLASS-REDEFINITION {1014324C43}>) [fast-method]
 10: ((:METHOD CLIM:RUN-FRAME-TOP-LEVEL :AROUND (CLIM:APPLICATION-FRAME)) #<CLIM-DEMO::FRAME-CLASS-REDEFINITION {1014324C43}>) [fast-method]
 11: ((LAMBDA (&REST IGNORE) :IN CLIM-DEMO::MAKE-DEMO-BUTTON))
 12: (CLIM-INTERNALS::INVOKE-CALLBACK #<CLIM-CLX::CLX-825-PUSH-BUTTON-PANE NIL {1008A0A0F3}> #<CLOSURE (LAMBDA (&REST IGNORE) :IN CLIM-DEMO::MAKE-DEMO-BUTTON) {1008A0948B}>)
 13: ((:METHOD CLIM:HANDLE-EVENT (CLIM:PUSH-BUTTON-PANE CLIM:POINTER-BUTTON-RELEASE-EVENT)) #<CLIM-CLX::CLX-825-PUSH-BUTTON-PANE NIL {1008A0A0F3}> #<unused argument>) [fast-method]
 14: (CLIM-EXTENSIONS:SIMPLE-EVENT-LOOP #<CLIM-DEMO:DEMODEMO {1005FC1463}>)
 15: ((:METHOD CLIM:READ-FRAME-COMMAND :AROUND (CLIM:APPLICATION-FRAME)) #<CLIM-DEMO:DEMODEMO {1005FC1463}> :STREAM NIL) [fast-method]
 16: ((:METHOD CLIM:DEFAULT-FRAME-TOP-LEVEL (CLIM:APPLICATION-FRAME)) #<CLIM-DEMO:DEMODEMO {1005FC1463}> :COMMAND-PARSER CLIM:COMMAND-LINE-COMMAND-PARSER :COMMAND-UNPARSER CLIM:COMMAND-LINE-COMMAND-UNPARSE..
 17: ((:METHOD CLIM:RUN-FRAME-TOP-LEVEL (CLIM:APPLICATION-FRAME)) #<CLIM-DEMO:DEMODEMO {1005FC1463}>) [fast-method]
 18: ((:METHOD CLIM:RUN-FRAME-TOP-LEVEL :AROUND (CLIM:APPLICATION-FRAME)) #<CLIM-DEMO:DEMODEMO {1005FC1463}>) [fast-method]
 19: (SB-INT:SIMPLE-EVAL-IN-LEXENV (CLIM-DEMO:DEMODEMO) #<NULL-LEXENV>)
 20: (EVAL (CLIM-DEMO:DEMODEMO))
```

Core/clim-core/frames.lisp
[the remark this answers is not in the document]
dkochmanski: please use list operators where we deal with lists (as you have proclaimed in the first paragraph of the answer).

Core/clim-core/frames.lisp
dkochmanski: please use list operators where we deal with lists

Core/clim-core/redefinition.lisp
dkochmanski: result is ignored so please use mapc.

Core/clim-core/frames.lisp
dkochmanski: - please use gensym here for `old-panes`, otherwise (however unlikely) pane name may be `old-panes`

IMO this let* bindings would be more readable if written in this spirit:

```
,@(if reinitializep
      (with-gensyms (gvar)
        (list* `(,gvar (frame-panes-for-layout frame))
               (loop for (name . form) in panes
                     collect `(,name ,(generate-ensure-pane-form
                                       name form gvar)))))
      (loop for (name . form) in panes
            collect `(,name ,(generate-ensure-pane-form name form nil))))
```

Core/clim-core/frames.lisp
dkochmanski: I will keep this discussion open to think about the "ideal" way you have mentioned when I dig deeper into the code.

Core/clim-core/redefinition.lisp
scymtym: This change will be in the second pull request.

Core/clim-core/frames.lisp
scymtym: This change will be in the second pull request.

Core/clim-core/frames.lisp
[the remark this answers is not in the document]
scymtym: Changed.

Core/clim-core/frames.lisp
scymtym: Changed.

PR #922, opened 2019-12-16
Listener: Add command COM-SHOW-CLASS-DOCUMENTATION.

duuqnd:
It can often be useful to be able to quickly get the documentation of
a class. This command takes a class name and prints the documentation
string if one is available. It also properly handles cases where either
the class does not exist or the documentation string does not exist.


dkochmanski:
lgtm, thanks!

PR #923, opened 2019-12-18
Fix history

heshrobe:
The default method for presentation-type-history has a bug for types without a history.  The case statement had a clause (nil nil) which is incorrect.  This macro expands to something that never matches.  The result is that the last clause matches and calls itself again.  The fix is that the clause should be ((nil) nil)


Core/clim-core/commands.lisp
```diff
@@ -223,6 +223,29 @@ designator) inherits menu items."
 
 (make-command-table 'user-command-table)
 
+<<<<<<< HEAD
```
dkochmanski: this is invalid syntax for common lisp (probably a result of merge conflicts). generally speaking please rather rebase pull requests instead of merging them - that will make peer review easier.

PR #925, opened 2019-12-22
Listener: remove cl-fad dependency

admich:
----

#

### DIFF
--- a/Apps/Listener/asdf.lisp
+++ b/Apps/Listener/asdf.lisp
@@ -50,7 +50,8 @@
       (mapc (lambda (path)
 	      (when (string-equal (pathname-type path) "asd")
 		(systems (truename path))))
-	    (cl-fad:list-directory (eval reg))))))
+            (uiop:directory-files
+             (uiop:ensure-directory-pathname (eval reg)))))))
 
 (defun asdf-operation-pretty-name (op)
   (case op
--- a/Apps/Listener/clim-listener.asd
+++ b/Apps/Listener/clim-listener.asd
@@ -15,7 +15,7 @@ experimentation. Present features include:
 - Navigation of the filesystem, including a directory stack
 - Launching of external programs sensitive to file type (determined by mailcap
   and mime.types files)"
-  :depends-on (#:mcclim #:clim-debugger #:uiop #:cl-fad #+sbcl #:sb-posix)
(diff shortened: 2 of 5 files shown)


scymtym:
I'm not sure whether your are aware of this, but this part of the code has a long history (in terms of implementation variants tried).

* The initial implementation used CL pathname-related functions

* 9eb1a373 introduced a dependency on osicat (initially for environment variables)

* d4549d47, d09dd416, 3770bec0 switch pathname-related functionality to osicat

* df27be05 removed the osicat dependency again, introducing the cl-fad dependency

Since d4549d47, d09dd416 and 3770bec0 don't contain any rationale and df27be05 explains the switch *from* osicat (avoiding ffi) but not the switch *to* `cl-fad`, I think discussing alternative solutions instead of again switching from one dependency to another would be in order.

In particular, I wonder whether we could go back to the earliest solution of not relying on any dependency and having a few utility functions since that

1. Avoided the dependency

2. Provided a more powerful command which allowed e.g. `Show Directory /tmp/*/bla/**/*.txt`

I suspect this solution would have disadvantages in terms of implementation-specific behavior and/or bugs, but any change should be made based on an informed and documented decision or this might go in circles forever.

admich:
I was not aware of this history. So for the moment I close this PR. Anyway I think that this PR goes in the direction that you hope (1. Avoided the dependency): I switch from  the `cl-fad` dependency to the `uiop` dependency that  is an almost ubiquitous package because is a part of `asdf`: McCLIM already depends from `uiop`.

Also with some work using `uiop` functions I think it is possible to provide "a more powerful command which allowed e.g. Show Directory /tmp/*/bla/**/*.txt" as you suggested.

dkochmanski:
We should weed out usage of uiop from McCLIM (except from asd definitions), not depend on it more. It adds complexity instead of removing it.

scymtym:
@admich I didn't mean to discourage you from working on this or proposing a solution, sorry.

> We should weed out usage of uiop from McCLIM (except from asd definitions), not depend on it more. It adds complexity instead of removing it.

@jackdaniel I don't follow the complexity argument, but that discussion is for another day. In any case, decisions like this (that is using uiop or not) should be made after a proper discussion and then documented including the rationale.

admich:
@scymtym no problem, don't worry. You didn't discourage me.

PR #931, opened 2020-01-06
make-pane-1: add a new protocol to resolve abstract pane names

dkochmanski:
Until now each backend had their own ad-hoc method. Now we introduce a
common protocol which unifies behavior and allows other clients to
resolve a pane-type to a concrete class. New function is called
find-concrete-pane-class and is specialized on a frame-manager.

This approach saves us some mess in clx's frame-manager
implementation.

### DIFF
--- a/Backends/CLX-fb/frame-manager.lisp
+++ b/Backends/CLX-fb/frame-manager.lisp
@@ -37,10 +37,10 @@
         (setf concrete-pane-class (find-class class-symbol)))))
   concrete-pane-class)
 
-(defmethod make-pane-1 ((fm clx-fb-frame-manager) (frame application-frame) type &rest args)
-  (apply #'make-instance
-	 (maybe-mirroring fm (clim-clx::find-concrete-pane-class type))
-	 :frame frame
-	 :manager fm
-	 :port (port frame)
-	 args))
+(defmethod find-concrete-pane-class ((fm clx-fb-frame-manager)
+                                     pane-type &optional (errorp t))
+  ;; This backend doesn't have any specialized pane implementations
+  ;; but depending on circumstances it may add optional mirroring to
+  ;; the class. Such automatically defined concrete class has the same
+  ;; name but with a gensym prefix and symbol in the backend package.
+  (maybe-mirroring fm (find-concrete-pane-class t pane-type errorp)))
--- a/Backends/CLX/frame-manager.lisp
+++ b/Backends/CLX/frame-manager.lisp
@@ -94,12 +94,6 @@
         (setf concrete-pane-class (find-class class-symbol)))))
   concrete-pane-class)
 
-(defmethod make-pane-1 ((fm clx-frame-manager) (frame application-frame) type &rest args)
-  (apply #'make-instance (find-concrete-pane-class fm type)
-	 :frame frame :manager fm :port (port frame)
-         args))
-
-
 (defmethod adopt-frame :before ((fm clx-frame-manager) (frame menu-frame))
   ;; Temporary kludge.
   (when (eq (slot-value frame 'climi::top) nil)
--- a/Backends/CLX/frame-manager.lisp
+++ b/Backends/CLX/frame-manager.lisp
@@ -58,56 +58,20 @@
 
 ;; Abstract pane lookup logic
 
-(defun find-first-defined-class (types)
-  (first
-   (remove-if #'null
-              (mapcar (lambda (class-name)
-                        (find-class class-name nil))
-                      types))))
+(defmethod find-concrete-pane-class ((fm clx-frame-manager)
+                                     pane-type &optional errorp)
+  ;; This backend doesn't have any specialized pane implementations
+  ;; but depending on circumstances it may add optional mirroring to
+  ;; the class. Such automatically defined concrete class has the same
+  ;; name but with a gensym prefix and symbol in the backend package.
+  (declare (ignore errorp))
+  (maybe-mirroring fm (call-next-method)))
 
-(defun find-symbol-from-spec (package-spec name-components)
-  (flet ((coerce-name-element (name-elt)
-           (typecase name-elt
-             (symbol (symbol-name name-elt))
-             (sequence (coerce name-elt 'string))
-             (t (princ-to-string name-elt)))))
-  (find-symbol
-   (apply #'concatenate 'string (mapcar #'coerce-name-element name-components))
-   package-spec)))
-
-(defun find-symbols (name-specs)
-  (remove-if #'null (mapcar #'(lambda (x) (find-symbol-from-spec (first x) (rest x))) name-specs)))
-
-(defun generate-standard-pane-specs (type)
-  (let ((mapping (get type 'climi::concrete-pane-class-name)))
-    `((,(symbol-package mapping) ,mapping)
-      (:climi ,mapping)
-      (:climi ,type #:-pane)
-      (:climi ,type))))
-
-(defun generate-clx-pane-specs (type)
-  (append
-   `((:clim-clx #:clx- ,type #:-pane)
-     (:clim-clx #:clx- ,type)
-     (:climi #:clx- ,type #:-pane)
-     (:climi #:clx- ,type))
-   (generate-standard-pane-specs type)))
-
-(defun find-concrete-pane-class (type)
-  (if (or (eql (symbol-package type)
-               (find-package '#:clim))
-          (eql (symbol-package type)
-               (find-package '#:climi))
-          (eql (symbol-package type)
-               (find-package '#:keyword))
-	  (get type 'climi::concrete-pane-class-name))
-      (find-first-defined-class (find-symbols (generate-clx-pane-specs type)))
-      type))
-
-;;; This is an example of how make-pane-1 might create specialized instances of
-;;; the generic pane types based upon the type of the frame-manager. However, in
-;;; the CLX case, we don't expect there to be any CLX specific panes. CLX uses
-;;; the default generic panes instead.
+;;; This is an example of how make-pane-1 might create specialized
+;;; instances of the generic pane types based upon the type of the
+;;; frame-manager. However, in the CLX case, we don't expect there to
+;;; be any CLX specific panes. CLX uses the default generic panes
+;;; instead.
 (defun maybe-mirroring (fm concrete-pane-class)
   (when (funcall (mirroring-p fm) concrete-pane-class)
     (let ((concrete-pane-class-symbol (if (typep concrete-pane-class 'class)
@@ -115,7 +79,8 @@
                                           concrete-pane-class)))
       (multiple-value-bind (class-symbol foundp)
           (alexandria:ensure-symbol
-           (alexandria:symbolicate (class-gensym fm) "-" (symbol-name concrete-pane-class-symbol))
+           (alexandria:symbolicate (class-gensym fm) "-"
+                                   (symbol-name concrete-pane-class-symbol))
            :clim-clx)
         (unless foundp
           (eval
@@ -126,17 +91,13 @@
                  ,concrete-pane-class-symbol)
               ()
               (:metaclass ,(type-of (find-class concrete-pane-class-symbol))))))
-        ;; (format *debug-io* "dummy class mirror ~A: ~A~%" concrete-pane-class-symbol class-symbol)
         (setf concrete-pane-class (find-class class-symbol)))))
   concrete-pane-class)
 
 (defmethod make-pane-1 ((fm clx-frame-manager) (frame application-frame) type &rest args)
-  (apply #'make-instance
-	 (maybe-mirroring fm (find-concrete-pane-class type))
-	 :frame frame
-	 :manager fm
-	 :port (port frame)
-	 args))
+  (apply #'make-instance (find-concrete-pane-class fm type)
+	 :frame frame :manager fm :port (port frame)
+         args))
 
 
 (defmethod adopt-frame :before ((fm clx-frame-manager) (frame menu-frame))
--- a/Core/clim-core/panes.lisp
+++ b/Core/clim-core/panes.lisp
@@ -363,11 +363,47 @@ order to produce a double-click")
   (print-unreadable-object (pane sink :type t :identity t)
     (prin1 (pane-name pane) sink)))
 
+(defgeneric find-concrete-pane-class (pane-realizer pane-type &optional errorp)
+  (:documentation "Resolves abstract pane type PANE-TYPE to a concrete
+pane class. Methods defined in backends should specialize on the
+PANE-REALIZER argument. When the PANE-TYPE can't be resolved NIL is
+returned or error is signaled depending on the argument ERRORP.")
+  (:method ((realizer t) pane-type &optional (errorp t))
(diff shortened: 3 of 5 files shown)


Core/clim-core/panes.lisp
```diff
@@ -363,11 +363,38 @@ order to produce a double-click")
...
+          (or (eql symbol-package (find-package '#:clim))
+              (eql symbol-package (find-package '#:climi))
+              (eql symbol-package (find-package '#:keyword))))
+        (let* ((symbol-name (symbol-name pane-type))
+               (clim-symbol (find-symbol symbol-name '#:climi)))
+          (if-let ((mapped (get clim-symbol 'concrete-pane-class-name)))
+            (find-class mapped t)
+            (let ((proper-name (concatenate 'string symbol-name "-PANE")))
+              (or (find-class (find-symbol proper-name '#:climi) nil)
+                  (find-class (find-symbol symbol-name '#:climi) nil))))            )
```
scymtym: Right paren preceded by bogus whitespace.

Rather than relying on `(find-class nil nil)` returning `nil` (it could signal an error since `nil` is not a valid class name), please do something like

```lisp
(when-let ((symbol (find-symbol proper-name '#:climi)))
  (find-class symbol nil))
```

Core/clim-core/panes.lisp
```diff
@@ -363,11 +363,38 @@ order to produce a double-click")
...
+      (return-from find-concrete-pane-class (find-class mapped t)))
+    (if (let ((symbol-package (symbol-package pane-type)))
+          (or (eql symbol-package (find-package '#:clim))
+              (eql symbol-package (find-package '#:climi))
+              (eql symbol-package (find-package '#:keyword))))
+        (let* ((symbol-name (symbol-name pane-type))
+               (clim-symbol (find-symbol symbol-name '#:climi)))
+          (if-let ((mapped (get clim-symbol 'concrete-pane-class-name)))
+            (find-class mapped t)
+            (let ((proper-name (concatenate 'string symbol-name "-PANE")))
```
scymtym: I'm not sure whether this is strictly necessary, but I always use `(string '#:-pane)` in these situations to support case-sensitive mode.

Core/clim-core/panes.lisp
```diff
@@ -363,11 +363,38 @@ order to produce a double-click")
...
+    ;; Default method tries to resolve the pane name as specified by a
+    ;; convention mentioned in the spec. Function is little
+    ;; complicated because we preserve old semantics adding rules to
+    ;; the class name resolution. Resolution works as follows:
+    ;;
+    ;; 1. Abstract mapping always takes a priority. When it exists we
+    ;;    expect that class is defined and signal error otherwise.
+    ;; 2. When the symbol is in clim/climi/keyword package:
+    ;;    - look for a class `CLIMI::{symbol-name}-PANE'
+    ;;    - look for a class `CLIMI::{symbol-name}'
```
scymtym: The emacs notation usually uses lower case for symbols and uppercase for meta-syntactic variables:

```
; `climi::SYMBOL-NAME-pane'
```

I think.

Core/clim-core/panes.lisp
```diff
@@ -363,11 +363,38 @@ order to produce a double-click")
   (print-unreadable-object (pane sink :type t :identity t)
     (prin1 (pane-name pane) sink)))
 
+(defgeneric find-concrete-pane-class (pane-realizer pane-type)
+  (:documentation "Resolves abstract pane name to a concrete pane ~
+class which may be instatinated. Function specializes on its first ~
+argument. Returns NIL if no concrete class exists.")
```
scymtym: * There are `~` characters at the end of two lines

* I would write the first sentence as "Resolve ABSTRACT-PANE-TYPE to a concrete pane class", but I guess that is another question for the coding style document.

* > Function specializes on its first argument.

  I'm not sure what this means for a generic function.

* The default method can also signal an error but the documentation string sounds as if that is not a possibility. One should change.

Backends/CLX/frame-manager.lisp
```diff
@@ -58,64 +58,27 @@
...
-      (:climi ,type #:-pane)
-      (:climi ,type))))
-
-(defun generate-clx-pane-specs (type)
-  (append
-   `((:clim-clx #:clx- ,type #:-pane)
-     (:clim-clx #:clx- ,type)
-     (:climi #:clx- ,type #:-pane)
-     (:climi #:clx- ,type))
-   (generate-standard-pane-specs type)))
```
scymtym: If I read the new version correctly, this naming scheme is no longer supported. Are there no pane classes which use this?

Backends/CLX/frame-manager.lisp
```diff
@@ -126,17 +89,13 @@
...
 (defmethod make-pane-1 ((fm clx-frame-manager) (frame application-frame) type &rest args)
-  (apply #'make-instance
-	 (maybe-mirroring fm (find-concrete-pane-class type))
-	 :frame frame
-	 :manager fm
-	 :port (port frame)
-	 args))
+  (apply #'make-instance (find-concrete-pane-class fm type)
+	 :frame frame :manager fm :port (port frame)
+         args))
```
scymtym: Multiple backends have this exact method. Could this become a default method?

Core/clim-core/panes.lisp
```diff
@@ -363,11 +363,38 @@ order to produce a double-click")
   (print-unreadable-object (pane sink :type t :identity t)
     (prin1 (pane-name pane) sink)))
 
+(defgeneric find-concrete-pane-class (pane-realizer pane-type)
+  (:documentation "Resolves abstract pane name to a concrete pane ~
+class which may be instatinated. Function specializes on its first ~
```
scymtym: In #929 you suggested

> Generate clause genearate-ensure-pane-form in your code should invoke make-pane-1 with the appropriate (see make-pane) realizer without trying to coerce the type on its own.

This would work best if either 
* `find-concrete-pane-class` returned a class *name*
* or `make-pane` accepted a class as the `type` argument

scymtym: A method `find-concrete-pane-class t class` that just returns the class does the job.

Core/clim-core/panes.lisp
```diff
@@ -363,11 +363,38 @@ order to produce a double-click")
   (print-unreadable-object (pane sink :type t :identity t)
     (prin1 (pane-name pane) sink)))
 
+(defgeneric find-concrete-pane-class (pane-realizer pane-type)
```
scymtym: Another observation: it will probably be a common case that a client cannot continue when no class is found. For that case (and probably as the default behavior), it would be good to signal an error when no suitable class can be found (similar to `cl:find-class`).

I suggest either `&optional errorp` or `&key if-does-not-exist` to control the behavior.

Backends/CLX/frame-manager.lisp
dkochmanski: classes which names are starting with #:clx- are produced with maybe-mirroring method. Since now we first resolve the clim class, then we optionally add #:clx-, and we *always* have this symbol in clim-clx package, this scheme is indeed unnecessary. It was already overboard with clim::clx- classes (scenario which never happened).

Backends/CLX/frame-manager.lisp
dkochmanski: good idea.

Core/clim-core/panes.lisp
dkochmanski: this method is responsible for finding a concrete class, not a symbol of said class. if you need the class name, you may use `class-name` on the result. abstract pane names are symbols, not classes.

I've suggested, that you should not rely on ad-hoc function which slaps -pane at the end of the symbol but rather rely on the realizer resolution as specified. Only after that you have clarified, that you need the exact class before going down the make-pane road because you have a reinitialization optimization in mind.

Core/clim-core/panes.lisp
dkochmanski: makes sense, I'll add &optional errorp

Core/clim-core/panes.lisp
dkochmanski: in that case: are SYMBOL and NAME separate arguments or a single argument? is it possible to distinguish with this notation?

Core/clim-core/panes.lisp
dkochmanski: fixed

Core/clim-core/panes.lisp
dkochmanski: fixed (in slightly different style, but preserving the spirit of this remark)

Core/clim-core/panes.lisp
dkochmanski: fixed

Core/clim-core/panes.lisp
dkochmanski: I've changed it to

````
    ;;    - look for a class `climi::{SYMBOL-NAME}-pane'
    ;;    - look for a class `climi::{SYMBOL-NAME}'
```

please resolve if you agree.

Core/clim-core/panes.lisp
dkochmanski: fixed

Backends/CLX/frame-manager.lisp
dkochmanski: fixed

dkochmanski:
I believe I've replied to all remarks. Commits are pushed.

Core/clim-core/panes.lisp
dkochmanski: after a bit of thought I've came to conclusion, that allowing passing verbatim class as a pane type won't hurt. fixed.

Core/clim-core/panes.lisp
scymtym: That's fine given the ambiguity you pointed out.

Core/clim-core/panes.lisp
```diff
@@ -363,11 +363,46 @@ order to produce a double-click")
   (print-unreadable-object (pane sink :type t :identity t)
     (prin1 (pane-name pane) sink)))
 
+(defgeneric find-concrete-pane-class (pane-realizer pane-type &optional errorp)
+  (:documentation "Resolves abstract pane name to a concrete pane
+class. Methods should specialize on the first argument.")
```
scymtym: > Resolves abstract pane name

I still think this should say "the abstract pane type PANE-TYPE …" or something similar. Specifically not "pane name".

> Methods should specialize on the first argument.

Why not the second? Maybe a frame manager has a special translation for one abstract pane type. Why shouldn't that be solved with an `eql`-specialized method?

Please document the `errorp` parameter.

Core/clim-core/panes.lisp
```diff
@@ -363,11 +363,46 @@ order to produce a double-click")
...
+    ;;
+    ;; 0. When the pane-type is a class return it as is.
+    ;; 1. Abstract mapping always takes a priority. When it exists we
+    ;;    don't look further.
+    ;; 2. When the symbol is in clim/climi/keyword package:
+    ;;    - look for a class `climi::{SYMBOL-NAME}-pane'
+    ;;    - look for a class `climi::{SYMBOL-NAME}'
+    ;; 3. Otherwise find a class named by the symbol.
+    (when (typep pane-type 'class)
+      (return-from find-concrete-pane-class pane-type))
```
scymtym: This should be a separate method in my opinion.

Core/clim-core/panes.lisp
dkochmanski: I've updated a documentation. As of method specializations: because it is easier to navigate the code. Spaghetti-like generic functions are easy to write, not so easy to understand. This is a simplification measure taken by a clostrophobic person.

Core/clim-core/panes.lisp
dkochmanski: ditto

Core/clim-core/panes.lisp
scymtym: > As of method specializations: because it is easier to navigate the code. Spaghetti-like generic functions are easy to write, not so easy to understand. This is a simplification measure taken by a clostrophobic person.

That doesn't make sense to me. The logical conclusion would be avoiding generic functions alltogether and always using `(defun (…) (typecase …))`.

dkochmanski: I can see that. In case of this function indeed that should be an ordinary function; latitude for having specialized panes is already granted by `make-pane-1`, the only reason we complicate make-pane protocol even further is because of mechanism you seem to need. I will look again into the other pull request and see, if it can't be achieved another way.

Regarding this particular function, it is the frame manager who find a pane, so this parameter should be specialized to allow extensions by backends. Having too many moving parts for sake of definition freedom is just unnecessary baggage.

Backends/CLX/frame-manager.lisp
```diff
@@ -58,64 +58,28 @@
...
-;;; the generic pane types based upon the type of the frame-manager. However, in
-;;; the CLX case, we don't expect there to be any CLX specific panes. CLX uses
-;;; the default generic panes instead.
+(defmethod find-concrete-pane-class ((fm clx-frame-manager)
+                                     pane-type &optional (errorp t))
+  ;; This backend doesn't have any specialized pane implementations
+  ;; but depending on circumstances it may add optional mirroring to
+  ;; the class. Such automatically defined concrete class has the same
+  ;; name but with a gensym prefix and symbol in the backend package.
+  (maybe-mirroring fm (call-next-method)))
```
scymtym: I believe this should be &optional errorp … (declare (ignore errorp)).

Backends/CLX-fb/frame-manager.lisp
```diff
@@ -37,10 +37,10 @@
...
-	 :manager fm
-	 :port (port frame)
-	 args))
+(defmethod find-concrete-pane-class ((fm clx-fb-frame-manager)
+                                     pane-type &optional (errorp t))
+  ;; This backend doesn't have any specialized pane implementations
+  ;; but depending on circumstances it may add optional mirroring to
+  ;; the class. Such automatically defined concrete class has the same
+  ;; name but with a gensym prefix and symbol in the backend package.
+  (maybe-mirroring fm (find-concrete-pane-class t pane-type errorp)))
```
scymtym: Couldn't this `(call-next-method)` like the method for `clx-frame-manager`?

dkochmanski: no, because the next method is clx-frame-manager, which would be essentially:

`(clim-fb-clx::maybe-mirroring (clim-clx::maybe-mirroring (find-concrete-pane-class t pane-type errorp)))`

Backends/CLX/frame-manager.lisp
dkochmanski: right you are. fixed.

Backends/CLX-fb/frame-manager.lisp
scymtym: Right.

Strangely though, `clim-fb-clx::maybe-mirroring` checks whether `concrete-pane-class` already is a subclass of `mirrored-sheet-mixin` as if it was intended to be idempotent. That might make `(clim-fb-clx::maybe-mirroring (clim-fb-clx::maybe-mirroring …))` work, but not `(clim-fb-clx::maybe-mirroring (clim-clx::maybe-mirroring …))`, so you are right.

PR #933, opened 2020-01-11
[hack][suggestion] redefinition on update suggestion

dkochmanski:
Some notes from looking into `wip-redefinition-update` :

* Can't redefine the application frame if `update-instances-on-redefinition` option has changed

```
Cannot CHANGE-CLASS objects into CLASS metaobjects.
   [Condition of type SB-PCL::METAOBJECT-INITIALIZATION-VIOLATION]
See also:
  The Art of the Metaobject Protocol, CLASS [:initialization]
```

* `with-delayed-redefinition-notifications` macro is unnecessary

Protocol has not enough justification. It is used in a single place
where a function call at the end of define-application-frame would
suffice:

```
  ,@(when update-instances-on-redefinition 
      `(;(handle-frame-class-redefinition (find-class ',name))
        (note-frame-class-redefinition (find-class ',name))))
```

Another reason why `with-drn` is not very useful outside of
define-application-frame is that redefining the application frame
superclass (i.e a mixin) does not trigger update in the subclass
instances. I don't see any compelling scenario where this is useful.

* `reinitialize-instance :after` method is unnecessary

Given we put note-frame-class-redefinition at the end of
define-application-frame we do not need delayed
reinitialize-instance. Since the mechanism is named
on-redefinition then I wouldn't expect reinitialization to trigger
the update (that would conflate two different mechanisms).

### DIFF
--- a/Core/clim-core/redefinition.lisp
+++ b/Core/clim-core/redefinition.lisp
@@ -22,42 +22,13 @@
 ;;; This is used for notifying application frame classes after aspects
 ;;; such as the class itself, the `generate-panes' method or the
 ;;; command table have been redefined. The notification sequence
-;;; begins with a call to `note-frame-class-redefined' which either
-;;; directly calls `handle-frame-class-redefinition' or schedules the
-;;; call to be made when leaving a
-;;; `with-delayed-redefinition-notifications'.
+;;; begins with a call to `note-frame-class-redefined' which calls
+;;; `handle-frame-class-redefinition'.
 ;;;
 ;;; This protocol is probably most commonly initiated by re-evaluating
 ;;; the `define-application-frame' form that defined the frame class
-;;; in question, but any event that causes the `reinitialize-instance'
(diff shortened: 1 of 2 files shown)


Core/clim-core/redefinition.lisp
```diff
@@ -98,80 +69,39 @@
...
-               ;; Update instances that haven't become garbage and
-               ;; haven't been CHANGE-CLASSed. Implicitly drop other
-               ;; instances.
+(defun note-frame-class-redefined (class)
+  (handle-frame-class-redefinition class))
+
+(defun handle-frame-class-redefinition
+    (class &aux (class-name (class-name class)))
+  (if-let ((instances (member-if #'tg:weak-pointer-value
+                                 (get class-name 'instances))))
```
dkochmanski: symbol plist could be replaced by a weak hash table with keys being classes.

dkochmanski:
closing due to no consensus of how this should be implemented

Issue #937, opened 2020-01-14
drawing operations performed twice during startup


dkochmanski:
This is a result of lack of coordination between the repaint protocol and the stream output recording protocol. Typical scenario for output recording stream:

applicaiton starts
window is opened
stream drawing operations happen | window repaint event queued (port thread)
stream-replay
window repaint event triggers handle-repaint and in consequence stream-replay

Default method handle-repaint on seos (standard-extended-output-stream) does nothing unlike the sors (standard-output-recording-stream), that's why this is more pressing for recording streams.

dkochmanski:
Issue is impossible to solve completely because of asynchronous nature of events. window-repaint-event may come at any time. solution I've taken: displaying the function first records the output, then dispatches a repaint. We have now three scenarios (we are assuming standard-repaint-mixin in this analysis):

1. window-repaint-event is in the queue, then the new event is compressed with it. during next event read window is repainted
2. there is no window-repaint-event in the queue, then our dispatched event is *the* event
2a. a new window-repaint-event comes before event read: it is compressed
2b. event is read, then window-repaint-event comes, window is repainted twice

in case of 2b it is impossible to say if the region was damaged after the last repaint or before it, hence we need to repaint to be sure.

in other words the change (in the upcoming pr) makes it less likely that sheet is repainted twice, however it may still happen.

Detecting if we are during the repaint and refuse to add window-repaint-event if it repaints a subregion of the current repaint doesn't make sense either, because:
1. we replay region a
2. we replay region b
3. region a is damaged
4. window-repaint-event for region a is sent
5. we replay region c

if we ignore the event, a is damaged.

Issue #927, opened 2019-12-30
proposal: display function performs no actual output


dkochmanski:
Displaying output with big coordinates may cause problems. Consider the following example:

```
; > (ql:quickload 'slim)
(slim:defapp (vis-a-vis :scroll-bars t) (x y)
  (clim:draw-oval* *standard-output* x y 100 100))

(vis-a-vis 100000 100000)
```
clx errors with:
```
The value
  99800
is not of type
  (SIGNED-BYTE 16)
when binding XLIB::X
   [Condition of type TYPE-ERROR]
```

because X11 coordinates are limited to 16 bits. This problem does not occur, if we first record the body and then replay it, because recording clips the region to the visible area. The proposal is to invoke display function always with recording `T` and drawing `NIL`, and then replaying the output record history. Incremental display protocols talk about "computing output records" what may be a clue, that this is the exact responsibility of display function.

Advantages:
- there is no actual drawing operations of output records which won't be visible on a sheet (possible performance gain)
- we are now able to add software transparency
- possibly we could not draw output records which are obscured by other output records

Potential problems:
- pixmap copying operations do not interact with output records, so using them in a display function will have no effect

Change seems to be as easy as fixing a function `invoke-display-function` by wrapping it in `with-output-recording-options` and calling `stream-replay` afterwards (alternatively in two places which call this function). Then the abovementioned snippet works just fine.

Considerations:
- problem may be easily mitigated by wrapping display function body by the programmer in with-recording-options
- likewise, if we incorporate this change, programmer still may wrap the body in with-output-recording-options with draw being t, however stream-replay called after that may obscure the output either way

Related to #793.  Opinions welcome.

Issue #793, opened 2019-07-05
Displaying large amounts of text gives multiple warnings about rendering string outside the mirror


john-a-carroll:
Attempting to display large amounts of text results in multiple messages

`WARNING: Trying to render string outside the mirror`

To reproduce, run the example code in #630 "glitches during long display". If the loop upper limit in the function `demo` is changed to 1791 or less then there are no problems.

This is with mcclim-20190521-git, clx-20190521-git, SBCL 1.5.0, macOS 10.12.6, XQuartz 2.7.11. The warning is signalled somewhere inside `mcclim-font:draw-glyphs`.

john-a-carroll:
I should have added: this is a regression, since there is no problem with mcclim-20181210-git, clx-20181210-git

dkochmanski:
I've failed to find the actual bug report, but this warning was introduced to make the issue more apparent. Here's the gist of it:

when we draw-text* (i.e do not replay output records), then coordinates may be arbitrary and in the worst case they go beyond the maximum mirror size (32k on X11). Since it is an initial display swizzling is not active yet, so some X11 servers signal an error. Now we catch the overboard position and instead of calling the draw function we just signal a warning.

The issue is valid, but warning is only a recently added canary, not the problem itself.

dkochmanski:
Quick check in git-blame shows commit 1363dbd98d9006377e4bdc399e53f1b1c416bb86 and issue #750. So this wasn't directly addressing wrong coordinates which I have described above, though my memory tells me that what I've described is correct.

john-a-carroll:
Aha, thanks for the explanation. Wrapping the output loop in

`(clim:with-output-recording-options (pane :record t :draw nil) ...)`

seems like a solution. I guess this delays actual output until coordinate swizzling has been set up. Is it the best approach?

dkochmanski:
yes, that's the best way to do it right now. I've created a proposal ticket which addresses the core of the issue.

PR #938, opened 2020-01-15
Fix various

dkochmanski:
atomic fixes for a few issues reported lately. also a cosmetic improvement to slim library.

### DIFF
--- a/Backends/CLX/fonts.lisp
+++ b/Backends/CLX/fonts.lisp
@@ -41,7 +41,7 @@
     (when fonts
       (xlib:open-font display (first fonts)))))
 
-(defmethod text-style-mapping ((port clx-port) text-style &optional character-set
+(defmethod text-style-mapping ((port clx-port) (text-style text-style) &optional character-set
                                &aux (text-style (climb:parse-text-style* text-style)))
   (declare (ignore character-set))
   (labels
--- a/Core/clim-basic/drawing/medium.lisp
+++ b/Core/clim-basic/drawing/medium.lisp
@@ -300,8 +300,9 @@
 (defun parse-text-style (style)
   (cond ((text-style-p style) style)
         ((null style) (make-text-style nil nil nil)) ; ?
-        ((and (listp style) (<= 3 (length style) 4))
-         (apply #'make-text-style style))
+        ((and (listp style) (alexandria:length= 3 style))
+         (destructuring-bind (family face size) style
+           (make-text-style family face size)))
         (t (error "Invalid text style specification ~S." style))))
 
 (defmacro with-text-style ((medium text-style) &body body)
--- a/Core/clim-basic/drawing/medium.lisp
+++ b/Core/clim-basic/drawing/medium.lisp
@@ -230,6 +230,11 @@
   ;; text-style like other methods. -- jd 2020-01-15
   text-style)
 
+(defmethod text-style-mapping ((port basic-port)
+                               text-style
+                               &optional character-set)
+  (text-style-mapping port (parse-text-style text-style) character-set))
+
 (defmethod text-style-mapping :around ((port basic-port)
                                        (text-style text-style)
                                        &optional character-set)
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -282,11 +282,10 @@ documentation produced by presentations.")
              (eq (sheet-parent (frame-panes frame))
                  (frame-top-level-sheet frame)))
     (sheet-disown-child (frame-top-level-sheet frame) (frame-panes frame)))
-  (loop
-     for (nil . pane) in (frame-panes-for-layout frame)
-     for parent = (sheet-parent pane)
-     if  parent
-     do (sheet-disown-child parent pane)))
+  (loop for (nil . pane) in (frame-panes-for-layout frame)
+        for parent = (sheet-parent pane)
+        if  parent
+          do (sheet-disown-child parent pane)))
 
 (defmethod generate-panes :after (fm (frame application-frame))
   (declare (ignore fm))
@@ -645,11 +644,7 @@ documentation produced by presentations.")
   (let ((*application-frame* frame)
         (event-queue (frame-event-queue frame)))
     (setf (slot-value frame 'top-level-sheet)
-          (make-pane-1 fm frame 'top-level-sheet-pane
-                       :name (frame-name frame)
-                       :pretty-name (frame-pretty-name frame)
-                       ;; sheet is enabled from enable-frame
-                       :enabled-p nil))
+          (find-pane-for-frame fm frame))
     (generate-panes fm frame)
     (setf (slot-value frame 'state) :disabled)
     (when (typep event-queue 'event-queue)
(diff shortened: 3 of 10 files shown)


Core/clim-core/frames.lisp
```diff
@@ -773,45 +768,44 @@ documentation produced by presentations.")
...
-                                           (t nil))
-                                       ,@(rest layout)
-                                       ,@(when pointer-documentation
-                                           '(%pointer-documentation%)))))
-                                 layouts)
-                         layouts)))))
-       ;; Update frame-current-panes and the special pane slots.
-       (update-frame-pane-lists frame))))
+     (with-look-and-feel-realization (fm frame)
+       (unless (frame-panes-for-layout frame)
```
scymtym: I thought about this change as well but should `with-look-and-feel-realization` actually bind `*application-frame*`?

Core/clim-core/frames.lisp
```diff
@@ -879,6 +873,24 @@ documentation produced by presentations.")
...
+
+(defmethod generate-panes (fm (frame application-frame))
+  (with-look-and-feel-realization (fm frame)
+    (unless (frame-panes-for-layout frame)
+      (setf (frame-panes-for-layout frame)
+            `((single-pane . ,(make-clim-interactor-pane :name 'single-pane)))))
+    (let ((single-pane
+            (alexandria:assoc-value (frame-panes-for-layout frame)
+                                    'single-pane :test #'eq)))
+      (setf (frame-panes frame) single-pane)))
```
scymtym: Maybe add a comment explaining that `make-clim-interactor-pane` may wrap the `single-pane` instance in other panes and therefore the lookup in `frame-panes-for-layout` is needed.

Core/clim-core/frames.lisp
dkochmanski: It is not said in the spec, but I believe that this is the right thing to do. Otherwise frame argument would be unnecessary. Moreover from more tangible perspective, it is important that *application-frame* is bound to the current frame if we want to be able to use i.e the instance slots in pane definition. Consider the following:

```common-lisp
(define-application-frame poem-viewer ()
  ((title  :initarg :title  :reader poem-title)
   (verses :initarg :verses :reader poem-verses))
  (:panes (app :application
               :display-function (lambda (frame pane)
                                   (format pane (poem-verses frame)))))
  (:layouts
   (default
    (labelling (:label (poem-title *application-frame*)) ; <----
      app))))

(run-frame-top-level
 (make-application-frame 'poem-viewer
                         :title "CLIM"
                         :verses "Roses are red,~
                                  Violets are blue,~
                                  CLIM is software,~
                                  Which does kaboom."))
```

Core/clim-core/panes.lisp
```diff
@@ -2584,14 +2584,18 @@ SCROLLER-PANE appear on the ergonomic left hand side, or leave set to
...
+      (cond ((consp display-function)
+	     (apply (car display-function)
+		    frame pane (cdr display-function)))
+	    (display-function
+	     (funcall display-function frame pane))
+	    (t nil))))
+  ;; DISPATCH-REPAINT instead of STREAM-REPLAY ensures, that we do not
+  ;; redraw everything twice when window-repaint-event happens, thanks
+  ;; to the event compression in the event queue. -- jd 2020-01-15
+  (dispatch-repaint pane (sheet-region pane)))
```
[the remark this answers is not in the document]
dkochmanski: Previously we did draw and record at the same time. That lead to some problems[1] I've created a ticket with proposal[2] to address this kind of issues for recorded streams.

Basically at first we were drawing `+everywhere+` (so if a drawing operation did go beyond 16bit coordinate limit we had error from clx), and then repaint was limited to the viewport.

Now we repaint only records which are in the viewport (it is still possible that we will repaint it twice[3]).

[1] https://github.com/McCLIM/McCLIM/issues/793
[2] https://github.com/McCLIM/McCLIM/issues/927
[3] https://github.com/McCLIM/McCLIM/issues/937

Core/clim-core/frames.lisp
dkochmanski: the lookup in frame-panes-for-layout is because there because we create the interactor only once (in the `unless` form), `make-clim-interactor-pane` returns the wrapper pane as the first value (stream is returned as the second value).

dkochmanski:
I've rebased the PR and answered comments. There were no code changes in the last push.

Core/clim-core/frames.lisp
scymtym: It is clear that `define-application-frame` has to generate code that binds `*application-frame*`, my question was whether `with-look-and-feel-realization` should do it.

The argument regarding the parameter makes sense to me.

scymtym:
Not sure which commit caused this, but the "List Test" demo is broken. Using the scroll bar to scroll the list has no immediate effect while the pointer button is still pressed (there is some flickering in the scroll bar, but the viewport is not updated). When the button is released, the viewport updates.

scymtym:
I don't know whether this is some other regression, but "Text size test" does not draw the "Family", "Face" and "Rectangle" gadgets and also flickers continuously. It is possibly that some or all of these problems are due to lack of rebasing onto master.

Core/clim-core/frames.lisp
```diff
@@ -923,8 +935,9 @@ documentation produced by presentations.")
           ,@user-default-initargs)
          ,@other-options)
 
-       ,(generate-generate-panes-form
-         name menu-bar panes layouts pointer-documentation)
+       ,@(when panes
```
scymtym: This should probably be `when (or panes layouts)` otherwise frames like `demodemo` that use layouts but do not use the `:panes` option don't work.

dkochmanski:
missing gadgets indeed appeared after rebasing. flicker is because going through the queue is much slower. I'm withdrawing the display invocation change because there are other regressions related to Drei. I've cherry-picked fix for german towns demo regression.

as of display invocation and double initial repaint I'll address it sometime else if I'm able to overcome drei issue (it basically shortcuts repaint with redisplay).

Core/clim-core/frames.lisp
dkochmanski: fixed

dkochmanski:
as of list scrolling issue it was due to dispatch-repaint (we should have used stream-replay instead, then the problem disappears), that said there are still problems with drei when doing that.

Backends/CLX/fonts.lisp
```diff
@@ -41,7 +41,7 @@
     (when fonts
       (xlib:open-font display (first fonts)))))
 
-(defmethod text-style-mapping ((port clx-port) text-style &optional character-set
+(defmethod text-style-mapping ((port clx-port) (text-style text-style) &optional character-set
                                &aux (text-style (climb:parse-text-style* text-style)))
```
scymtym: Isn't `parse-text-style*` the identity when `text-style` already is a `text-style` instance?

dkochmanski: no, text-style may be incomplete (i.e it may have NIL components). It would be the identity for `parse-text-style` (without a star)

Core/clim-basic/drawing/medium.lisp
```diff
@@ -295,8 +300,9 @@
 (defun parse-text-style (style)
   (cond ((text-style-p style) style)
         ((null style) (make-text-style nil nil nil)) ; ?
-        ((and (listp style) (<= 3 (length style) 4))
-         (apply #'make-text-style style))
+        ((and (listp style) (alexandria:length= 3 style))
+         (destructuring-bind (family face size) style
+           (make-text-style family face size)))
```
scymtym: I think this could be a separate commit which should also add tests for `parse-text-style`.

Backends/CLX/fonts.lisp
scymtym: Right, I already forgot. But that reminds me: can we rename `parse-text-style*` to `ensure-fully-specified-text-style` or similar?

Core/clim-basic/drawing/medium.lisp
dkochmanski: fixed

Backends/CLX/fonts.lisp
dkochmanski: I'm not much into `overly-long-and-descriptive-name`'s, also it is indeed a parse-text-style variant, so "starred" name makes perfect sense here to me. note, that it is not only to ensure that the text style is fully specified but also to normalize the size

scymtym: Fair enough. I might bring it up again later, though :)

PR #940, opened 2020-01-15
Gadget cleanup and SCHEDULE-EVENT fix

scymtym:
Regarding the removed slider method, here is comparison between the removed method (right) and the current method (left):
mcclim-sliders-old-and-new


Core/clim-basic/windowing/sheets.lisp
```diff
@@ -123,9 +123,6 @@
 (defmethod sheet-children ((sheet basic-sheet))
   nil)
 
-(defmethod sheet-adopt-child ((sheet basic-sheet) (child sheet))
-  (error "~S attempting to adopt ~S" sheet child))
-
```
dkochmanski: why this method is removed?

dkochmanski:
all looks good to me except the removed code with sheet-adopt-child (which change I do not understand -- do you want to simply remove the primary method so it fails in more generic way?)

Core/clim-basic/windowing/sheets.lisp
scymtym: In general, I think `no-applicable-method` is the appropriate response for an unsupported operation on the given arguments in the context of a generic function.

In this particular case, the method makes working with mixins harder since either 
* users of the mixins (that is subclasses of the mixin classes) would have to tweak their list of direct superclasses such that the mixin precedes `basic-sheet`
* or the mixin would have to be realized in terms of `:around` methods that work as primary methods and must not call the next method which would be awkward.

Finally, there shouldn't be any code relying on this method since it signals an error (unless some code relies on the error being signaled, but in that case, a different design would likely be better).

dkochmanski: I see. Putting aside these practical concerns, from the specification perspective this change is fine, because the relationship protocol is defined for mixins (leaf, single, multiple and parent, see 7.2.2), not for basic-sheet.

PR #943, opened 2020-01-18
Fix dnd translators 2

dkochmanski:


dkochmanski:
closing to reduce confusion. this pr is referenced in the "master" pull request, whether this approach should be incorporated is a question of debate.

Issue #946, opened 2020-01-20
check-box and radio-box are not displayed


admich:
After the last PR #940  the check-box and the radio-box are not displayed anymore.

dkochmanski:
Issue is caused by a commit e7dd0930 `gadgets: simplify superclass lists of {SLIDER,{RADIO,CHECK}-BOX}-PANE`

dkochmanski:
quick fix is to reverse removal of basic-pane from the superclasses (as a direct superclass):

```
diff --git a/Core/clim-core/gadgets/concrete.lisp b/Core/clim-core/gadgets/concrete.lisp
index 19d6ad9a..3faf87c8 100644
--- a/Core/clim-core/gadgets/concrete.lisp
+++ b/Core/clim-core/gadgets/concrete.lisp
@@ -596,7 +596,8 @@
 (defclass slider-pane (slider
                        gadget-color-mixin
                        value-changed-repaint-mixin
-                       activate/deactivate-repaint-mixin)
+                       activate/deactivate-repaint-mixin
+                       basic-pane)
   ())
 
 (defmethod compose-space ((pane slider-pane) &key width height)
@@ -770,7 +771,8 @@
                           rack-layout-mixin
                           activate/deactivate-repaint-mixin
                           sheet-multiple-child-mixin
-                          activate/deactivate-children-mixin)
+                          activate/deactivate-children-mixin
+                          basic-pane)
   ()
   (:default-initargs :background *3d-normal-color*))
 
@@ -804,7 +806,7 @@
                           rack-layout-mixin
                           activate/deactivate-repaint-mixin
                           sheet-multiple-child-mixin
-                          activate/deactivate-children-mixin)
+                          basic-pane)
   ()
   (:default-initargs :background *3d-normal-color*))
```

I'm going to first find why this causes the regression before reverting the change.

dkochmanski:
OK, the gist of the problem is this:

1. method `sheet-children` has a specialization on `basic-pane` to return NIL (it is wrong, because it should be `sheet-leaf-mixin` specialization, but numerous panes does not have it as a superclass so removing it would be correct still will break many things, so better to have it in a bigger PR.

2. sheet-multiple-child-mixin has method sheet-children alright and sheet-adopt-child method defiend as follows:

(defmethod sheet-adopt-child ((sheet sheet-multiple-child-mixin)
			      (child sheet-parent-mixin))
  (push child (sheet-children sheet)))

[...]

PR #947, opened 2020-01-20
Fix 946

dkochmanski:


dkochmanski:
I'm merging this now because currently gadgets on master branch do not work as expected. Alternative solution would be to reverse previous PR completely but given what we've learned some gadgets would be still affected by the issue. This change does not conflict with a possibility to make a list of inherited classes irrelevant to the object behavior, so other remarks from #946 may be still proposed as the next PR.

PR #954, opened 2020-01-23
Fix bug in completion of logical pathnames

heshrobe:
This fixes a bug in the function filename-completer in /core/clim-core/presentation-defs.lisp when attempting to complete logical pathnames.  The issue is that when scanning the parent directory for completions, you may find pathnames that have characters that aren't legal in logical pathnames.  This introduces a local function, egitimate-logical-pathname that is used to filter these out when looking for possible completions.

### DIFF
--- a/Core/clim-core/presentation-defs.lisp
+++ b/Core/clim-core/presentation-defs.lisp
@@ -1623,12 +1623,12 @@ protocol retrieving gestures from a provided string."))
 		(new-directory (cons
 				(first his-directory)
 				(append home (rest (rest his-directory))))))
-	     (make-pathname :host (pathname-host pathname)
-			    :device (pathname-device pathname)
-			    :name (pathname-name pathname)
-			    :version (pathname-version pathname)
-			    :type (pathname-type pathname)
-			    :directory new-directory))))
+	   (make-pathname :host (pathname-host pathname)
+			  :device (pathname-device pathname)
+			  :name (pathname-name pathname)
+			  :version (pathname-version pathname)
+			  :type (pathname-type pathname)
+			  :directory new-directory))))
     ;; Slow but accurate
     (let* ((raw-pathname (pathname string))
 	   (raw-directory (pathname-directory raw-pathname))
@@ -1657,32 +1657,51 @@ protocol retrieving gestures from a provided string."))
       ;; Now prune out all completions that don't start with the string
       (let ((type (pathname-type actual-pathname)))
 	(when (null type)
-	  ;; If the user didn't supply a file type, don't burden him with all
-	  ;; sorts of version numbers right now.
-	  (let ((new-completions nil))
-	    (dolist (pathname completions)
-	      (cond
-		;; meaning this is actually a directory
-		((and (null (pathname-name pathname))
-		      (null (pathname-type pathname)))
-		 (pushnew (make-pathname :host (pathname-host original-pathname)
-					 :device (pathname-device original-pathname)
-					 :directory (butlast (pathname-directory pathname))
-					 :name (first (last (pathname-directory pathname)))
-					 :type nil)
-			  new-completions))
-		(t
-		 (pushnew (make-pathname :host (pathname-host original-pathname)
-					 :device (pathname-device original-pathname)
-					 :directory (pathname-directory original-pathname)
-					 :name (pathname-name pathname)
-					 :type (pathname-type pathname))
-			  new-completions))))
-	    (setq completions (nreverse new-completions))))
+	  (flet ((legitimate-logical-pathname (name)
+		   (let ((word (string name)))
+		     (dotimes (i (length word))
+		       (let ((ch (schar word i)))
+			 (unless (and (typep ch 'standard-char)
+				      (or (alpha-char-p ch) (digit-char-p ch) (char= ch #\-)))
+			   (return-from legitimate-logical-pathname nil))))
+		     t)))
+	    ;; If the user didn't supply a file type, don't burden him with all
+	    ;; sorts of version numbers right now.
+	    (let ((new-completions nil))
+	      (dolist (pathname completions)
+		(cond
+		  ;; meaning this is actually a directory
+		  ((and (null (pathname-name pathname))
+			(null (pathname-type pathname)))
+		   (when (and (loop for word in  (butlast (pathname-directory pathname))
+				 always (legitimate-logical-pathname word))
+			      (legitimate-logical-pathname (first (last (pathname-directory pathname)))))
+		     (pushnew (if (typep original-pathname 'logical-pathname)
+				  (make-pathname :host (pathname-host original-pathname)
+						 :device (pathname-device original-pathname)
+						 :directory (first (last (pathname-directory pathname)))
+						 :name nil
+						 :type nil)
+				  (make-pathname :host (pathname-host original-pathname)
+						 :device (pathname-device original-pathname)
+						 :directory (butlast (pathname-directory pathname))
+						 :name (first (last (pathname-directory pathname)))
+						 :type nil))
+			      new-completions)))
+		  (t
+		   (when (and (legitimate-logical-pathname (pathname-name pathname))
+			      (legitimate-logical-pathname (pathname-type pathname)))
+		     (pushnew (make-pathname :host (pathname-host original-pathname)
+					     :device (pathname-device original-pathname)
+					     :directory (pathname-directory original-pathname)
+					     :name (pathname-name pathname)
+					     :type (pathname-type pathname))
+			      new-completions)))))
+	      (setq completions (nreverse new-completions)))))
 	(complete-from-possibilities original-string completions '(#\space)
-						    :action action
-						    :name-key #'namestring
-						    :value-key #'identity)))))
+				     :action action
+				     :name-key #'namestring
+				     :value-key #'identity))))) 
 
 (define-presentation-method accept ((type pathname) stream (view textual-view)
                                     &key (default *default-pathname-defaults* defaultp)
--- a/Core/clim-core/presentation-defs.lisp
+++ b/Core/clim-core/presentation-defs.lisp
@@ -26,8 +26,8 @@
 (define-presentation-type form ()
   :options (auto-activate (preserve-whitespace t) (subform-read nil))
   :inherit-from `((expression) :auto-activate ,auto-activate
-                  :preserve-whitespace ,preserve-whitespace
-                  :subform-read ,subform-read ))
+		  :preserve-whitespace ,preserve-whitespace
+		  :subform-read ,subform-read ))
 
 ;;; Actual definitions of presentation methods and types.  They've
 ;;; been separated from the macro and presentation class definitions and
@@ -46,8 +46,8 @@
     type
     (when (null parameters)
       (let ((clos-class (find-class name nil))) ; Don't error out.
-        (when (and clos-class (typep clos-class 'standard-class))
-          (return-from presentation-typep (typep object name)))))
+	(when (and clos-class (typep clos-class 'standard-class))
+	  (return-from presentation-typep (typep object name)))))
     (funcall-presentation-generic-function presentation-typep object type)))
 
 ;;; Not defined as a generic function, but what the hell.
@@ -60,37 +60,37 @@
 
 (defun get-ptype-from-class-of (object)
   (let* ((name (class-name (class-of object)))
-         (ptype-entry (gethash name *presentation-type-table*)))
+	 (ptype-entry (gethash name *presentation-type-table*)))
     (unless ptype-entry
       (return-from get-ptype-from-class-of nil))
     ;; Does the type have required parameters?  If so, we can't use it...
     (let ((parameter-ll (parameters-lambda-list ptype-entry)))
       (values name
-              (if (eq (car parameter-ll) '&whole)
-                  (cddr parameter-ll)
-                  parameter-ll)))))
+	      (if (eq (car parameter-ll) '&whole)
+		  (cddr parameter-ll)
+		  parameter-ll)))))
 
 (defmethod presentation-type-of ((object standard-object))
   (multiple-value-bind (name lambda-list)
       (get-ptype-from-class-of object)
     (cond ((and name
-                (or (null lambda-list)
-                    (member (first lambda-list) lambda-list-keywords)))
-           name)
-          (name
-           'standard-object)
-          (t (let* ((class (class-of object))
-                    (class-name (class-name class)))
-               (or class-name class))))))
+		(or (null lambda-list)
+		    (member (first lambda-list) lambda-list-keywords)))
+	   name)
+	  (name
+	   'standard-object)
+	  (t (let* ((class (class-of object))
+		    (class-name (class-name class)))
+	       (or class-name class))))))
 
 (defmethod presentation-type-of ((object structure-object))
   (multiple-value-bind (name lambda-list)
       (get-ptype-from-class-of object)
     (if (and name
-             (or (null lambda-list)
-                 (member lambda-list lambda-list-keywords)))
-        name
-        (call-next-method))))
+	     (or (null lambda-list)
+		 (member lambda-list lambda-list-keywords)))
+	name
+	(call-next-method))))
 
 (define-presentation-generic-function
     %map-over-presentation-type-supertypes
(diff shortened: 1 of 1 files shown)


Core/clim-core/presentation-defs.lisp
```diff
@@ -1623,12 +1623,12 @@ protocol retrieving gestures from a provided string."))
...
-			    :name (pathname-name pathname)
-			    :version (pathname-version pathname)
-			    :type (pathname-type pathname)
-			    :directory new-directory))))
+	   (make-pathname :host (pathname-host pathname)
+			  :device (pathname-device pathname)
+			  :name (pathname-name pathname)
+			  :version (pathname-version pathname)
+			  :type (pathname-type pathname)
+			  :directory new-directory))))
```
dkochmanski: please detabify your changes. In emacs that would be:

(setq indent-tabs-mode nil)
M-x whitespace-cleanup

Core/clim-core/presentation-defs.lisp
```diff
@@ -1657,32 +1657,51 @@ protocol retrieving gestures from a provided string."))
...
-					 :type (pathname-type pathname))
-			  new-completions))))
-	    (setq completions (nreverse new-completions))))
+	  (flet ((legitimate-logical-pathname (name)
+		   (let ((word (string name)))
+		     (dotimes (i (length word))
+		       (let ((ch (schar word i)))
+			 (unless (and (typep ch 'standard-char)
+				      (or (alpha-char-p ch) (digit-char-p ch) (char= ch #\-)))
+			   (return-from legitimate-logical-pathname nil))))
```
dkochmanski: please use loop and more specific predicates, i.e

```
(loop for ch across (string name)
      unless (and (standard-char-p ch)
                  (alphanumericp ch))
        return nil
      finally (return t))
```

scymtym:
```lisp
(loop …
      unless (and …)
        return nil
     finally (return t))
```
can be simplified to `(loop ... always (and …))`.

heshrobe: I've accepted you suggestions and will resubmit soon.
Daniel, with regard to tabs, the easier emacs fix is to select the function and then use mx detabify

In terms of an example to illustrate the problem:
1) Make a directory, say "foo"
2) create a file in it with a character not allowed in a logical pathname, e.g. foo_bar
3) create a logical pathname that resolves to that directory

The following code will do that:
(ensure-directories-exist "~/foobar/bar/")
(let* ((final-pathname (pathname "~/foobar/**/*.*")))
  (setf (logical-pathname-translations "foo")
	`(("**;*.*"	,final-pathname))))
(with-open-file (f "~/foobar/bar/baz_bar.lisp" :direction :output :if-does-not-exist :create)
  )
(with-open-file (f "~/foobar/bar/baz-bar.lisp" :direction :output :if-does-not-exist :create)
  )

With the patch doing (clim:accept 'clim:pathname) and responding foo:ba followed by tab (or whatever the complete gesture is) will work, but won't without it.

dkochmanski:
changes still have plenty of newly introduced tabs.

Core/clim-core/presentation-defs.lisp
```diff
@@ -1610,101 +1610,118 @@ protocol retrieving gestures from a provided string."))
...
+                                                     :type nil)
+                                      (make-pathname :host (pathname-host original-pathname)
+                                                     :device (pathname-device original-pathname)
+                                                     :directory (butlast (pathname-directory pathname))
+                                                     :name (first (last (pathname-directory pathname)))
+                                                     :type nil))
+                                  new-completions)))
+                      (t
+                       (when (and (legitimate-logical-pathname (pathname-name pathname))
+                                  (legitimate-logical-pathname (pathname-type pathname)))
```
dkochmanski: we do not push the pathname at all if it is not a logical pathname?

Core/clim-core/presentation-defs.lisp
```diff
@@ -1610,101 +1610,118 @@ protocol retrieving gestures from a provided string."))
...
+            ;; sorts of version numbers right now.
+            (let ((new-completions nil))
+              (loop for pathname in completions
+                 do (cond
+                      ;; meaning this is actually a directory
+                      ((and (null (pathname-name pathname))
+                            (null (pathname-type pathname)))
+                       (when (and (loop for word in  (butlast (pathname-directory pathname))
+                                     always (legitimate-logical-pathname word))
+                                  (legitimate-logical-pathname (first (last (pathname-directory pathname)))))
```
dkochmanski: why do you test for the last element of a directory separately?

Core/clim-core/presentation-defs.lisp
```diff
@@ -1610,101 +1610,118 @@ protocol retrieving gestures from a provided string."))
...
+                         (pushnew (if (typep original-pathname 'logical-pathname)
```
dkochmanski: this has been already computed in lat above (result is in a variable logical-pathname-p), no need to call typep again

Core/clim-core/presentation-defs.lisp
```diff
@@ -1610,101 +1610,118 @@ protocol retrieving gestures from a provided string."))
...
+                                      (make-pathname :host (pathname-host original-pathname)
+                                                     :device (pathname-device original-pathname)
+                                                     :directory (first (last (pathname-directory pathname)))
+                                                     :name nil
+                                                     :type nil)
+                                      (make-pathname :host (pathname-host original-pathname)
```
dkochmanski: instead of duplicating make-pathname form introduce local variables name and directory and setf them, then do a single pushnew

Core/clim-core/presentation-defs.lisp
```diff
@@ -1610,101 +1610,118 @@ protocol retrieving gestures from a provided string."))
...
+                                    (or (alphanumericp ch) (char= ch #\-)))))))
+            ;; If the user didn't supply a file type, don't burden him with all
+            ;; sorts of version numbers right now.
+            (let ((new-completions nil))
+              (loop for pathname in completions
+                 do (cond
+                      ;; meaning this is actually a directory
+                      ((and (null (pathname-name pathname))
+                            (null (pathname-type pathname)))
+                       (when (and (loop for word in  (butlast (pathname-directory pathname))
```
dkochmanski: do we not complete at all when the pathname is not a logical one?

Core/clim-core/presentation-defs.lisp
```diff
@@ -1610,101 +1610,118 @@ protocol retrieving gestures from a provided string."))
...
-					 :name (pathname-name pathname)
-					 :type (pathname-type pathname))
-			  new-completions))))
-	    (setq completions (nreverse new-completions))))
-	(complete-from-possibilities original-string completions '(#\space)
-						    :action action
-						    :name-key #'namestring
-						    :value-key #'identity)))))
+        (when (null type)
+          (flet ((legitimate-logical-pathname (name)
```
dkochmanski: please lift this function to the flet which wraps whole function body

heshrobe:
With regard to point 3, I guess I don’t understand how to build a regression test for something that is as GUI oriented as pathname completion.  Are there examples in the current regression suite that I can use as a model?


> On Jan 31, 2020, at 04:24, Daniel Kochmanski <notifications@github.com> wrote:
> 
> @dkochmanski requested changes on this pull request.
> 
> I did a second pass over this PR. Some points are unclear to me. Also:
> 
> Please introduce this change in three commits:
> 
> 	• whitespace and indentation cleanup of the whole file (no tabs)
> 	• fix to logical pathnames
> 	• add a regression test to Tests/presentations.lisp
> In Core/clim-core/presentation-defs.lisp:
> 
> > +                                  (legitimate-logical-pathname (first (last (pathname-directory pathname)))))
> +                         (pushnew (if (typep original-pathname 'logical-pathname)
> +                                      (make-pathname :host (pathname-host original-pathname)
> +                                                     :device (pathname-device original-pathname)
> +                                                     :directory (first (last (pathname-directory pathname)))
> +                                                     :name nil
> +                                                     :type nil)
> +                                      (make-pathname :host (pathname-host original-pathname)
> +                                                     :device (pathname-device original-pathname)
> +                                                     :directory (butlast (pathname-directory pathname))
> +                                                     :name (first (last (pathname-directory pathname)))
> +                                                     :type nil))
> +                                  new-completions)))
> +                      (t
> +                       (when (and (legitimate-logical-pathname (pathname-name pathname))
> +                                  (legitimate-logical-pathname (pathname-type pathname)))
> 
> we do not push the pathname at all if it is not a logical pathname?
> 
> In Core/clim-core/presentation-defs.lisp:
> 
> > +                   (let ((word (string name)))
> +                     (loop for i below (length word)
> +                        for ch = (schar word i)
> +                        always (and (standard-char-p ch)
> +                                    (or (alphanumericp ch) (char= ch #\-)))))))
> +            ;; If the user didn't supply a file type, don't burden him with all
> +            ;; sorts of version numbers right now.
> +            (let ((new-completions nil))
> +              (loop for pathname in completions
> +                 do (cond
> +                      ;; meaning this is actually a directory
> +                      ((and (null (pathname-name pathname))
> +                            (null (pathname-type pathname)))
> +                       (when (and (loop for word in  (butlast (pathname-directory pathname))
> +                                     always (legitimate-logical-pathname word))
> +                                  (legitimate-logical-pathname (first (last (pathname-directory pathname)))))
> 
> why do you test for the last element of a directory separately?
> 
> In Core/clim-core/presentation-defs.lisp:
> 
> > +                     (loop for i below (length word)
> +                        for ch = (schar word i)
> +                        always (and (standard-char-p ch)
> +                                    (or (alphanumericp ch) (char= ch #\-)))))))
> +            ;; If the user didn't supply a file type, don't burden him with all
> +            ;; sorts of version numbers right now.
> +            (let ((new-completions nil))
> +              (loop for pathname in completions
> +                 do (cond
> +                      ;; meaning this is actually a directory
> +                      ((and (null (pathname-name pathname))
> +                            (null (pathname-type pathname)))
> +                       (when (and (loop for word in  (butlast (pathname-directory pathname))
> +                                     always (legitimate-logical-pathname word))
> +                                  (legitimate-logical-pathname (first (last (pathname-directory pathname)))))
> +                         (pushnew (if (typep original-pathname 'logical-pathname)
> 
> this has been already computed in lat above (result is in a variable logical-pathname-p), no need to call typep again
> 
> In Core/clim-core/presentation-defs.lisp:
> 
> > +            (let ((new-completions nil))
> +              (loop for pathname in completions
> +                 do (cond
> +                      ;; meaning this is actually a directory
> +                      ((and (null (pathname-name pathname))
> +                            (null (pathname-type pathname)))
> +                       (when (and (loop for word in  (butlast (pathname-directory pathname))
> +                                     always (legitimate-logical-pathname word))
> +                                  (legitimate-logical-pathname (first (last (pathname-directory pathname)))))
> +                         (pushnew (if (typep original-pathname 'logical-pathname)
> +                                      (make-pathname :host (pathname-host original-pathname)
> +                                                     :device (pathname-device original-pathname)
> +                                                     :directory (first (last (pathname-directory pathname)))
> +                                                     :name nil
> +                                                     :type nil)
> +                                      (make-pathname :host (pathname-host original-pathname)
> 
> instead of duplicating make-pathname form introduce local variables name and directory and setf them, then do a single pushnew
> 
> In Core/clim-core/presentation-defs.lisp:
> 
> > +        (when (null type)
> +          (flet ((legitimate-logical-pathname (name)
> +                   (let ((word (string name)))
> +                     (loop for i below (length word)
> +                        for ch = (schar word i)
> +                        always (and (standard-char-p ch)
> +                                    (or (alphanumericp ch) (char= ch #\-)))))))
> +            ;; If the user didn't supply a file type, don't burden him with all
> +            ;; sorts of version numbers right now.
> +            (let ((new-completions nil))
> +              (loop for pathname in completions
> +                 do (cond
> +                      ;; meaning this is actually a directory
> +                      ((and (null (pathname-name pathname))
> +                            (null (pathname-type pathname)))
> +                       (when (and (loop for word in  (butlast (pathname-directory pathname))
> 
> do we not complete at all when the pathname is not a logical one?
> 
> In Core/clim-core/presentation-defs.lisp:
> 
> > -					 :type nil)
> -			  new-completions))
> -		(t
> -		 (pushnew (make-pathname :host (pathname-host original-pathname)
> -					 :device (pathname-device original-pathname)
> -					 :directory (pathname-directory original-pathname)
> -					 :name (pathname-name pathname)
> -					 :type (pathname-type pathname))
> -			  new-completions))))
> -	    (setq completions (nreverse new-completions))))
> -	(complete-from-possibilities original-string completions '(#\space)
> -						    :action action
> -						    :name-key #'namestring
> -						    :value-key #'identity)))))
> +        (when (null type)
> +          (flet ((legitimate-logical-pathname (name)
> 
> please lift this function to the flet which wraps whole function body
> 
> —
> 

Howard Shrobe

Director CyberSecurity@CSAIL
hes@csail.mit.edu
617-253-7877 work
617-312-6439 cell

dkochmanski:
you are right of course, sorry. Please add an interactive regression test to clim-examples. i.e

```
(in-package :clim-demo)

(define-application-frame stream-completion-test ()
  ()
  (:menu-bar nil)
  (:pane :interactor))

(defmethod initialize-instance :after ((frame stream-completion-test) &key)
  (ensure-directories-exist "~/foobar/bar/")
  (let* ((final-pathname (pathname "~/foobar/**/*.*")))
    (setf (logical-pathname-translations "foo")
	  `(("**;*.*"	,final-pathname))))
  (with-open-file (f "~/foobar/bar/baz_bar.lisp" :direction :output :if-does-not-exist :create))
  (with-open-file (f "~/foobar/bar/baz-bar.lisp" :direction :output :if-does-not-exist :create)))

(define-stream-completion-test-command (com-test-logical-pathname-completion :name t)
    ()
  (format t "Try completions for `foo:ba'. <describe expected output>.~%")
  (format t "Accepted ~s." (accept 'pathname)))

(run-frame-top-level
 (make-application-frame 'stream-completion-test))

```

(leave out run-frame-top-level and add a test to "Tests" section in file Examples)

scymtym:
> With regard to point 3, I guess I don’t understand how to build a regression test for something that is as GUI oriented as pathname completion.  Are there examples in the current regression suite that I can use as a model?

I may be wrong, but it seems like the final `accept` call is the only interactive part. Maybe the function can be split into one function that computes the completion entries and a second function that calls `accept`? If this split is possible, a unit test for the first function should be possible.

heshrobe:
Jan is correct; I’ve restructured the function into the parts that build the list of candidates and then the part that does the actual call to clim:complete-from-possibiities.  To create a test case, I would have to make a directory and a few files within it.   I think the appropriate place would be a temporary directory; in Unix-like systems that would be /tmp.  But I’m not sure what the equivalent in would be in windows (I looked this up and it has changed in different versions and there is both a system-wide and a per user temp file).  The other approach would to just create a directory under *default-pathname-defaults* (or under the McClim heirarchy somewhere) and then to delete it after the test runs.

Any suggestions?

> On Feb 2, 2020, at 08:03, Jan Moringen <notifications@github.com> wrote:
> 
> With regard to point 3, I guess I don’t understand how to build a regression test for something that is as GUI oriented as pathname completion. Are there examples in the current regression suite that I can use as a model?
> 
> I may be wrong, but it seems like the final accept call is the only interactive part. Maybe the function can be split into one function that computes the completion entries and a second function that calls accept? If this split is possible, a unit test for the first function should be possible.
> 
> —
> 

Howard Shrobe

Director CyberSecurity@CSAIL
hes@csail.mit.edu
617-253-7877 work
617-312-6439 cell

scymtym:
Thank you for exploring ways to make this fix testable.

> To create a test case, I would have to make a directory and a few files within it.   I think the appropriate place would be a temporary directory; in Unix-like systems that would be /tmp.  But I’m not sure what the equivalent in would be in windows (I looked this up and it has changed in different versions and there is both a system-wide and a per user temp file).  

One option would be `uiop:with-temporary-file` but I can imagine @dkochmanski has an opinion on that.

> The other approach would to just create a directory under `*default-pathname-defaults*` (or under the McClim heirarchy somewhere) and then to delete it after the test runs.

I advice against both since one will litter the home directories of people running the tests while the other will litter the source code.

dkochmanski:
if writing a test is hard and has side effects which are not wanted, then I think we should just not write it after all.

/as of uiop, indeed, I don't think it would be a viable dependency for McCLIM, but I don't have that strong opinion about systems implementing tests/

PR #957, opened 2020-01-29
Fix command menu bug

heshrobe:
Context: an application frame with both a command menu pane and a pointer-documentation pane and where some command in the menu has arguments.  This happens specifically when one of the arguments is expected to be a data-structure (other than a symbol).

1) If the mouse just moves over this menu item can cause an error.  The reason is that the menu items are presented as commands; when the mouse moves over the menu-item the pointer-documentation is updated by calling present on this command.  This recursively calls present on each of the arguments.  One of these (sub) presentation methods might attempt to perform an operation on its argument; but in fact it is passed the symbol for undefined arguments, even though the presentation type is the one appropriate for the command.

2) Even if moving the mouse over the menu item doesn't cause an error, you get a message saying that you've provided an incomplete command, rather than calling the partial-command completer.

The fix involves:
1) Defining a new presentation-type: command-menu-item
2) Changing display-command-table-menu to present the menu elements as command-manu-item
3) Providing a presentation-translator for this presentation-type whose documentation function returns just the name of the command
(all these are in /core/clim-code/gadgets/menu

4) Changing the accept method for comand so that the body is wrapped in an with-input-context so that if a menu-item is clicked on, it can br handled specially; in particular so that the partial command parser is called.

### DIFF
--- a/Core/clim-core/commands.lisp
+++ b/Core/clim-core/commands.lisp
@@ -1309,6 +1309,48 @@ examine the type of the command menu item to see if it is
             (position *unsupplied-argument-marker* command)))
 	  (t (values command type)))))
 
+(define-presentation-method accept ((type command) stream
+				    (view textual-view)
+				    &key)
+  (setq command-table (find-command-table command-table))
+  (let ((start-position (and (input-editing-stream-p stream)
+                             (stream-scan-pointer stream)))
+        ;; this also requires some thought, but I suspect that
+        ;; we can just kludge it this way in this presentation type,
+        ;; because we can't think of any other presentation types
+        ;; that would establish a shadowing context within the one
+        ;; established by ACCEPT-1.
+        (replace-input-p nil))
+    (multiple-value-bind (object new-type)
+        ;; We establish a new input context so that clicks throw to us
+        ;; rather than to the input context established in ACCEPT-1.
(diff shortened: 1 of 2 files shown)


Core/clim-core/commands.lisp
```diff
@@ -49,7 +49,7 @@
   ((menu-name :reader command-menu-item-name :initarg :menu-name)
    (type :initarg :type :reader command-menu-item-type)
    (value :initarg :value :reader command-menu-item-value)
-   (documentation :initarg :documentation)
+   (documentation :initarg :documentation :initform nil)
```
dkochmanski: please write a more descriptive commit message

"Step one in fixing command menu problems" doesn't describe the change. You could write (for instance):

```
presentations: introduce a new presentation type command-menu-item

- add a new presentation type
- add a translator from the command-menu-item to command
- use the new presentation-type when displaying command table menu
- [list other changes]

These change is necessary to fix a bug mentioned in #955, that is
[put a brief description of the problem]. Closes #955.
```

It looks to me that all three commits do one atomic change (i.e we can't just take
the first commit and use it, because other changes does not follow), in that case
please squash all three commits into one and take all changes into account in
description. Atomic commits are important in case when we want to bisect changes
in order to find a regression.

PR #958, opened 2020-01-30
text-style-mapping: do not cache partially specified text styles

dkochmanski:
*DEFAULT-TEXT-STYLE* may change and then mapping would be invalid.

Also:

- add a method to the function TEXT-STYLE-COMPONENTS specialized on
  DEVICE-FONT-TEXT-STYLE returning (values :device :device :device);
  method must be defined for all subclasses of text-style

- be more robust in parse-text-style* by returning
  DEVICE-FONT-TEXT-STYLE as is; this text style is always considered
  to be fully specified.

### DIFF
--- a/Backends/common/fonts.lisp
+++ b/Backends/common/fonts.lisp
@@ -177,36 +177,46 @@ of letters specified in a separate kerning-table."))
 (defgeneric climb:font-glyph-dx (font code))
 (defgeneric climb:font-glyph-dy (font code))
 
-
-;; clim:make-device-font-text-style
 (declaim (inline climb:normalize-font-size))
 (defun climb:normalize-font-size (size)
-  (cond ((numberp size)
-         (round (max size 2)))
-        ((null size)
+  (cond ((null size)
          (let ((size* (text-style-size *default-text-style*)))
            (etypecase size*
              (number size*)
-             (symbol (getf +font-sizes+ size*)))))
-        (T
-         (or (getf +font-sizes+ size nil)
-             (error "~s is not a valid text style size!" size)))))
+             (symbol (getf +font-sizes+ size* nil)))))
+        ((eq size :smaller)
+         (getf +font-sizes+ (text-style-size *default-text-style*) nil))
+        ((eq size :larger)
+         (getf +font-sizes+ (text-style-size *default-text-style*) nil))
+        ((realp size)
+         (round (max size 2)))
+        ((getf +font-sizes+ size nil))
+        (t
+         (error "~s is not a valid text style size!" size))))
 
 (defun parse-text-style* (style)
   "Returns complete text-style without NIL components and with numeric size."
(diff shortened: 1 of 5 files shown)


Core/clim-basic/drawing/medium.lisp
```diff
@@ -216,10 +218,15 @@
...
-                  (call-next-method)))
+  ;; Cache mappings only for fully specified text styles.
+  (if (or (device-font-text-style-p text-style)
+          (multiple-value-bind (family face size)
+              (clim:text-style-components text-style)
+            (and family face size)))
+      (ensure-gethash text-style
+                      (port-text-style-mappings port)
+                      (call-next-method))
+      (call-next-method)))
```
scymtym: Could we move the `parse-text-style*` call that many (all?) backends do into this else branch?

dkochmanski: parse-text-style* is just an utility defined for ttf renderer, backends are free to use it but I don't think it belongs here.

Backends/common/fonts.lisp
```diff
@@ -198,7 +198,9 @@ of letters specified in a separate kerning-table."))
 
 (defun parse-text-style* (style)
   "Returns complete text-style without NIL components and with numeric size."
-  (cond ((text-style-p style)
+  (cond ((device-font-text-style-p style)
+         style)
+        ((text-style-p style)
          (multiple-value-bind (family face size)
              (clim:text-style-components style)
            (if (and family face size (numberp size))
```
scymtym: I suggest something like this:

```lisp
(defun parse-text-style* (style)
  "Returns complete text-style without NIL components and with numeric size."
  (flet ((new-text-style (family face font-size)
           (let ((default *default-text-style*))
             (make-text-style (or family (text-style-family default))
                              (or face (text-style-face default))
                              (climb:normalize-font-size font-size)))))
    (cond ((device-font-text-style-p style)
           style)
          ((text-style-p style)
           (multiple-value-bind (family face size)
               (text-style-components style)
             (if (and family face size (realp size))
                 style
                 (new-text-style family face size))))
          ((null style)
           (parse-text-style* *default-text-style*))
          ((and (listp style) (alexandria:length= 3 style))
           (apply #'new-text-style style))
          (t
           (error "Invalid text style specification ~S." style)))))
```

dkochmanski: I've pushed change with something in a similar spirit.

scymtym: Why not put the repeated code into a local function?

dkochmanski: the repeated code is a single function call hence it doesn't save much to me. but it doesn't clutter the code either, so I've force-pushed version with flet. if you are satisfied with the branch please merge it.

dkochmanski:
Lgtm, thanks

Issue #945, opened 2020-01-18
Proposal: WITH-OUTPUT-TO-PDF-FILE macro


admich:
The PDF backend is used through `WITH-OUTPUT-TO-PDF-STREAM` that is
based on the analogous macro for Postscript backend defined in the
Spec.

I think it is better to define for PDF backend a
`WITH-OUTPUT-TO-PDF-FILE` with a string or a pathname as argument
instead of a stream. In this way the user don't have to open the file
stream (the user needs also to known that the stream must be open with
keyword `:element-type '(unsigned-byte 8)`). If we remove the
WITH-OUTPUT-TO-PDF-STREAM McCLIM can remove the dependency to
FLEXI-STREM system.

The difference from PDF to Postscript is giustified because PDF is a
file format so the output is ever a file, instead the Postscript
output can be send on a stream directly atttached to a postscript
printer.

We can also add a `WITH-OUTPUT-TO-POSTSCRIPT-FILE` macro without remove
the `WITH-OUTPUT-TO-POSTSCRIPT-STREAM` defined in the Specification.

dkochmanski:
as suggested on the mailing list, imo we should use -stream variant where file-var may be coerced to a stream when it contains a pathname designator by opening a file.

scymtym:
Fixed by #959.

PR #959, opened 2020-02-01
Pdf graft

### LINKED ISSUE #944: PDF backend problem with :orientation :landscape
With code:
```
(with-open-file (s "tmp.pdf" :direction :output :if-exists :supersede
                   :element-type '(unsigned-byte 8))  
  (clim-pdf::with-output-to-pdf-stream (stream s :device-type :a4 :orientation :landscape)
    (draw-rectangle* stream 10 10 200 100 :filled nil)
    (draw-text* stream "first point" 10 10 :align-x :left :align-y :top)
    (draw-text* stream "second point" 200 100 :align-x :right :align-y :bottom)))
```
in the pdf file the rectangle is rotated and the text is not visible. With :orientation :portrait instead the pdf output is right.

admich:
This PR improve the PDF backend:
1. `with-output-to-pdf-stream` (and also `with-output-to-postscript-stream`) can accept also a  string with a file name or a pathname as file-stream. (issue #945) 
2. Now `:orientation :landscape`, `:scale-to-fit t`, and `:trim-page-to-output-size t` works. (issue #944)
3. Now the `clim-pdf-stream` is a child of a `pdf-graft`
4. Now it is possible to use graft with units different from `:device` (e.g. `:inches`, `:millimeters`, `screen-sized`)
5. Some tests added

Fix #944 #945


Backends/PDF/sheet.lisp
```diff
@@ -166,17 +161,29 @@
   (loop for child in (sheet-children sheet)
      do (invalidate-cached-transformations child)))
 
+(defmethod graft-units-transformation ((graft pdf-graft))
+  (case (graft-units graft)
```
scymtym: This should probably be `ecase`.

Backends/PDF/sheet.lisp
```diff
@@ -166,17 +161,29 @@
...
+    (:device +identity-transformation+)
+    (:inches (make-scaling-transformation* (* 72) (* 72)))
+    (:millimeters (make-scaling-transformation* (/ 720 254) (/ 720 254)))
+    (:screen-sized (make-scaling-transformation* (* (graft-width graft)) (* (graft-height graft))))))
+
+(defmethod graft-orientation-transformation ((graft pdf-graft))
+  (case (graft-orientation graft)
```
scymtym: This should probably be `ecase`.

Backends/PostScript/sheet.lisp
```diff
@@ -58,6 +58,19 @@
                                                 ,file-stream ,@options))))
 
 (defun invoke-with-output-to-postscript-stream (continuation
+                                                file-stream &rest args &key device-type
+                                                                 multi-page scale-to-fit
+                                                                 (orientation :portrait)
```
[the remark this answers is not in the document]
admich: I follow your second suggestion.

admich:
I have made all your requested changes.

scymtym:
> I have made all your requested changes.

Thank you.

One more question: Is the protocol consisting of `graft-units-transformation` and `graft-orientation-transformation` specific to the PDF backend or is it intended to be used in other places as well? In the second case, there should be generic function definitions with documentation strings and the function names should be in a different package.

Backends/PDF/class.lisp
```diff
@@ -53,23 +50,15 @@
...
                            sheet-mute-repainting-mixin
-                           ;; ?
-                           mirrored-sheet-mixin
-                           ;; FIXME: Tim Moore suggested (2006-02-06,
-                           ;; mcclim-devel) that this might better be
-                           ;; a superclass of
-                           ;; STANDARD-OUTPUT-RECORDING-STREAM.  This
-                           ;; should be revisited when we grow another
-                           ;; non-interactive backend (maybe a cl-pdf
-                           ;; backend?).  -- CSR.
```
dkochmanski: please adopt port/graft changes to postscript backend too.

as a side note: we should create an output-stream-backend with base classes meant to be subclassed for both postscript and pdf (possibly also raster) at some point of time.

Backends/PDF/sheet.lisp
```diff
@@ -129,41 +116,85 @@
...
+               (make-reflection-transformation* 0 0 1 0)))))
+
+(defmethod sheet-native-transformation ((sheet pdf-graft))
+  (with-slots (native-transformation) sheet
+    (unless native-transformation
+      (setf native-transformation
+            (compose-transformations
+             (graft-orientation-transformation sheet)
+             (graft-units-transformation sheet))))
+    native-transformation))
```
dkochmanski: please define graft-*-transformation functions with defun, it is a local functionality which is not exported, so adding implicit generic functions doesn't make sense now.

When we sort out how we should handle transformations (the other ticket) this protocol may be necessary and then it will land in clime or climb package, but for now there is no benefit of that.

Backends/PDF/class.lisp
admich: I agree with you that we need an `output-stream-backend`, and for this reason I think that we can mantain the postscript backend without modification now and waiting for future changing. For exampe the new postscript graft could use the graft-*-transformation protocol that you suggest in the next request to define with `defun` because it is local to pdf backend...but if it will be used also by the postscript backend...it will be no more local.

Backends/PDF/sheet.lisp
admich: If you agree with my reply to the previous request I make this changing, otherwise I will work a little more on the postscript backend.

Backends/PDF/class.lisp
dkochmanski: I think that the best course of action is to duplicate changes from pdf in postscript backend -- it will be easier to create the output-stream-backend if both backends do not diverge too much.

Backends/PDF/sheet.lisp
dkochmanski: ditto, please make it defun and duplicate as defun in postscript backend (they are different packages, so there won't be collision).

admich:
I made the changes in ps backend and use defun for graft-*-transformation.
I think that before merging this PR we can also make some other changes, because the pdf and ps backend still have some differences that can be eliminated, but I need your feedback and suggestion. They are:
1. ps now didn't accept device-type like '(width height), we can add dthis functionality.
2. the trim-page-to-output-size effect is obtained in ps with :device-type :eps and in pdf with :trim-page-to-output-size t. We can remove in with-output-to-pdf-strem the :trim-page-to-output-size keyword and use some value (like :trim or :trim-page-to-output-size) for :device-type to obtain the effect.

admich:
and:
3. In with-output-to-postscript-stream there isn't a way to choose the graft units. So or I add the :units keyword also in with-output-to-postscript-stream (but in the spec there is  with-output-to-postscript-stream without :units keyword), or we remove :units keyword from both with-output-to-*-stream and we  allow the :device-type keyword to include units information like :device-type '(:a4 :millimeters)

dkochmanski:
Thank you. I think that all these things should be addressed when the common base for output-only backend is defined. Notice, that i.e scale-to-fit is not implemented at all and multi-page is simply ignored (in principle it should break the page when output crosses its boundaries).

dkochmanski:
it looks good to me for now, I imagine that the output-only backend as a common superclass is the next step. raster output should fall in this category too. I'm not merging it because there is still @scymtym's pending review.

admich:
Ok, anyway scale-to-fit now works for both ps and pdf. I added also tests for this. Instead multi-page is completly unimplemented as you stated.

scymtym:
@admich If this is fine with you, I would now add this to the master branch with the following additional changes:
* typo fixes in commit messages
* remove `:stream` initarg from the `stream` slot of both port classes since the stream is extracted from the server path, not passed as an initarg
* change the `make-graft` method in the Postscript backend to `climb:make-graft`

admich:
@scymtym it's fine for me.

scymtym:
Integrated into master, thank you.

PR #963, opened 2020-02-11
text-style-mapping: specialize on cons only in core

dkochmanski:
Method specialized on cons parses the text style and calls
text-style-mapping on the parsed text style. Fixes a regression in a
demo "German Towns".


dkochmanski:
cherry picked in fix-various

PR #964, opened 2020-02-11
Fix drag and drop while shift modifier is active

### LINKED ISSUE #961: distribute-event does not distribute motion events when shift is pressed
steps to reproduce

1. start drag and drop demo
2. create a circle
3. start dragging it
4. while dragging press shift (don't release it)
5. wiggle the pointer

- Expected: dnd feedback follows the pointer
- Observed: feedback picture does not move until shift is released

initial glance shows that:
- distribute-event is called on the motion event
- dispatch-event is not called when that happens

scymtym:
Without this change, instances of `text-selection-mixin` rerouted certain pointer events to `eos/shift-{release,drag}` as soon as the shift modifier was active. This behavior was wrong because it did not consider whether a text selection activity was in progress or not (which is indicated by the `dragging-p` slot of the pane and thus easily checked).

This will fix #961

### DIFF
--- a/Core/clim-basic/extended-streams/text-selection.lisp
+++ b/Core/clim-basic/extended-streams/text-selection.lisp
@@ -166,10 +166,10 @@
       (repaint-markings pane old-markings (slot-value pane 'markings)))))
 
 (defun shift-rl-click-event-p (event)
-  (let ((b (pointer-event-button event)))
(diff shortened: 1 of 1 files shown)


dkochmanski:
lgtm, thanks!

PR #965, opened 2020-02-12
Fix 942

dkochmanski:
This PR fixes the issue with duplicated definitions.

### DIFF
--- a/Backends/CLX/medium-xrender.lisp
+++ b/Backends/CLX/medium-xrender.lisp
@@ -6,7 +6,8 @@
 (in-package #:clim-clx)
 
 (defclass clx-render-medium (clx-medium
-                             climb:multiline-text-medium-mixin)
+                             climb:multiline-text-medium-mixin
+                             climb:font-rendering-medium-mixin)
   ((picture :initform nil)))
 
 (defun clx-render-medium-picture (medium)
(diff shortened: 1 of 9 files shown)


Backends/CLX/medium-xrender.lisp
```diff
@@ -6,7 +6,8 @@
 (in-package #:clim-clx)
```
scymtym: Commit message: "backens" -> "backend" or "backends".

Core/clim-basic/decls.lisp
```diff
@@ -1201,11 +1203,6 @@ rendered on MEDIUM with the style LINE-STYLE."))
 
 ;;;
 
-(defgeneric text-style-character-width (text-style medium char))
-;; fall back, where to put this?
-(defmethod text-style-character-width (text-style medium char)
-  (text-size medium char :text-style text-style))
-
```
scymtym: I'm not sure this change is an improvement. I usually consider unspecialized default methods as part of the protocol. I don't have a super strong opinion on this, though.

dkochmanski: fixed (I've put it in :method clause of defgeneric to signify the fact that it is part of this protocol function). n.b it is an internal protocol which is expected to be implemented by backends for purpose of extended output stream width specifications (i.e there is no clim:text-style-character-width).

Backends/CLX/medium-xrender.lisp
dkochmanski: fixed

PR #969, opened 2020-02-21
filling-output: better default for fill-width

dkochmanski:
Previous default value was 80 characters wide, while stream has its
own line width. It is better default, because when text-margins are
specified then filling-width does not ignore them unless explicitly
told to.

### DIFF
--- a/Core/clim-basic/extended-streams/stream-output.lisp
+++ b/Core/clim-basic/extended-streams/stream-output.lisp
@@ -315,7 +315,7 @@
              (eol-action (stream-end-of-line-action stream))
              (eol-p (> (+ cx width) right-margin)))
         (when (or (null eol-p) (member eol-action '(:allow :scroll)))
-          (stream-write-output stream string nil start end)
+          (stream-write-output stream string start end)
           (incf cx width)
           (setf (cursor-position (stream-text-cursor stream)) (values cx cy))
           (when (and (> cx right-margin)
@@ -328,50 +328,48 @@
         ;; closing the text-output-record and have multiline records to
         ;; allow gimmics like a dynamic reflow). Also text-style doesn't
         ;; change until the end of this function. -- jd 2019-01-10
-        (let* ((width (if (text-style-fixed-width-p text-style medium)
-                          (text-style-width text-style medium)
-                          (lambda (string start end)
-                            (text-size medium string
-                                       :text-style text-style
-                                       :start start :end end))))
-               (splits (line-breaks string width
-                                    :initial-offset (- cx left-margin)
-                                    :margin (- right-margin left-margin)
-                                    :break-strategy (ecase eol-action
-                                                      (:wrap NIL)
-                                                      (:wrap* T))
-                                    :start start :end end)))
-          (do ((start start (car split))
-               (split splits (rest split)))
-              ((null split)
-               (stream-write-output stream string nil start end)
-               (setf (cursor-position (stream-text-cursor stream))
-                     (values (+ left-margin
-                                (stream-string-width stream string
-                                                     :start start :end split
-                                                     :text-style text-style))
-                             (nth-value 1 (stream-cursor-position stream)))))
-            (ecase eol-action
-              (:wrap  (stream-write-output stream string nil start (car split)))
-              (:wrap* (let ((pos (position #\space string
-                                           :from-end t :start start :end (car split)
-                                           :test-not #'char=)))
-                        (when pos (incf pos))
-                        (stream-write-output stream string nil start (or pos (car split))))))
-            ;; print a soft newline
-            (seos-write-newline stream t)))))))
+        ;;
+        ;; Writing a newline may cause the cursor increment, so we
+        ;; need to compute split for each line after the soft newline
+        ;; has been written. -- jd 2020-03-01
+        (loop with width = (if (text-style-fixed-width-p text-style medium)
+                               (text-style-width text-style medium)
+                               (lambda (string start end)
+                                 (text-size medium string
+                                            :text-style text-style
+                                            :start start :end end)))
+              with margin = (- right-margin left-margin)
+              with cursor = (stream-text-cursor stream)
+              with break  = (ecase eol-action
+                              (:wrap nil)
+                              (:wrap* t))
+              for offset = (- (cursor-position cursor) left-margin)
+              for split  = (car (line-breaks string width
+                                             :count 1
+                                             :initial-offset offset
+                                             :margin margin
+                                             :break-strategy break
+                                             :start start :end end))
+              do (stream-write-output stream string start split)
+                 (when (= split end)
+                   (setf (cursor-position cursor)
+                         (values (+ left-margin
+                                    (stream-string-width stream string
+                                                         :start start :end split
+                                                         :text-style text-style))
+                                 (nth-value 1 (cursor-position cursor))))
+                   (return))
+                 (seos-write-newline stream t)
+                 (setf start split))))))
 
-(defgeneric stream-write-output (stream line string-width &optional start end)
+(defgeneric stream-write-output (stream line &optional start end)
   (:documentation
-   "Writes the character or string LINE to STREAM. This function produces no more
-than one line of output i.e., doesn't wrap. If STRING-WIDTH is non-nil, that is
-used as the width where needed; otherwise STREAM-STRING-WIDTH will be called."))
+   "Writes the character or string LINE to STREAM. This function
+produces no more than one line of output i.e., doesn't wrap."))
 
 ;;; The cursor is in stream coordinates.
-(defmethod stream-write-output ((stream standard-extended-output-stream)
-                                line string-width
+(defmethod stream-write-output ((stream standard-extended-output-stream) line
                                 &optional (start 0) end)
-  (declare (ignore string-width))
   ;; Do not capture medium transformation - this is a stream operation and we
   ;; draw at the current cursor position. -- jd 2019-01-04
   (with-identity-transformation (stream)
@@ -391,11 +389,11 @@ used as the width where needed; otherwise STREAM-STRING-WIDTH will be called."))
         (end (or end (length string))))
     (with-cursor-off stream
       (loop for i from start below end do
-           (when (char= #\Newline
-                        (char string i))
-             (seos-write-string stream string seg-start i)
-             (seos-write-newline stream)
-             (setq seg-start (1+ i))))
+        (when (char= #\Newline
+                     (char string i))
+          (seos-write-string stream string seg-start i)
+          (seos-write-newline stream)
+          (setq seg-start (1+ i))))
       (seos-write-string stream string seg-start end))))
 
 (defmethod stream-character-width ((stream standard-extended-output-stream) char
--- a/Core/clim-basic/extended-streams/text-formatting.lisp
+++ b/Core/clim-basic/extended-streams/text-formatting.lisp
@@ -179,10 +179,10 @@
                             (etypecase ,after-line-break
                               (string   (write-string ,after-line-break ,stream))
                               (function (funcall ,after-line-break ,stream soft-newline-p))))))
-                      ;; When after-line-break goes beyond the previous
-                      ;; position we advance the cursor.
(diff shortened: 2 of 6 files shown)


dkochmanski:
I've noticed some more issues with filling-output, will push to this PR and remove [wip] tag when done.

dkochmanski:
OK, this is ready for being merged, I'm satisfied with the branch.

Core/clim-basic/utils.lisp
```diff
@@ -808,161 +808,76 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+;;; T, then we return the rightmost opportunity which does not violate
+;;; hard line breaks. Second value indicates the kind of a line break.
+(defun line-end (string break-fn start limit from-end)
+  (loop with opportunity = nil
+        for i from start below limit
+        do (case (maybe-funcall break-fn string start i)
+             (:hard (return (values (1+ i) :hard)))
+             (:soft (if from-end
+                        (setf opportunity (1+ i))
+                        (return (values (1+ i) :soft)))))
```
scymtym: This looks like it anticipates `break-fn` being `nil` and just returns `limit` in that case. Why is this allowed? Couldn't the caller just not call this function in such cases?

Core/clim-basic/utils.lisp
```diff
@@ -808,161 +808,76 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+                                       :soft)))
+                                  (vector
+                                   (lambda (string start index)
+                                     (declare (ignore start string))
+                                     (when (find index break-strategy)
+                                       :soft)))))
+                      (initial-offset (or initial-offset 0))
+                      (margin (or margin
+                                  (* 80 (funcall width-fn "m" 0 1))))
+                      (start (or start 0))
```
scymtym: I think this is going way too far. It separates the lambda list from the documentation and the invariants at the beginning of the body by more than a screen.

The computation for `break-fn` might warrant helper function in any case.

Core/clim-basic/utils.lisp
```diff
@@ -808,161 +808,76 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+;;; Returns an index of the character _before_ which we should break
+;;; the line (may be used as END argument in SUBSEQ). When FROM-END is
+;;; T, then we return the rightmost opportunity which does not violate
+;;; hard line breaks. Second value indicates the kind of a line break.
+(defun line-end (string break-fn start limit from-end)
+  (loop with opportunity = nil
+        for i from start below limit
+        do (case (maybe-funcall break-fn string start i)
+             (:hard (return (values (1+ i) :hard)))
+             (:soft (if from-end
```
scymtym: Please describe in the comment what `break-fn` is allowed to return.

Core/clim-basic/utils.lisp
```diff
@@ -983,36 +900,56 @@ BREAK-STRATEGY may be:
...
-                                     collect i)
-                                  'vector))
-                         (vector
-                          break-strategy))))
-    (%line-breaks-2 string width initial-offset margin start end opportunities)))
+  (check-type count (or null (integer 0)))
+  (when (and count (zerop count))
+    (return-from line-breaks))
+  ;; Margin explicitly specified as NIL (break only on hard breaks).
+  (when (null margin-p)
```
scymtym: I think this should be `(not margin-p)` since it is a Boolean.

Core/clim-basic/utils.lisp
```diff
@@ -808,161 +808,76 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
+             (:hard (return (values (1+ i) :hard)))
+             (:soft (if from-end
+                        (setf opportunity (1+ i))
+                        (return (values (1+ i) :soft)))))
+        finally
+           (if (null opportunity)
+               (return (values limit :emergency))
+               (return (values opportunity :soft)))))
+
+(defun line-breaks (string width
```
scymtym: Maybe it didn't load everything properly, but `(line-breaks "foo bar baz" 8)` seems to go into an infinite loop.

Core/clim-basic/utils.lisp
```diff
@@ -983,36 +900,56 @@ BREAK-STRATEGY may be:
...
-                                     when (member (char string i) break-strategy)
-                                     collect i)
-                                  'vector))
-                         (vector
-                          break-strategy))))
-    (%line-breaks-2 string width initial-offset margin start end opportunities)))
+  (check-type count (or null (integer 0)))
+  (when (and count (zerop count))
+    (return-from line-breaks))
+  ;; Margin explicitly specified as NIL (break only on hard breaks).
```
scymtym: Wouldn't "explicitly specified as NIL" be `(and margin-p (null margin))`?

Tests/utils.lisp
```diff
@@ -17,149 +17,41 @@
 
 ;;; Line splitting utility
 ;;; ============================================================================
-
```
scymtym: I would prefer squashing the previous commit and this one into one new commit.

Also, the coverage achieved by this test seems insufficient: https://techfak.de/~jmoringe/9679c6b9cf9e6c559fa1ab6e1531f421.html

Core/clim-basic/extended-streams/stream-output.lisp
```diff
@@ -344,34 +344,32 @@
           (do ((start start (car split))
                (split splits (rest split)))
               ((null split)
-               (stream-write-output stream string nil start end)
+               (stream-write-output stream string start end)
                (setf (cursor-position (stream-text-cursor stream))
                      (values (+ left-margin
                                 (stream-string-width stream string
                                                      :start start :end split
                                                      :text-style text-style))
```
scymtym: Unless I'm reading this wrong, the string width is computed here a second time. Maybe we should use the optimization to avoid the repeated computation instead of removing it.

Core/clim-basic/extended-streams/stream-output.lisp
```diff
@@ -328,38 +328,39 @@
...
+                               (text-style-width text-style medium)
+                               (lambda (string start end)
+                                 (text-size medium string
+                                            :text-style text-style
+                                            :start start :end end)))
+              with margin = (- right-margin left-margin)
+              with cursor = (stream-text-cursor stream)
+              with break  = (ecase eol-action
+                              (:wrap NIL)
+                              (:wrap* T))
```
scymtym: Can you downcase `NIL` and `T`?

Core/clim-basic/extended-streams/stream-output.lisp
```diff
@@ -328,38 +328,39 @@
...
+              for offset = (- (cursor-position cursor) left-margin)
+              for split  = (car (line-breaks string width
+                                             :count 1
```
scymtym: Is the increased cost of doing this repeatedly somewhat compensated by the `:count 1`?

scymtym:
`(clim-demo:demodemo)` » Misc. Tests » Line Widths results in "The variable CLIM-CLX::^CLEANUP is unbound." Not sure whether this is a regression or even related to the pull request.

Core/clim-basic/utils.lisp
```diff
@@ -808,161 +808,76 @@ a flag CLOSED is T then beginning and end of the list are consecutive too."
...
-;;; in "real" languages lines may /need to/ or /must to/ be split under
-;;; various conditions - hyphen, hard space, infix numeric separators
-;;; etc. Optimally this should be implemented in cl-unicode and used
-;;; from there.  [1] https://unicode.org/reports/tr14/ -- jd 2019-01-08
-(defun line-breaks (string width &key (break-strategy t) initial-offset margin (start 0) end)
-  "Function takes a string and returns a list of indexes where it should be split.
+;;; Returns an index of the character _before_ which we should break
+;;; the line (may be used as END argument in SUBSEQ). When FROM-END is
+;;; T, then we return the rightmost opportunity which does not violate
+;;; hard line breaks. Second value indicates the kind of a line break.
```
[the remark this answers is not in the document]
dkochmanski:
```
foo bar\nqux bam
   ^   ^^    ^
```

In this string we have three line break opportunities, between foo and bar, between bar and qux and between qux and bam. Rightmost break opportunity is between qux and bam, but newline is a hard line break, so we pick that. (otherwise we would not break on a mandatory break character). Previous version of the code assumed, that there are no newlines and such in the text, this version is more robust (and in the future i'd like to use it earlier than in seos-write-string, namely in stream-write-string during line breaks).

Core/clim-basic/utils.lisp
dkochmanski: break-fn should never be nil, that's a leftover from testing. fixed.

Core/clim-basic/utils.lisp
dkochmanski: OK, I've moved it to a separate function.

Core/clim-basic/extended-streams/stream-output.lisp
dkochmanski: you are correct - line-breaks calls it at least once, I don't know what optimization you have in mind though.

Core/clim-basic/extended-streams/stream-output.lisp
dkochmanski: fixed.

Core/clim-basic/extended-streams/stream-output.lisp
dkochmanski: it is, that was one of reasons behind rewriting break-lines utility.

Core/clim-basic/utils.lisp
dkochmanski: fixed (margin is bound in aux according to margin-p).

Tests/utils.lisp
dkochmanski: squashed. as of coverage, it is hard to add by hand tests to it. fwiw it is also tested in more elaborate way in seos baseline regression test and in misc-tests.

Core/clim-basic/utils.lisp
dkochmanski: the other value is NIL. I've added `(nil)` clause and used ecase.

Core/clim-basic/extended-streams/text-formatting.lisp
```diff
@@ -154,7 +154,8 @@
                                 :after-line-break-initially
                                 :after-line-break-subsequent))
     (with-gensyms (continuation old-fresh-line-fn fresh-line-fn
-                                initial-indent ink text-style)
+                                initial-indent ink text-style
+                                cx cy nx ny)
```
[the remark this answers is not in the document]
dkochmanski: fixed

dkochmanski:
lack of cleanup is unrelated to this change. it is a regression exhibited by tiling background in one of tests (there is another issue with background repainting, but it is something even older). I've disabled offending part in misc-tests and added a comment (see last commit).

With that I think I've responded to all remarks (with some unresolved questions).

Core/clim-basic/utils.lisp
```diff
@@ -983,36 +905,56 @@ BREAK-STRATEGY may be:
...
-  (unless end (setq end (length string)))
-  (unless margin (setq margin (* 80 (etypecase width
-                                      (function (funcall width "m" 0 1))
-                                      (number width)))))
-  (unless initial-offset (setq initial-offset 0))
+to the BREAK-STRATEGY whenever it assigns any meaning to to them.
+
+COUNT specifies how many breaks we want to collect."
+  (array-in-bounds-p string start)
+  (array-in-bounds-p string (1- end))
```
scymtym: This doesn't seem to have any effect (besides potential type errors).

dkochmanski: fixed

dkochmanski:
I've relaxed one of your tests regarding newline (see last commit). Also I've fixed the issue for a negative index -- I believe that all remarks are addressed now (and randomized tests seem to pass too).

dkochmanski:
I'm merging it for lack of further activity (there are no remaining issues, only a doubt whether we should somehow exclude newline characters when computing line break index (that would require returning two values or punting on hard line breaks and would be different from what Unicode mandates).

PR #973, opened 2020-02-23
Fix completion for logical pathnames

heshrobe:
This fixes completions for all types of pathnames.
This addresses the same issues as 954 which has been closed.

Currently if the pathname begins with ~/ (meaning start from my home directory)
the pathname will not complete;

Also If the pathname is a logical pathname and there
is a file in the directory that the pathname resolves to and that file's name
includes a character that is illegal for a logical pathname then the pathname
will not complete.

The original filename-completer function is now broken into 3 steps:

1) Given the string to be complered, figure out what directory is
implied.  This considers both logical pathnames and pathnames relative to the
home directory.

2) In that directory collect all the files whose names begin with the string.
If the pathname is a logical pathname then filter out all files whose names
include characters that are illegal in logical pathnames (i.e. everything but
alpha numerics and dash).

3) Given the list of candidates from step 2 call the completer

This can be tested interactively (e.g. in the lisp-listener) by creating a
directory  and a logical host translating to it:
```
(progn
  (ensure-directories-exist "~/foobar/bar/")
  (setf (logical-pathname-translations "foo")
	  '(("**;*.*"	"~/foobar/**/*.*")))

  (with-open-file (f "~/foobar/bar/baz_bar.lisp" :direction :output :if-exists :supersede :if-does-not-exist :create ))
  (with-open-file (f "~/foobar/bar/baz-bar.lisp" :direction :output :if-exists :supersede :if-does-not-exist :create)))
```

1. (accept 'pathname) and typing ~/foobar/b followed by tab should complete to ~/foobar/bar
2. (accept 'pathname) and typing foo:b followed by tab should complete to foo:bar;
3. (accept 'pathname) and typing foo:bar;baz followed by tab should complete foo:bar;baz-bar.lisp

### DIFF
--- a/Core/clim-core/presentation-defs.lisp
+++ b/Core/clim-core/presentation-defs.lisp
@@ -1607,82 +1607,110 @@ protocol retrieving gestures from a provided string."))
 (defmethod presentation-type-of ((object pathname))
   'pathname)
 
-(defun filename-completer (string action)
+(defun filename-completer-get-directory (string)
   (flet
       ((deal-with-home (pathname his-directory)
-	 ;; SBCL (and maybe others) treat "~/xxx" specially, returning a pathname
-	 ;; whose directory is (:ABSOLUTE :HOME xxx)
-	 ;; But if you call Directory on that pathname the returned list
-	 ;; are all complete pathnames without the :Home part!.
-	 ;; So this replaces the :HOME with what it actually means
-	 (let* ((home-env-variable (get-environment-variable "HOME"))
-		(home (loop for pos = 1 then (1+ next-pos)
-			 for next-pos = (position #\/ home-env-variable :start pos)
-			 collect (subseq home-env-variable pos next-pos)
-			 until (null next-pos)))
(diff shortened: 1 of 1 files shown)


dkochmanski:
sorry for the long wait. your branch works for me and after testing I've found no problems with it. thank you!

heshrobe:
Daniel. No problem.  Thanks for all the work you put into this.  Right now, in my copious free time, I’m working on a slightly better version of format-graph-from-roots that has a pass in which the layout goes through an optimization pass.  I have a version that runs in the original version (Allegro) but the McClim version is somewhat different so it will take a bit more work.  This will actually be an addition (a new style of graph) rather than a patch.

> On Apr 19, 2020, at 12:35, Daniel Kochmanski <notifications@github.com> wrote:
> 
> 
> Merged #973 into master.
> 
> —
> 

Howard Shrobe

Director CyberSecurity@CSAIL
hes@csail.mit.edu
617-253-7877 work
617-312-6439 cell

PR #978, opened 2020-03-04
Clouseau Improvements

scymtym:
1. Export classes and adjust documentation so that the Clouseau tutorial works again

2. Cleanup inconsistencies between `ensure-place` and `formatting-place` (as pointed out by @robert-strandh )

3. When printing slot names, omit package if it is the same as the "context package"

4. Better support for reinitialization of Clouseau structures

5. Prettier and more readable history

### DIFF
--- a/Apps/Clouseau/src/application.lisp
+++ b/Apps/Clouseau/src/application.lisp
@@ -38,7 +38,10 @@
 ;;; Application frame
 
 (define-application-frame inspector ()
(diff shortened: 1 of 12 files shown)


dkochmanski:
lgtm, I have no critical remarks, but I have just a surface understanding of clouseau.

scymtym:
Thanks.

PR #979, opened 2020-03-04
presentations: use information from the patched defclass on ecl

dkochmanski:
It is allowed for define-presentation-type to assume, that class has
been already defined (that is -- defclass must be evaluated before the
matching define-presentation-type). McCLIM patches defclass to store
information about defined classes at compilation time to allow them to
be defined both as top-level forms in the same file without additional
eval-when by some implementations which use that information; add ECL
to this set.


dkochmanski:
I hereby approve this pull request :)

PR #983, opened 2020-03-25
I think I figured out why :clx-ttf didn't work in Windows

pouar:
This seems to get it working

### DIFF
--- a/Backends/CLX/port.lisp
+++ b/Backends/CLX/port.lisp
@@ -72,13 +72,11 @@
 	    :protocol protocol))))
 
 (defun helpfully-automagic-clx-server-path (port-type)
-  #+windows
-  (parse-clx-server-path '(:clx :host "localhost" :protocol :internet))
-  #-windows
   (restart-case (automagic-clx-server-path port-type)
     (use-localhost ()
-      :report "Use local unix display"
-      (parse-clx-server-path `(,port-type :host "" :protocol :unix)))))
+      :report "Use local display"
+      #+windows (parse-clx-server-path `(,port-type :host "localhost" :protocol :internet))
+      #-windows (parse-clx-server-path `(,port-type :host "" :protocol :unix)))))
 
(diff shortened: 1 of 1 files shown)


Backends/CLX/port.lisp
```diff
@@ -72,13 +72,10 @@
...
 
 (defun helpfully-automagic-clx-server-path (port-type)
-  #+windows
-  (parse-clx-server-path '(:clx :host "localhost" :protocol :internet))
-  #-windows
   (restart-case (automagic-clx-server-path port-type)
     (use-localhost ()
       :report "Use local unix display"
-      (parse-clx-server-path `(,port-type :host "" :protocol :unix)))))
+      (parse-clx-server-path `(,port-type :host #-windows "" #+windows "localhost" :protocol #-windows :unix #+windows :internet)))))
```
dkochmanski: please rewrite it to feature only single feature condition:

```
(restart-case (…)
  (use-localhost ()
    :report "Use a local display."
    #-windows (parse-clx-server-path …)
    #+windows (parse-clx-server-path ...)))
```

note, that it would be enough to replace `:clx` with `,port-type`, but it may be worth trying to parse display variable first in a case we are running under cygwin or mingw.

also please rewrite commit message

```
clx: try to parse DISPLAY environment variable on all platforms

Change function helpfully-automagic-clx-server-path to try
parsing DISPLAY envioronment variable disregarding of the system.
Previously Windows was excluded from that this, but solutions like
mingw or cygwin may export DISPLAY. 

Fix a typo where Windows always picked :clx instead of a request port-type.
```

pouar: Like this?

dkochmanski:
Also please change the restart string, it is not only "unix" restart anymore.

pouar:
this good?

dkochmanski:
lgtm, thanks!

PR #985, opened 2020-03-27
clx-medium: inherit from font-rendering-medium-mixin

### LINKED ISSUE #984: 76dd65c27756d seems to have broken the non-XRender based render
Although now that the XRender based one now works in Windows, this is less of an issue, but still.

dkochmanski:
clx-medium implements font renderer protocols and that's where it gets
text-style* methods from. 76dd65c27756d introduced a regression which
may these methods not applicable. Closes #984.

### DIFF
--- a/Backends/CLX/medium.lisp
+++ b/Backends/CLX/medium.lisp
@@ -47,7 +47,9 @@
 
 ;;; CLX-MEDIUM class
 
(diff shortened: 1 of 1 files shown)


pouar:
This seems to fix it, I was going to suggest this, but I wasn't familiar enough with the codebase to know whether it was a fix or a workaround.

dkochmanski:
superseded by #986

PR #989, opened 2020-03-31
Fix 987/988

nsrahmad:


Core/clim-core/graph-formatting.lisp
```diff
@@ -279,7 +279,8 @@
 					   (with-output-to-output-record
 					       (stream 'standard-graph-node-output-record new-node
 						       :object child)
-					     (funcall object-printer child stream))))
+                                             (with-end-of-line-action (stream :allow)
+                                               (funcall object-printer child stream)))))
```
dkochmanski: is it possible for the object to exceed the page too? in that case we shoudl wrap it also in with-end-of-page-action (stram :allow), no?

nsrahmad: Yes it is possible, But the default end-of-page-action (:scroll) works fine. I think there is no need to change that.

dkochmanski: right you are, thanks

dkochmanski:
merging, thank you

PR #991, opened 2020-04-01
fix-various: backends, text styles, line styles and designs

dkochmanski:
- this method is invalid because it does apply the transformation
  twice thanks to the transform-coordinatex-mixin inherited being
  mixed into the basic-medium

- a method for this generic function already exists and it is
  specialized on transform-coordinates-mixin. It does the right thing,
  that is first inverts the transformation and then draws each
  rectangle


Backends/CLX/medium.lisp
```diff
@@ -124,19 +140,36 @@
...
-                (xlib:gcontext-dashes gc)
-                (case (line-style-dashes line-style)
-                  ((t nil) (round (* (line-style-scale line-style medium) 3)))
-                  (otherwise (line-style-effective-dashes line-style medium)))))))))
+          (let ((effective (line-style-effective-dashes line-style medium)))
+            (if-let ((dashes (line-style-dashes line-style)))
+              (setf (xlib:gcontext-line-style gc) :dash
+                    (xlib:gcontext-dashes gc) effective)
+              (setf (xlib:gcontext-line-style gc) :solid
+                    (xlib:gcontext-dashes gc) effective))))))))
```
scymtym, line 155: I suggest returning two values from `line-style-effective-dashes`: 1) `:dash` or `:solid` 2) scaled dash lengths.

Backends/CLX/medium.lisp
```diff
@@ -124,19 +140,36 @@
...
+
+
+(defmethod (setf medium-transformation) :around (transformation (medium clx-medium))
+  (let ((old-tr     (medium-transformation medium))
+        (line-style (medium-line-style medium))
+        (new-tr     (call-next-method)))
+    (with-slots (gc) medium
+      (when (and gc
+                 (eq :coordinate (line-style-unit line-style))
```
scymtym, line 164: If `:coordinate` is not the default, maybe do this check first and skip everything else if possible.

Backends/CLX/medium.lisp
```diff
@@ -124,19 +140,36 @@
...
+                 (not (transformation-equal old-tr new-tr)))
+        (let ((thickness    (line-style-effective-thickness line-style medium))
+              (dash-pattern (line-style-effective-dashes    line-style medium))
+              (dashesp      (if (line-style-dashes line-style) :dash :solid)))
```
scymtym, line 168: This would also benefit from `line-style-effective-dashes` returning two values.

Backends/CLX/medium.lisp
```diff
@@ -577,49 +610,47 @@ translated, so they begin at different position than [0,0])."))
...
-                    (min-x (round-coordinate (- x radius)))
-                    (min-y (round-coordinate (- y radius)))
-                    (max-x (round-coordinate (+ x radius)))
-                    (max-y (round-coordinate (+ y radius))))
-               (when (and (typep min-x '(signed-byte 16))
-                          (typep min-y '(signed-byte 16)))
-                   (xlib:draw-arc mirror gc min-x min-y
-                                  (- max-x min-x) (- max-y min-y)
-                                  0 (* 2 pi) t))))))))
+      (let ((radius (/ (line-style-effective-thickness line-style medium) 2)))
```
scymtym, line 613: This seems to round+clamp, then divide by 2, then round again, whereas the previous code only rounded in the end. Is this intentional?

Backends/CLX/medium.lisp
```diff
@@ -577,49 +610,47 @@ translated, so they begin at different position than [0,0])."))
...
-                 (let ((min-x (round-coordinate (- x radius)))
-                       (min-y (round-coordinate (- y radius)))
-                       (max-x (round-coordinate (+ x radius)))
-                       (max-y (round-coordinate (+ y radius))))
-                   (when (and (typep min-x '(signed-byte 16))
-                              (typep min-y '(signed-byte 16)))
-                     (xlib:draw-arc mirror gc min-x min-y
-                                    (- max-x min-x) (- max-y min-y)
-                                    0 (* 2 pi) t))))))))))
+      (let ((radius (/ (line-style-effective-thickness line-style medium) 2)))
```
scymtym, line 636: See above.

Backends/CLX/medium.lisp
```diff
@@ -645,15 +676,6 @@ translated, so they begin at different position than [0,0])."))
...
                                            (min #x7FFF (max #x-8000 (round-coordinate y2))))))))))))))))
 
-(defmethod medium-draw-lines* ((medium clx-medium) coord-seq)
-  ;; Invert the transformation and apply it here, as the :around
-  ;; methods on transform-coordinates-mixin will cause it to be
-  ;; applied twice, and we need to undo one of those.
-  (let ((tr (invert-transformation (medium-transformation medium))))
-    (with-transformed-positions (tr coord-seq)
-      (do-sequence ((x1 y1 x2 y2) coord-seq)
-        (medium-draw-line* medium x1 y1 x2 y2)))))
```
scymtym, line 655: I suggest doing this in a separate commit.

Core/clim-basic/drawing/design.lisp
```diff
@@ -360,6 +364,12 @@
     (color-rgb (indirect-ink-ink +foreground-ink+))
     (opacity-value color)))
 
+;;; Regions
+(defmethod design-ink ((design region) x y)
+  (if (region-contains-position-p design x y)
+      (design-ink +foreground-ink+ x y)
+      +transparent-ink+))
```
scymtym, line 371: Is this mandated by the spec? If so, maybe cite the relevant section. I was only aware of `draw-design` for this kind of thing.

Core/clim-basic/drawing/design.lisp
```diff
@@ -300,6 +300,10 @@
         (symbol-value dynamic-variable-symbol)
         default-ink)))
 
+(defmethod color-rgba ((ink indirect-ink))
+  (let ((ink (indirect-ink-ink ink)))
+    (color-rgba ink)))
```
scymtym, line 305: This is pretty testable, so if you don't mind …

Core/clim-basic/drawing/medium.lisp
```diff
@@ -459,7 +462,7 @@
 
 (defgeneric (setf graphics-state) (new-gs gs)
   (:method ((new-gs graphics-state) (gs graphics-state))
-    #+(or) "This is a no-op, so :after methods have primary method")
+    #+(or) "This is a no-op, so :after methods have the primary method.")
```
scymtym, line 465: Maybe "This is a no-op, but :after methods don't work without a primary method"?

Core/clim-basic/drawing/medium.lisp
```diff
@@ -96,7 +96,8 @@
       ((:sans-serif) 3)))
 
   (defun face-key (face)
-    (if (equal face '(:bold :italic))
+    (if (or (equal face '(:bold :italic))
+            (equal face '(:italic :bold)))
```
scymtym, line 100: I suggest checking the text style parameters and constructing the key at the same time (and adding a test): https://github.com/scymtym/McCLIM/commit/35398b64119088dce75715a019e1385b179e0c37

Backends/CLX/medium.lisp
dkochmanski: that would be fine but I want to factor out these functions (line-style-effective-dashes and line-style-effective-thickness), to be shared between backends, and these two values are very clx-specific.

Backends/CLX/medium.lisp
dkochmanski: we need to call the next method anyway, so the first let is justified, I've made it easier for the eye with

```
(defmethod (setf medium-transformation) :around (transformation (medium clx-medium))
  (let ((old-tr     (medium-transformation medium))
        (line-style (medium-line-style medium))
        (new-tr     (call-next-method)))
    (unless (and (eq :coordinate (line-style-unit line-style))
                 (not (transformation-equal old-tr new-tr)))
      (when-let ((gc (slot-value medium 'gc)))
        (let ((thickness    (line-style-effective-thickness line-style medium))
              (dash-pattern (line-style-effective-dashes    line-style medium))
              (dashesp      (if (line-style-dashes line-style) :dash :solid)))
          (setf (xlib:gcontext-line-style gc) dashesp
                (xlib:gcontext-dashes gc)     dash-pattern
                (xlib:gcontext-line-width gc) thickness))))
    new-tr))
```

Backends/CLX/medium.lisp
```diff
@@ -77,13 +77,24 @@
...
+    (prog1 :round
+      (warn "Unknown cap style ~S, using :round." clim-shape))))
+
+(defun translate-join-shape (clim-shape)
+  (if-let ((clx-shape (cdr (assoc clim-shape +join-shape-map+))))
+    clx-shape
+    (prog1 :miter
+      (if (eq clim-shape :none)
+          (warn "Unsupported join style :NONE, using :MITER.")
+          (warn "Unknown join style ~s, using :MITER." clim-shape)))))
```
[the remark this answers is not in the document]
dkochmanski: I agree that it is clearer, I've followed existing convention for translate-cap-shape (changing both).

Backends/CLX/medium.lisp
```diff
@@ -77,13 +77,24 @@
                                (:round . :round)
                                (:no-end-point . :not-last)))
 
+(defconstant +join-shape-map+ '((:miter . :miter)
```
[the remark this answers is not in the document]
dkochmanski: fixed

Backends/CLX/medium.lisp
dkochmanski: ditto

Backends/CLX/medium.lisp
dkochmanski: it is intentional. clamp + round in line-style-effective-thickness (and dashes) is for all lines - clx signalled errors when the value was not a) rounded to integer, b) integer fitting in a byte (protocol thing). This radius is for points, from the center of the point we draw it with such radius (which also needs to be an integer), since value is already clamped there is no need to clamp it further. For instance radius for the effective thickness 33 would be 33/2, which is not an integer.

Backends/CLX/medium.lisp
dkochmanski: ditto

Backends/CLX/medium.lisp
dkochmanski: fixed

Core/clim-basic/drawing/medium.lisp
dkochmanski: fixed

Core/clim-basic/drawing/medium.lisp
dkochmanski: I've cherry-picked your change and split it in two (change to internals and a test).

Core/clim-basic/drawing/design.lisp
dkochmanski: it is, but it is quite obvious, Designs chapter has a whole section "http://bauhh.dyndns.org:8000/clim-spec/14-4.html#_745", I don't think that adding a comment is necessary.

http://bauhh.dyndns.org:8000/clim-spec/14-4.html#_745

dkochmanski:
I believe I've addressed all remarks (either by answering them or by fixing the issue).
Moreover I've cherry-picked a commit from another branch for medium-clear-area.

Core/clim-basic/drawing/design.lisp
dkochmanski: fixed

Backends/CLX/medium.lisp
scymtym: The function could return a Boolean as the second return value to indicate dashed vs. solid. That could be useful for other backends as well.

Also what is the meaning of `xlib:gcontext-dashes` when `xlib:gcontext-line-style` is `:solid`?

dkochmanski: as of their meaning – I don't know. I've decided to not change the behavior I do not understand (it was carried from the old code).

dkochmanski: I would skip the single-value behavior now, because we have similar functions in other parts of the codebase and it will be easier to change them at one go if they are consistent.

Core/clim-basic/drawing/design.lisp
scymtym: You are right. It is clear in the specification.

Backends/CLX/medium.lisp
scymtym: Rounding and clamping only once at the end seems to achieve those things. I'm not sure I understand.

dkochmanski: me neither, could you show me how do you imagine the code the way you want it to look?

scymtym: For `-draw-point*`, I guess it would be
```lisp
(let ((unrounded-unclamped-circumference (backend-independent-line-style-effective-thickness line-style medium)))
  (if (< unrounded-unclamped-circumference 2)
    ;; draw points without computing radius
    (let ((radius (/ unrounded-unclamped-circumference 2))) ; maybe also round here
      ;; draw arcs using radius
      )))
```

Note that there would be no clamping when drawing points. I assume clamping to `[1,255]` is intended for the line drawing functions, but the point drawing functions use `xlib:draw-arc` which is probably not limited to radii below 255.

The other difference is no rounding and division if the radius is below 1.

I'm not sure whether rounding is needed where it says "maybe also round here". That's a matter of intended behavior. For that matter, the points could be drawn with
```lisp
(xlib:draw-arc … (round-coordinate (- x radius)) (round-coordinate (- x radius)) rounded-circumference rounded-circumference …)
```
but that would result in yet another behavior.

dkochmanski:
clamping to [1,255] is wrong. I apparently had some huge transformation in the testing code which gone beyond 2^16 for the :coordinate unit. Fixed accordingly. Apparently I've overwritten the author change for the text style commits and I still have no idea how to change that.

dkochmanski:
pushed with the right commit author for text-style changes.

scymtym:
Commit messages:
* Remove final period in "clx: medium: remove the method MEDIUM-DRAW-LINES*."
* I don't understand the body of the "graphics-state-line-style-border: specialize on medium" commit messages, mostly because I'm not sure what the "stub method" is in this context

dkochmanski:
thanks, I've squashed your commit and improved commit messages in question.

dkochmanski:
I've addressed remaining remarks. I've named the appropriate variable `diameter` not `circumference`. effective-dash-pattern returns two values.

Backends/CLX/medium.lisp
```diff
@@ -99,13 +109,18 @@
...
 
 (defun line-style-effective-dashes (line-style medium)
-  (let ((scale (line-style-scale line-style medium)))
-    (map 'list #'(lambda (dash) (round (* dash scale)))
-         (line-style-dashes line-style))))
+  (when-let ((dashes (line-style-dashes line-style)))
+    (let ((scale (line-style-scale line-style medium)))
+      (flet ((scale-dash (dash)
+               (* dash scale)))
+        (values (if (member dashes '(t nil))
```
scymtym, line 120: This could be `(eq dashes t)` since `(eq dashes nil)` is impossible in the `when-let` body.

dkochmanski: since there is no need to set dashes when it is solid, I've changed it once again to return one value:

```(defun line-style-effective-dashes (line-style medium)
  (when-let ((dashes (line-style-dashes line-style)))
    (let ((scale (line-style-scale line-style medium)))
      (if (eq dashes t)
          (* scale 3)
          (map 'list (lambda (d) (* scale d)) dashes)))))
```

and use it like this

```
(if-let ((dash-pattern (line-style-effective-dashes line-style medium)))
            (setf (xlib:gcontext-line-style gc) :dash
                  (xlib:gcontext-dashes gc) (if (atom dash-pattern)
                                                (round dash-pattern)
                                                (mapcar #'round dash-pattern)))
            (setf (xlib:gcontext-line-style gc) :solid))
```

scymtym: Right, returning two values was only necessary because we weren't sure whether `xlib:gcontext-dashes` has to be called in the `:solid` case as well.

PR #997, opened 2020-04-05
Menu buttons created from menu items pick up text-style

scymtym:
As described in the specification, `define-command-table` already allowed specifying a text style and the specified text style was already stored in the menu item. However, before this change, `make-menu-button-from-menu-item` did not pass it on when creating menu buttons.


dkochmanski:
lgtm, thanks!

PR #999, opened 2020-04-12
Fix CLX SYNTHESIZE-POINTER-MOTION-EVENT, use WITH-CURRENT-SOURCE-FORM in presentation macros

scymtym:


dkochmanski:
lgtm, thanks!

PR #1001, opened 2020-05-04
Improve setf star

dkochmanski:


Core/clim-basic/setf-star.lisp
```diff
@@ -33,37 +33,51 @@
...
+    (when aux
+      (error "The use of &aux is not allowed in generic function lambda list."))
+    ;; See https://gitlab.com/embeddable-common-lisp/ecl/-/issues/583.
+    ;; Basically ECL and SBCL punts on defsetf keyword arguments which
+    ;; are not interned in a keyword package.
+    (loop for (key &rest rest) in keys
+          unless (keywordp (car key))
+            do (let ((new-key (alexandria:make-keyword (car key))))
+                 (warn "Coercing ~s to keyword ~s." (car key) new-key)
+                 (setf (car key) new-key)))
```
scymtym, line 54: I think we should not do anything here and just assume the implementations will get fixed.

Core/clim-basic/setf-star.lisp
```diff
@@ -33,37 +33,51 @@
...
-init-form supplied-p-parameter)."
-  (mapcar (lambda (a) `(,a nil ,(gensym))) args))
-
 (defmacro defgeneric* (fun-name lambda-list &body options)
   "Defines a SETF* generic function.  FUN-NAME is a SETF function
 name.  The last required argument is the single argument to the function in a
 SETF place form; the other arguments are values collected from the
 SETF new value form."
   (unless (setf-name-p fun-name)
     (error "~S is not a valid name for a SETF* generic function." fun-name))
```
scymtym, line 42: Would be nice to use `with-current-source-form` for all syntax errors.

Core/clim-basic/setf-star.lisp
```diff
@@ -33,37 +33,51 @@
...
+          (args (butlast required))
+          (place (alexandria:lastcar required))
+          (optional (mapcar (lambda (opt)
+                              (list (first opt) nil (gensym)))
+                            optional))
+          (key-args (mapcar (lambda (key)
+                              (destructuring-bind ((key arg) val suppliedp) key
+                                (declare (ignore val suppliedp))
+                                `((,key ,arg) nil ,(gensym))))
+                            keys)))
```
scymtym, line 66: I would use quasiquote for both or none.

Core/clim-basic/setf-star.lisp
```diff
@@ -33,37 +33,51 @@
...
+    ;; Basically ECL and SBCL punts on defsetf keyword arguments which
+    ;; are not interned in a keyword package.
+    (loop for (key &rest rest) in keys
+          unless (keywordp (car key))
+            do (let ((new-key (alexandria:make-keyword (car key))))
+                 (warn "Coercing ~s to keyword ~s." (car key) new-key)
+                 (setf (car key) new-key)))
+    (let ((setf-name (cadr fun-name))
+          (gf (make-setf*-gfn-name fun-name))
+          (args (butlast required))
```
scymtym, line 57: Why split this off? Couldn't line 79 just use `,,@required`?

Tests/setf-star.lisp
```diff
@@ -76,3 +76,27 @@ RETURN-POLAR is true then return the polar coordinates."
...
+                                             ((:duf one))
+                                             ((:qux two)))
+  (setf (my-point-x point) arg)
+  (setf (my-point-y point) (+ one two kaboom1 kaboom2)))
+
+(test my-qux-dufbar-test
+ (let ((point (make-instance 'my-point :x 0 :y 0)))
+   (setf (my-qux-dufbar point 1 2 :duf 3 :qux 4) 5)
+   (is (and (= 5 (my-point-x point))
+            (= 10 (my-point-y point))))))
```
scymtym, line 102: Seems to need one additional column of indentation.

Core/clim-basic/setf-star.lisp
dkochmanski: We need to compute args and place anyway for the defsetf lambda list, so we are not computing anything unnecessarily. having both placed separately signifies, that the last required argument is the place. It is only for the convenience of a person who reads the macro.

Core/clim-basic/setf-star.lisp
dkochmanski: you are right, fixed.

Core/clim-basic/setf-star.lisp
dkochmanski: fixed

Tests/setf-star.lisp
dkochmanski: fixed

Core/clim-basic/setf-star.lisp
dkochmanski: for the sake of a kawaii world, I've wrapped both errors in the macro. fixed

dkochmanski:
thank you for the review, I've addressed all remarks (adjusted in all cases except the one where we split args and the place, I've argued that it  is more readable that way).

Core/clim-basic/setf-star.lisp
dkochmanski: I had to withdraw this change -- utils are loaded /after/ setf-star file, so the macro is not recognized.

scymtym: We could at some point switch to `trivial-with-current-source-form` which is now in quicklisp. But that can also be done after this pull request (if at all).

Core/clim-basic/setf-star.lisp
```diff
@@ -33,37 +33,42 @@
...
-                ,@(and keyp `(&key ,@keys))
+                ,@(and keyp `(&key ,@key-args))
                 ,@(and allow-other-keys `(&allow-other-keys)))
-               ,(butlast required)
-             `(funcall #',',gf ,,@required
-                ,,@(mapcar (lambda (x) `(and ,(third x) ,(car x))) optional)
-                ,@,@(mapcar (lambda (x) `(and ,(third x) `(,,(make-keyword (car x)) ,,(car x)))) keys)))
+               ,args
+             `(apply #',',gf ,,@args ,,place
+                (append (list ,,@call-opt) ,,@call-key)))
```
scymtym, line 71: It would be nice to avoid the `apply`, `append`, `list` at runtime if `call-opt` and/or `call-key` are empty.

dkochmanski: see a new version of the code. it avoid apply and avoids appending at runtime (and list).

scymtym:
I think `setf-name-p` should be changed like this:

```lisp
(defun setf-name-p (name)
  (typep name '(cons (eql setf) (cons symbol null))))
```

scymtym:
Just noticed that the generated `(,NAME nil ,(gensym))` lambda list entries for optional parameters now result in warnings since the "supplied-p" variables are not used.

This seems to work (I also realized that we don't have to loop over the keyword parameters twice):
```lisp
(defmacro defgeneric* (fun-name lambda-list &body options)
  "Defines a SETF* generic function.  FUN-NAME is a SETF function
name.  The last required argument is the single argument to the function in a
SETF place form; the other arguments are values collected from the
SETF new value form."
  (unless (setf-name-p fun-name)
    (error "~S is not a valid name for a SETF* generic function." fun-name))
  (multiple-value-bind (required optional rest keys allow-other-keys aux keyp)
      (alexandria:parse-ordinary-lambda-list lambda-list)
    (when aux
      (error "The use of &aux is not allowed in generic function lambda list."))
    (let ((setf-name (second fun-name))
          (gf (make-setf*-gfn-name fun-name))
          (args (butlast required))
          (place (alexandria:lastcar required))
          (optional (mapcar #'first optional))
          (fun-keys (gensym "FUN-KEYS")))
      (multiple-value-bind (key-args call-key)
          (loop for ((key arg)) in keys
                for suppliedp = (gensym)
                collect `((,key ,arg) nil ,suppliedp) into key-args
                collect `(and ,suppliedp `(,,key ,,arg)) into call-key
                finally (return (values key-args call-key)))
        `(progn
           (defsetf ,setf-name
               (,place
                ,@(and optional `(&optional ,@optional))
                ,@(and rest `(&rest ,rest))
                ,@(and keyp `(&key ,@key-args))
                ,@(and allow-other-keys `(&allow-other-keys)))
               ,args
             (let ((,fun-keys (append ,@call-key)))
               `(funcall (function ,',gf) ,,@args ,,place ,,@optional ,@,fun-keys)))
           (defgeneric ,gf ,lambda-list ,@options))))))
```

dkochmanski:
this version is not right, we need to expand optional arguments like fun-keys (that is, if they are not supplied then they should not be present). Method may depend on the fact whether the argument was supplied or not. Fix for that, and improved setf-name-p are pushed

dkochmanski:
final version pushed. it does a similar thing for optional parameters.

dkochmanski:
thanks, I've added your changes

scymtym:
Looks good to me,

PR #1004, opened 2020-05-16
tests: fix the package conflict between clim-tests and clim

dkochmanski:
Recently exported symbol clim:test caused a symbol name conflict with
fiveam:test symbol because clim-tests USE both packages. Use
shadowing-import to rectify that.

### DIFF
--- a/Tests/package.lisp
+++ b/Tests/package.lisp
@@ -1,14 +1,8 @@
 (cl:defpackage #:clim-tests
-  (:use
-   #:clim-lisp
-   #:clim
-   #:clime
-   #:fiveam)
-  (:import-from
-   #:climi
-   #:coordinate=)
-  (:export
-   #:run-tests))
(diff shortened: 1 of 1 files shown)


Tests/package.lisp
```diff
@@ -1,14 +1,8 @@
...
-   #:clim
-   #:clime
-   #:fiveam)
-  (:import-from
-   #:climi
-   #:coordinate=)
-  (:export
-   #:run-tests))
+  (:use #:clim-lisp #:clim #:clime #:fiveam)
+  (:shadowing-import-from #:fiveam #:test)
```
scymtym, line 3: How about selectively importing from `fiveam` instead? We only use a few symbols and issues like this would be less likely in the future.

scymtym: Discussed in IRC.

PR #1007, opened 2020-05-20
A few improvements for Clouseau

scymtym:
Also fixes the problem of pointer documentation for drag and drop translators sometimes not being displayed.

### DIFF
--- a/Core/clim-core/presentation-drag-and-drop.lisp
+++ b/Core/clim-core/presentation-drag-and-drop.lisp
@@ -249,4 +249,5 @@
                    :window window
                    :x x
                    :y y
-                   :stream stream))))))
+                   :stream stream)))
+      (force-output stream))))
--- a/Apps/Clouseau/src/commands.lisp
+++ b/Apps/Clouseau/src/commands.lisp
@@ -1,4 +1,4 @@
-;;;; Copyright (C) 2018, 2019 Jan Moringen
+;;;; Copyright (C) 2018, 2019, 2020 Jan Moringen
 ;;;;
 ;;;; This library is free software; you can redistribute it and/or
 ;;;; modify it under the terms of the GNU Library General Public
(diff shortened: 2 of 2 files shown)


Core/clim-core/presentation-drag-and-drop.lisp
```diff
@@ -249,4 +249,5 @@
                    :window window
                    :x x
                    :y y
-                   :stream stream))))))
+                   :stream stream)))
+      (force-output stream))))
```
dkochmanski, line 253: shouldn't that be finish-output? are we OK with losing data if it can't be flushed immedietely?

scymtym: I'm not sure what losing data means in this context. Could you elaborate?

dkochmanski: `force-output` is specified as follows
```
 force-output initiates the emptying of any internal buffers but does not wait for completion or acknowledgment to return. 
```

if it is not possible to empty internal buffers immediately, then data may not be sent. semantically to ensure that output reached its destination you should use finish-output, which is specified as follows
```
 finish-output attempts to ensure that any buffered output sent to output-stream has reached its destination, and then returns. 
```

Or did I miss the purpose of this call entirely?

scymtym: The change is intended to fix the problem that pointer documentation is sometimes not displayed.

I think the problem is that the CLIM stream buffers until the line is complete or the output is explicitly forced/finished.

Regarding the difference between force and finish: I think force is appropriate because I want make sure the string is displayed at some point, but I don't want block until it is displayed on the screen (which implies an X protocol roundtrip for the `clx` backend) .

dkochmanski: makes sense, thanks.

PR #1008, opened 2020-05-21
In file Backends/PostScript/font/font.lisp

heshrobe:
Method climb:text-bounding-rectangle*

Ensure that the bounding box's y2 is at least as large as y1.  This method calls psfont-text-extents to get the size of the string
but that function does not returns arguments that guaranetee that y2 > y1.  Without this patch, the bounding box of a string
is incorrect.  Formatting functions such as format-graph-from-roots use the bounding box to figure out how to move sub-records
and if the bounding-box is inconsistent with the actual string the output will be incorrect.  Here is a test function to illustrate the issue.
without this patch, the graph is misaligned and the borders don't outline the text:
```lisp
(defun test-ps-1 (&optional (pathname #p"~/foo-1.ps"))
  (with-open-file (file pathname :direction :output :if-exists :supersede :if-does-not-exist :create)
    (clim:with-output-to-postscript-stream (stream file
                                                   :device-type :eps)
      (terpri stream)
      (let ((example '(("Agg" b c) (b d e) (c f g) (d) (e) (f) (g))))
        (flet ((children (node)
                 (loop for symbol in (cdr node)
                       for child = (assoc symbol example)
                       when child
                         collect child)))
          (clim:surrounding-output-with-border (stream :shape :rectangle)
          (format-graph-from-root
           (first example)
           #'(lambda (node stream)
               (clim:surrounding-output-with-border (stream :shape :rectangle)
                 (clim:draw-text* stream (string (first node))
                                  0 0
                                  )))
           #'Children
           :Stream stream
           :orientation :vertical
           )))))))
```

### DIFF
--- a/Backends/PostScript/font/font.lisp
+++ b/Backends/PostScript/font/font.lisp
@@ -163,7 +163,7 @@
                         (error "Unknown font ~S." font-name)))
          (size (font-name-size font-name)))
     (* (/ size 1000) (font-info-ascent font-info))))
-	 
+
 
 (defmethod text-style-descent (text-style (medium postscript-font-medium))
   (let* ((font-name (text-style-mapping (port medium)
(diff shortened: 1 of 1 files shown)


dkochmanski:
thanks! merged

PR #1010, opened 2020-05-24
Fix the defaulting of values for generation-separation and within-generation-separation

heshrobe:
In function format-graph-from-roots:

The values of the options generation-separation and within-generation-separation are supplied in the definition of the class standard-graph-output-record.  The supplied values are appropriate for a horizontal graph but should be reversed for a vertical graph to avoid leaving too much space within generations.

The fix is to provide the default values for these two options in the definition of format-graph from-roots and to swap the values if the orientation is vertical.

However, the values are swapped only if no defaults were provided to avoid over riding specific intent provided by the caller.


dkochmanski:
superseded by #1063 (defaults are defined in initialize-instnace after method for the graph).

PR #1011, opened 2020-06-01
Move FRAME-EXIT check into RUN-FRAME-TOP-LEVEL

goose121:
The comment above the function said it would be a better way of doing things, and I agree. The only thing I'm uncertain about is whether my implementation of the check is as clean as it could be; the named block within the loop that exists only so the `frame-layout-changed` handler can return from it to continue to the next iteration.


dkochmanski:
I agree, this is much cleaner. lgtm, thanks!

PR #1013, opened 2020-06-12
Refresh FRAME-QUERY-IO and INTERACTORP variables after commands

goose121:
Without refreshing these variables, if a command updates the frame's query I/O, the fresh line on the query I/O would be output to the wrong stream (possibly leading to the next iteration of the command loop not starting on its own line in the new stream) and the ABORT restart would print its message to the frame's old query I/O stream, which could be surprising to a client who set the query I/O stream and subsequently invoked the ABORT restart.

### DIFF
--- a/Core/clim-core/frames.lisp
+++ b/Core/clim-core/frames.lisp
@@ -535,7 +535,14 @@ documentation produced by presentations.")
                          (write-string prompt frame-query-io)
                          (funcall prompt frame-query-io frame))
                      (force-output frame-query-io))))
-               (execute-command)
+               (unwind-protect (execute-command)
+                 ;; The command may have set the frame's query I/O, in
+                 ;; which case the values of FRAME-QUERY-IO and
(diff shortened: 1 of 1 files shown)


dkochmanski:
I wonder what is more surprising here -- command is started in one environment, so it seems to be a sane strategy to use the same environment when printing abort (and a fresh line), no? and rebind these to the new values when the /next/ command is read.

goose121:
It seems to me that it'd be the expectation, at least in the case of the abort restart, that it'd be printed to "the frame's query I/O", which imo would mean the frame's current query I/O? It seems surprising to me for something to use an old value of an object's slot for much of anything, I would think; the `fresh-line` I'm not as sure about, but I think the abort should definitely use the frame's query I/O.

dkochmanski:
you are right, sorry for delaying this for this long; I've cherry-picked the commit to master.

PR #1014, opened 2020-06-15
A fix and a few optimizations for the CLX backend

scymtym:
* Clamp individual line dash lengths to `[0,255]`

* Use `clamp` instead of `min` and `max` since it is more readable and saves one comparison

* In `medium-draw-rectangles*`, allocated the coordinate vector upfront instead of growing it incrementally


dkochmanski:
lgtm, thanks!

PR #1015, opened 2020-06-15
Fix bounding region for degenerate ellipses

scymtym:
Attempting to compute the bounding rectangle for certain degenerate ellipses (non-invertible but non-zero transform) resulted in an error while attempting to invert the transform. Since these ellipses are basically lines, compute the bounding rectangle by computing the end points of the line.

### DIFF
--- a/Core/clim-basic/geometry/region-bounding-rectangles.lisp
+++ b/Core/clim-basic/geometry/region-bounding-rectangles.lisp
@@ -47,12 +47,45 @@
       (values (- cx x) (- cy y) (+ cx x) (+ cy y)))))
 
 (defmethod bounding-rectangle* ((region elliptical-thing))
-  (with-slots (tr start-angle end-angle) region
-    (multiple-value-bind (cx cy) (ellipse-center-point* region)
-      (if (every #'zerop (multiple-value-list (ellipse-radii region)))
-          (values cx cy cx cy)
-          (ellipse-bounding-rectangle region)))))
-
+  (let ((transform (polar->screen region))
+        dx dy)
+    (cond ;; If TRANSFORM is invertible, REGION is not degenerate
+          ;; (i.e. a line or a point).
+          ((invertible-transformation-p transform)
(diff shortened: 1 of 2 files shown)


Core/clim-basic/geometry/region-bounding-rectangles.lisp
```diff
@@ -47,12 +47,45 @@
...
+                   (dx (abs dx))
+                   (dy (abs dy)))
+               (multiple-value-bind (dx- dy-)
+                   (if (and (or (null start-angle) (<= start-angle 0))
+                            (or (null end-angle) (<= 0 end-angle)))
+                       (values dx dy)
+                       (values 0 0))
+                 (multiple-value-bind (dx+ dy+)
+                     (if (and (or (null start-angle) (<= start-angle (/ pi 2)))
+                              (or (null end-angle) (<= (/ pi 2) end-angle)))
```
dkochmanski, line 80: is this condition right? imagine that the start angle is 3pi/2 and he end angle is pi, then pi/2 is contained in the angle. We have the utility to abstract this nuance away: `arc-contains-angle-p` (in region-utilities.lisp).

also, from the fact that the angle 0 is in the arc,  you can't infer that dy is part of the arc (and vice versa, from the fact that pi/2 is part of the arc, you can't infer that dx is part of the arc), so either I'm not reading it right or it you should check for all four angles, i.e like this (did not try to compile it):

```common-lisp
(let* ((sa (or start-angle 0))
       (ea (or end-angle (* 2 pi)))
       (h+ (if (arc-contains-angle-p sa ea (* 0/2 pi)) dx 0))
       ;; left-handedness
       (v- (if (arc-contains-angle-p sa ea (* 1/2 pi)) dy 0))
       (h- (if (arc-contains-angle-p sa ea (* 2/2 pi)) dx 0))
       (v+ (if (arc-contains-angle-p sa ea (* 3/2 pi)) dy 0)))
  (values (- cx h-) (- cy v-) (+ cx h+) (+ cy v+)))
```

PR #1017, opened 2020-06-24
Fix presentation translators

dkochmanski:


Core/clim-core/presentations.lisp
```diff
@@ -501,6 +500,20 @@ filled in."
 (defmethod get-ptype-metaclass ((type class))
   type)
 
+;;; The presentation type specifier may be in one of these forms:
+;;;
+;;;     name
```
scymtym, line 509: Maybe move this comment to the generic function since the "name" case is handled by a different method?

Core/clim-core/presentation-translators.lisp
```diff
@@ -78,25 +78,19 @@ and used to ensure that presentation-translators-caches are up to date.")
                             :initform (make-hash-table :test #'eq)
                             :documentation "Holds transators with a simple
   from-type (i.e. doesn't contain \"or\" or \"and\").")
-   (translator-cache-generation :accessor translator-cache-generation
-                                :initform 0)
-   (presentation-translators-cache
-    :writer (setf presentation-translators-cache)
-    :initform (make-hash-table :test #'equal))))
+   (cache-gen :initform 0)
+   (cache :initform (make-hash-table :test #'equal))))
```
scymtym, line 77: I like `presentation-translators-cache` -> `cache` but I would prefer `cache-generation` for the other slot since "gen" can mean "generator" or "generation".

Core/clim-core/presentation-translators.lisp
```diff
@@ -78,25 +78,19 @@ and used to ensure that presentation-translators-caches are up to date.")
...
-      table
-    (unless (or (= generation *current-translator-cache-generation*)
+(defun presentation-translators-cache (table)
+  (check-type table translator-table)
+  (with-slots (cache cache-gen) table
+    (unless (or (= cache-gen *current-translator-cache-generation*)
                 (zerop (hash-table-size cache)))
       (clrhash cache))
-    (setf generation *current-translator-cache-generation*)
+    (setf cache-gen *current-translator-cache-generation*)
```
scymtym, line 88: Maybe use the accessor here instead of deleting it? `with-slots` generates a `slot-value` call for each use of `cache`.

Core/clim-core/presentation-translators.lisp
```diff
@@ -165,11 +165,8 @@ and used to ensure that presentation-translators-caches are up to date.")
          (real-to-type (expand-presentation-type-abbreviation to-type)))
     ;; I did not investigate yet to decide what is the right thing. Let's issue
     ;; a warning for a time being. See #778. -- jd 2019-06-11
-    (cond ((not (ignore-errors (get-ptype-metaclass real-from-type)))
-           (alexandria:simple-style-warning "~s does not have a presentation metaclass. Translator may not work." real-from-type))
-          #+ (or)
-          ((not (ignore-errors (get-ptype-metaclass real-to-type)))
-           (alexandria:simple-style-warning "~s does not have a presentation metaclass. Translator may not work." real-to-type)))
```
scymtym, line 178: It is not duplicated. One branch has `real-from-type`, the other has `real-to-type`. I have no idea whether it should be deleted on a different bases.

Core/clim-core/presentation-translators.lisp
```diff
@@ -105,7 +105,7 @@ and used to ensure that presentation-translators-caches are up to date.")
                         old))
   (invalidate-translator-caches)
   (setf (gethash (name translator) (translators table)) translator)
-  (push translator (gethash (from-type translator)
+  (push translator (gethash (presentation-type-name (from-type translator))
```
scymtym: Can the recently added tests for adding and removing presentation translators be extended to verify this fix?

Tests/presentations.lisp
```diff
@@ -105,4 +105,8 @@
   (expect-t-t '(or integer symbol) '(or symbol integer))
 
   (expect-t-t '(or real complex) 'number)
-  #+nil (expect-t-t 'number '(or real complex)))
+  (expect-t-t 'number '(or real complex)))
+
+(test presentations.type-relations.5
+  (let ((type (expand-presentation-type-abbreviation '(member "a" 3))))
+    (expect-t-t type '(or string integer))))
```
scymtym, line 112: Does fiveam support something like "expected failures"? Otherwise having more failing tests makes test suite less useful.

Core/clim-core/presentation-translators.lisp
```diff
@@ -334,26 +319,26 @@ and used to ensure that presentation-translators-caches are up to date.")
...
-                     if (stupid-subtypep (to-type translator) to-type)
-                       do (vector-push-extend (cons translator table-counter)
-                                              translator-vector))))
-            (map-over-ptype-superclasses #'get-translators from-name)))
+        ;; We need to go over each translator in all tables, because
+        ;; the FROM-TYPE may be a subtype of a translator FROM-TYPE
+        ;; despite not having the same presentation metaclass, like
+        ;; "meta" presentation types OR, AND etc. -- jd 2020-06-24
+        (loop with translators = (translators (presentation-translators table))
+              for tr being the hash-value of translators
```
scymtym, line 337: Is the hash-table still used as a hash-table somewhere? Otherwise changing the representation could make sense.

Core/clim-core/presentation-translators.lisp
```diff
@@ -73,11 +73,6 @@ and used to ensure that presentation-translators-caches are up to date.")
   ((translators :accessor translators :initarg :translators
```
scymtym, line 73: Commit messagae: "This change render" -> "This change renders"

Tests/presentation-translators.lisp
```diff
@@ -0,0 +1,42 @@
...
+        (non-1 (find-presentation-translators 'number 'string 'pt.smoke-ct2))
+        (non-2 (find-presentation-translators '(or string number) 'string 'pt.smoke-ct2))
+        (non-3 (find-presentation-translators '(member "dan" :foo) 'string 'pt.smoke-ct2)))
+    (is (member tr all-1))
+    (is (member tr all-2))
+    (is (member tr all-3))
+    (is (member tr all-4))
+    (is (member tr all-5))
+    (is (member tr all-6))
+    (is (member tr all-7))
```
scymtym, line 39: I'm not sure what "all-6 and all-7 should both pass," in the commit message means. Doesn't this assert that the translator is found?

Core/clim-core/presentations.lisp
```diff
@@ -501,6 +500,20 @@ filled in."
...
+;;;    (name &rest params)
+;;;   ((name &rest params) &rest options)
+(defmethod get-ptype-metaclass ((type cons))
+  (let ((ptype-name (first type)))
+    (if (atom ptype-name)
+        (get-ptype-metaclass ptype-name)
+        (let ((ptype-name (first type)))
```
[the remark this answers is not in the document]
dkochmanski: it should be `(let ((ptype-name (first ptype-name))) …)`, fixed

Core/clim-core/presentations.lisp
dkochmanski: fixed.

Core/clim-core/presentation-translators.lisp
dkochmanski: done (and added accessors)

Core/clim-core/presentation-translators.lisp
dkochmanski: changed

Core/clim-core/presentation-translators.lisp
dkochmanski: removed the commit, nice catch

Core/clim-core/presentation-translators.lisp
dkochmanski: this hash table is removed in a later commit, so this is added only for the change atomicity, I don't think there is a need for testing a removed code

Tests/presentations.lisp
dkochmanski: it doesn't, but it would be nice to have in the fiveam. As of the usefulness, I suppose it depends on what do you expect from tests: to show defects or to show regressions.

Core/clim-core/presentation-translators.lisp
dkochmanski: it is used in many places, add-translator, and remove-translator in this file, plus in commands.lisp (i.e find-presentation-translator)

Core/clim-core/presentation-translators.lisp
dkochmanski: fixed

Tests/presentation-translators.lisp
dkochmanski: in the test `presentation-translators.meta` we have multiple cases. I've added tests which does not pass, all-6 case is
(member "a" "b") -> string, and there exists a translator (or real string) -> , and (member "a" "b") is a subtype of (or real string) -- simpilar in the case all-7 (in both cases the translator should be found). I've changed to commit message to elaborate a bit more.

Core/clim-core/presentation-translators.lisp
scymtym: I see.

Tests/presentations.lisp
scymtym: I use automatic execution the test suite and result analysis for spotting regressions but I agree that indicating defects is valuable too. So yak shaving in fiveam is the solution, then?

Core/clim-core/presentation-translators.lisp
scymtym: OK.

Tests/presentation-translators.lisp
scymtym: I see. My remark about expected failures applies here as well in that case.

PR #1024, opened 2020-07-02
presentation-typep: implement missing methods

dkochmanski:
Presentation types BLANK-AREA and MENU-ITEM had no implementation of
the presentation method PRESENTATION-TYPEP. Default method was
returning mistakenly T and recent fix to that revealed that issue.

### DIFF
--- a/Core/clim-core/menu-choose.lisp
+++ b/Core/clim-core/menu-choose.lisp
@@ -215,7 +215,16 @@
           (when deexpose ; Checkme as well.
             (disown-frame fm frame)))))))
 
+;;; MENU-ITEM is used as a default presentation type for displaying
+;;; objects in the menu. Display object is described in the spec of
+;;; the function FRAME-MANAGER-MENU-CHOOSE. From this description it
+;;; seems that anything may be a display object, so we fix the
+;;; predicate PRESENTATION-TYPEP to always return T. -- jd 2020-07-02
 (define-presentation-type menu-item ())
+
+(define-presentation-method presentation-typep (object (type menu-item))
+  (declare (ignore object type))
+  t)
 
 (defmethod menu-choose
     (items &rest args &key associated-window &allow-other-keys)
--- a/Core/clim-core/presentations/presentation-type-functions.lisp
+++ b/Core/clim-core/presentations/presentation-type-functions.lisp
@@ -82,18 +82,29 @@ otherwise return false."
 (define-presentation-generic-function %presentation-typep presentation-typep
   (type-key parameters object type))
 
+;;; The default behavior is implemented in the function below. If we
+;;; end up here that means that the presentation parameters were
+;;; specified or that the presentation type does not correspond to a
+;;; class. In that case the method PRESENTATION-TYPEP must be
+;;; implemented by the programmer. -- jd 2020-07-02
 (define-default-presentation-method presentation-typep (object type)
-  (declare (ignore object type))
-  nil)
+  (declare (ignore object))
+  (error "The presentation type ~s doesn't implement a ~s method."
+         type 'presentation-typep))
 
 (defun presentation-typep (object type)
-  (with-presentation-type-decoded (name parameters)
-      type
-    (when (null parameters)
-      (let ((clos-class (find-class name nil))) ; Don't error out.
-        (when (and clos-class (typep clos-class 'standard-class))
-          (return-from presentation-typep (typep object name)))))
-    (funcall-presentation-generic-function presentation-typep object type)))
+  (with-presentation-type-decoded (name parameters) type
+    (case name
+      ((t) t)                           ; super type
+      ((nil) nil)                       ; empty type
+      (otherwise
+       (let ((clos-class (find-class name nil))) ; Don't error out.
+         (when (and clos-class (typep clos-class 'standard-class))
+           (unless (typep object name)
+             (return-from presentation-typep nil))
+           (when (null parameters)
+             (return-from presentation-typep t))))
+       (funcall-presentation-generic-function presentation-typep object type)))))
 
 (define-presentation-generic-function
     %presentation-subtypep
--- a/Core/clim-core/presentations/presentation-type-functions.lisp
+++ b/Core/clim-core/presentations/presentation-type-functions.lisp
@@ -82,11 +82,9 @@ otherwise return false."
 (define-presentation-generic-function %presentation-typep presentation-typep
   (type-key parameters object type))
 
-;;; FIXME this method should return NIL. There is a problem with
-;;; MENU-ITEM translators which doesn't work without this kludge.
 (define-default-presentation-method presentation-typep (object type)
   (declare (ignore object type))
-  t)
+  nil)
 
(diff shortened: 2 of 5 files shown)


Core/clim-core/presentations/presentation-type-functions.lisp
```diff
@@ -82,11 +82,9 @@ otherwise return false."
 (define-presentation-generic-function %presentation-typep presentation-typep
   (type-key parameters object type))
 
-;;; FIXME this method should return NIL. There is a problem with
-;;; MENU-ITEM translators which doesn't work without this kludge.
 (define-default-presentation-method presentation-typep (object type)
-  (declare (ignore object type))
-  t)
+  (declare (ignore object))
+  (warn "Presentation type ~s doesn't implement PRESENTATION-TYPEP." type))
```
scymtym, line 87: I don't think warnings work very well for this purpose (see "Drawing outside the mirror" warning). If not having a method is a violation of the protocol, this should be an error.

Core/clim-core/standard-presentations.lisp
```diff
@@ -129,28 +129,29 @@
 (define-presentation-type blank-area ()
   :inherit-from t)
 
+;;; Do other slots of this have to be bound in order for this to be
+;;; useful?  Guess we'll see.
+(defconstant *null-presentation* (make-instance 'standard-presentation
+                                                :object nil
+                                                :type 'blank-area
+                                                :view +textual-view+))
```
scymtym, line 137: Why the change to `defconstant`? If we ever remove the custom `defconstant` and switch to `cl:defconstant` (I think there was a plan to do so at some point), this will lead to problems.

dkochmanski: this variable is specified to be a constant

Core/clim-core/presentations/presentation-type-functions.lisp
dkochmanski: it is not a violation, but it is a likely issue in the user code to call presentation-typep on a presentation without this method

dkochmanski: (at least it was a likely issue in McCLIM code :-)

Core/clim-core/standard-presentations.lisp
scymtym: I think deviating from the spec is best in this case since our `defconstant` does not make `*null-presentation*` a constant, but using `cl:defconstant` will lead to problems (requiring `make-load-form` methods, object identify issues).

It is strange that the name uses earmuffs while for example `+black+` uses `+`.

dkochmanski: I'll change it to defvar then, to avoid at least reinitialization during recompile.

dkochmanski: changed.

Core/clim-core/presentations/presentation-type-functions.lisp
scymtym: I guess the questions are "What do I have to do when I see this warning for one of my presentation types?" and "What happens if I don't do anything?".

dkochmanski: presentation-typep is not called on user-defined presentation types from mcclim core, so you, as a caller of the presentation-typep, know best what you need to do (i.e replace the call with hardcoded nil, or implement the method).

Core/clim-core/presentations/presentation-type-functions.lisp
```diff
@@ -82,18 +82,29 @@ otherwise return false."
...
+;;; end up here that means that the presentation parameters were
+;;; specified or that the presentation type does not correspond to a
+;;; class. In that case the method PRESENTATION-TYPEP must be
+;;; implemented by the programmer. -- jd 2020-07-02
 (define-default-presentation-method presentation-typep (object type)
-  (declare (ignore object type))
-  nil)
+  (declare (ignore object))
+  (error "The presentation type ~s doesn't implement a ~s method."
+         type 'presentationt-typep))
```
scymtym, line 93: Typo.

Core/clim-core/presentations/presentation-type-functions.lisp
```diff
@@ -82,18 +82,29 @@ otherwise return false."
 (define-presentation-generic-function %presentation-typep presentation-typep
```
scymtym, line 82: Should the commit message say `presentation-typep` instead of `presentation-subtypep`? The latter doesn't seem to be changed in any way.

Core/clim-core/presentations/presentation-type-functions.lisp
dkochmanski: fixed

Core/clim-core/presentations/presentation-type-functions.lisp
dkochmanski: yes, typep. fixed

Core/clim-core/standard-presentations.lisp
```diff
@@ -129,28 +129,31 @@
 (define-presentation-type blank-area ()
   :inherit-from t)
 
+;;; Do other slots of this have to be bound in order for this to be
+;;; useful?  Guess we'll see.
+;;;
+;;; This variable is specified to be a constant.
```
nsrahmad, line 135: earmuffs suggests this is a mistake in spec. Lispworks CLIM guide calls it a variable.

----
Screenshot from 2020-07-02 19-48-35

Core/clim-core/menu-choose.lisp
```diff
@@ -215,8 +215,13 @@
           (when deexpose ; Checkme as well.
             (disown-frame fm frame)))))))
 
+;; The MENU-ITEM object is described in the specification of the
+;; function FRAME-MANAGER-MENU-CHOOSE. -- jd 2020-07-02
 (define-presentation-type menu-item ())
 
+(define-presentation-method presentation-typep (object (type menu-item))
+  (presentation-translator-menu-item-p (alexandria:ensure-car object)))
+
```
nsrahmad, line 224: @dkochmanski This method might be needed for items in "menu-choose" menus. Testing with "Accepting-values" test, nothing in the menu-choose menu is highlight-able.

Core/clim-core/standard-presentations.lisp
scymtym: Yeah, we talked about this in the other comment. Should we just change it in our copy of the spec?

dkochmanski: I've added a commit which modifies the spec, thanks

Core/clim-core/menu-choose.lisp
dkochmanski: After reading the spec again it seems that it is not specified anywhere what is the display object. I'm fixing this predicate to always return T. Thanks.

Core/clim-core/presentations/presentation-type-functions.lisp
scymtym: Resolved in IRC discussion.

Issue #452, opened 2018-03-17
Possible bugs in Presentation system


nsrahmad:
I am not sure which of below are bugs.

1) In the below screenshot, will the number `2` be highlighted?
![screenshot_20180317_235156](https://user-images.githubusercontent.com/10996250/37558721-4e8512a4-2a3e-11e8-90ae-a4f0389b56a9.png)
`2` isn't available as integer (Even if `presentation-type-of` says it's integer). 
---
2) Presenting any form as form presents something that's `standard-presentation`.
![screenshot_20180318_002510](https://user-images.githubusercontent.com/10996250/37558966-e15541b8-2a42-11e8-82e6-633488c71bcc.png)

---

3) Ask for `form` and you shall recieve `expression`?
![screenshot_20180318_000303](https://user-images.githubusercontent.com/10996250/37558842-58a3d8d6-2a40-11e8-9477-d050fd62cf9d.png)

deterministic-arts:
You may actually have a point in number 3. 

Point 2 seems to be a misunderstanding of the output. In your example, the first output line

```(PRINT "Hello")```

is the actual output generated by `present`, a presentation, whose presentation object is `(PRINT "Hello")` with presentation type `form`. When you do a subsequent `(accept 'form)`, the line should be active.

The second output line is not generated by `present`, but by the REPL! The `present` function returns the output record (an object of type `standard-presentation` in this case), which gets inserted into the listener's output history.  For the listener, this is simply the value of the form, you asked it to evaluate, and hence it prints it (just as it would print the value of, say, `(+ 1 2)`)

deterministic-arts:
I am not sure about your point 1. Generally speaking, something on the screen is available as a value of (presentation) type `T`, if the output has been generated by any means, which arranges for this, i.e., `present`, `with-output-as-presentation`, etc. Your example does nothing of this sort, so I am not quite sure, why you would expect the `2` to be available with presentation type `integer`. (It is available as `form`, since that's the type specified in the `accept` call).

nsrahmad:
> Point 2 seems to be a misunderstanding of the output. In your example, the first output line

>(PRINT "Hello")

>is the actual output generated by present, a presentation, whose presentation object is (PRINT  "Hello") with presentation type form. When you do a subsequent (accept 'form), the line should be active.

[...]

PR #1026, opened 2020-07-03
Presentation Types #1

dkochmanski:


dkochmanski:
now that we enforce a proper implementation of presentation-typep, I need to go through all our examples and see where we error

dkochmanski:
I've fixed places which incorrectly assumed, that the presentation-typep defaults to T.
Adding the error when object does not fulfill the predicate revealed some issues. A new test file is added.

nsrahmad:
@dkochmanski This will require more time to think through. There are several issues here.

First, You are trying to enforce the type of object matches its presentation type. But that is not how it worked till now.

From mcclim-devel archives, Timothy Moore [writes](https://mailman.common-lisp.net/pipermail/mcclim-devel/2005-August/000653.html):
>I believe that the user (programmer) has the freedom to pair any object 
with any presentation type in a presentation; if that breaks other 
code, then that's his problem. If an application needs a strong 
guarantee that the user (user) enters a valid object for a presentation 
type then the accept method should check that before returning.

Also see the replies by deterministic-arts in https://github.com/McCLIM/McCLIM/issues/452 .

This is of course directly taken from Dynamic Windows [(pdf)](http://bitsavers.trailing-edge.com/pdf/symbolics/software/genera_8/Programming_the_User_Interface.pdf). Here they think this might be an issue:

-----
Screenshot from 2020-07-04 12-47-16
-----

But in a really good [paper](https://dl.acm.org/doi/pdf/10.1145/73660.73678?download=true) about how presentations work, They present it as a delibrate choice:

----
Screenshot from 2020-07-04 12-54-45
----

Second issue is enforcing the implementation of `presentation-typep`. Role of the `presentation-typep` isn't clear. however, from the [spec](http://bauhh.dyndns.org:8000/clim-spec/edit/anno?aid=30B86EBC72FB8351CA3C8EC7233E8573%2F23.3#Z) ((see paragarph above the annotation form):

>The function presentation-typep uses the parameters to check object membership in a type. Adding parameters to a presentation type specifier produces a subtype, which contains some, but not necessarily all, of the objects that are members of the unparameterized type. Thus the parameters can turn off the sensitivity of some presentations that would otherwise be sensitive.

If there are no parameters and there is no corresponding CLOS class, there is not much `presentation-typep` can do. Returning T with the above attitude of "pairing any object with any presentation type is allowed" actually makes sense.

dkochmanski:
Thank you for starting this discussion, it is indeed a big change and we need to talk about it more. I still think that we should enforce the correspondence between the object and the presentation type. That said we should do it slowly: first signal (once) a style warning, then a warning, after that a continuable error and finally an error (for both missing `presentation-typep` method and object not being `presentation-typep` to the presentation).

I'll answer your remarks in the second reply, I need to read carefully and think about it carefully.

dkochmanski:
Here are my notes I've worked on yesterday (on and off)
https://gist.github.com/dkochmanski/34fa787574adb3a6cdbfff58100ae39d

dkochmanski:
I've disabled calling presentation-typep in the initialize-instance method (I've left the code, but commented out it with #+(or) and added a comment).

nsrahmad:
@dkochmanski I have read your document, and agree with most things. The only thing I disagree is making `presentation-type-of` not generic. It is really only useful if it is generic function. For example, adding `(defmethod clim:presentation-type-of ((person person)) 'person)` to your person example, `(present (make-instance 'person :age 15))` creates the presentation of type `person`. IMO it is better than having it presented as "standard-object".

Core/clim-core/presentations/presentation-types.lisp
```diff
@@ -42,8 +42,21 @@
 (defvar *print-presentation-verbose* nil)
 
 (defmethod initialize-instance :after ((instance presentation-mixin) &key)
-  (unless (presentation-type instance)
```
nsrahmad, line 45: Typos in commit message
"presentation is created AND the object is not"
"signal the error, when we try TO present"

Examples/dnd-commented.lisp
```diff
@@ -160,15 +160,7 @@ SIDEFUL-FIGURE  -- (or SQUARE TRIANGLE)
...
-SIDLESS-FIGURE-ACCEPTOR.
-
-Known issues
-------------
-
-- Compound destination types can't be dropped onto.
-
-Fixing that requires reworking frame-drag-and-drop method to accept ~
-any presentation in tracking-pointer and check type as we go.")))
+SIDLESS-FIGURE-ACCEPTOR.")))
```
nsrahmad, line 163: SIDLESS -> SIDELESS

dkochmanski:
when the presentation type `person` has required parameters, then specifying the presentation type as `person` is not valid. allowing specializations on the function presentation-type-of may introduce additional inconsistencies, i.e

```
(defclass foo () ())
(defclass bar () ())
(defmethod presentation-type-of ((x foo)) 'bar)
```
Another example, assuming that the presentation method foo has argument climi::params (then parameters are decoded)
```
(defclass person () ((age :initarg :age :accessor age)))
(define-presentation-type person (min max))
(defmethod presentation-type-of ((obj person)) 'person)
(define-presentation-method foo (object (type person))
  (<= min (age object) max))
(call foo object (presentation-type-of object))
```
foo will error on destructuring bind, because the presentation type person has parameters required parameters
`(destructuring-bind (name min max) 'person …)`

either way that document still needs work and we may discuss it after we are done with inheritance PRs. This pull request does not touch presentation-type-of.

Examples/puzzle.lisp
```diff
@@ -41,7 +41,7 @@
 	  command))))
 
 (define-presentation-type puzzle-cell ()
-  :inherit-from '(integer 1 15))
+  :inherit-from '(integer 0 15))
```
nsrahmad, line 44: `(integer 1 15)` is right. Only cells from 1 to 15 are interactive. You will never see the 0th cell.

Tests/presentations/presentation-inheritance.lisp
```diff
@@ -0,0 +1,27 @@
...
+;;; existing class, its inherit-from must match the superclasses of
+;;; the class itself. Otherwise an error should be signaled.
+(defclass pti.0001.class () ())
+(test pti.0001.inherit-expression
+  (fails
+    (signals error
+      (define-presentation-type pti.0001.class ()
+        :inherit-from 'expression))))
+
+;;; The presentation type PTI.0002.CLASS* should inherit its defautl
```
nsrahmad, line 14: typo "defautl" -> "default"

Examples/puzzle.lisp
dkochmanski: that's a good example of issues which may be revealed by calling presentation-typep when creating the presentation. 

`puzzle-cell` represents is a cell-id. The puzzle "0" may be presented on a cell with the id 4. Cell with id 0 may be presenting a puzzle with value 8.

Core/clim-core/presentations/presentation-types.lisp
dkochmanski: fixed

Examples/dnd-commented.lisp
dkochmanski: fixed

Tests/presentations/presentation-inheritance.lisp
dkochmanski: fixed

Examples/puzzle.lisp
nsrahmad: > puzzle-cell represents is a cell id.

No, it doesn't.

dkochmanski: ``
(updating-output (stream :unique-id cell-id 
				       :cache-value value)
		(formatting-cell (stream :align-x :right)
		  (unless (zerop value)
		    (with-output-as-presentation
			(stream cell-id 'puzzle-cell)
		      (format stream "~2D" value)))))
```

presented object here is` cell-id`, not `value`.

also, see this modified demo screencap which allows inspecting cells: http://turtleware.eu/static/paste/0f8f1506-cells.webm

Comment (nsrahmad) [reply Examples/puzzle.lisp:44]: All right, I see now. You have a very fundamental gap in understanding how presentations work.

Any CLIM based program can be divided into the application-logic part and the user-interface part.

`cell-id` here is part of the application-logic. it is the index into the `puzzle-array`.It has type `integer` and its value ranges from 0 to 15. `value` would be the the integer value at any index `cell-id` of `puzzle-array`. Value of `value` too ranges from 0 to 15.

Now cosider the user interface part.

A new presentation-type is defined as `puzzle-cell`. its type is defined as `integer`. ignore its range for now.

`with-output-as-presentation` takes an `stream`, an `object` and a `presentation-type` and creates a presentation of type `presentation-type`. Note that presentation itself is an object (This is the fundamental detail you miss). It stores the underlying object and its presentation type. In this example, `with-output-as-presentation (stream cell-id 'puzzle-cell)` would create a new presentation object of type `puzzle-cell` (of type integer, ignore range for now) with its underlying object `cell-id` (of type integer, ranges from 0 to 15).

Now consider the display function.
```lisp
(defmethod draw-puzzle ((puzzle puzzle) stream &key max-width max-height)
  (declare (ignore max-width max-height))
  (let ((puzzle-array (puzzle-puzzle puzzle)))
    (formatting-table (stream)
      (dotimes (row 4)
	(formatting-row (stream)
	  (dotimes (column 4)
	    (let* ((value (aref puzzle-array row column))
		   (cell-id (encode-puzzle-cell row column)))
	      (updating-output (stream :unique-id cell-id 
				       :cache-value value)
		(formatting-cell (stream :align-x :right)
		  (unless (zerop value)
		    (with-output-as-presentation
			(stream cell-id 'puzzle-cell)
		      (format stream "~2D" value))))))))))))
```

 It creates a grid of table cells from 0 to 15 to layout the `puzzle-array`. There are 16 cells. The display function iterates over these 16 cells, gets the index or `cell-id` (of type `integer`, ranges from 0 to 15), checks its `value` (of type integer, ranges from 0 to 15) and CREATES the presentation of type `puzzle-cell` (of type integer, ignore its range for now) and which displays the `value` on screen. Note that when the `value` is zero, it skips the `with-output-as-presentation`, No presentation object is created.

So for 16 cells, only 15 objects of type `puzzle-cell` is created, because as noted, the `value` zero is skipped. The type of `puzzle-cell` must be `integer` and it range must be from 1 to 15. Each of these `puzzle-cell` objects remember the underlying lisp object `cell-id` (of type integer, ranges from 0 to 15).

Another hint would be the `highlight-presentation` method, which flips the ink for any presentation of type `puzzle-cell` if it gets pointed by mouse. You could see that by pointing mouse to 1 to 15 on display.

And now, about Clouseau.

You should have asked yourself why you couldn't inspect the empty cell (with `cell-id` 15)?

As to why Clouseau shows the `cell-id` when you point to the `puzzle-cell` presentation-type? 

Because in the body of `define-presentation-action`, `object` is bound to the underlying lisp object. If you want to inspect the presentation itself, then modify it as follows (In the image, I clicked on `1`, the object is of type `puzzle-cell`, underlying lisp object is `0`):

Screenshot from 2020-07-08 13-26-43

dkochmanski: I think that my understanding of the concept is just fine. In this application, you manipulate presentations of type puzzle-cell which store objects ranging from 0-15, where you always display only 15 objects, because when the (application logic) cell value is 0, then the cell is not displayed. presentation-type puzzle-cell denotes the object which is the cell id, while its visual representation is drawn with format (and the content of a drawn value has nothing to do with the presentation type puzzle-cell, it has to do with value in the array stored in the application frame).

In other words, what is displayed on the screen, are cell ids, not values from the array puzzle.

dkochmanski:
I'm closing this PR and opening a new one, because I've changed it significantly and it requires a new review. The puzzle demo change stayed, but other things are quite different.

PR #1027, opened 2020-07-05
Fix out-of-extend use in DRAW-TEXT-OUTPUT-RECORD

scymtym:
It is not generally correct or safe to store the string passed to `medium-draw-text*` in the created `draw-text-output-record` instance:

* The string could be changed after the `medium-draw-text*` call:

  ```common-lisp
  (loop with buffer = (make-string 10)
        do (fill-buffer buffer)
           (draw-text* stream buffer ...))
  ```

* The string could have dynamic extent in which case storing it in the `draw-text-output-record` would lead to out-of-extent accesses:

  ```common-lisp
  (let ((string ...))
    (declare (dynamic-extent string))
    (draw-text* stream string ...))
  ```

  This is an actual problem which leads to bad crashes for seemingly innocent things like `(princ 5 <output-recording-stream>)` in recent SBCL versions.

The fix implemented in this change lets `generate-medium-recording-body` handle an argument called `string` by inserting a call to `subseq` with appropriate start and end positions and rebinding the variables `string`, `start` and `end` with appropriate values.

### DIFF
--- a/Core/clim-basic/extended-streams/recording.lisp
+++ b/Core/clim-basic/extended-streams/recording.lisp
@@ -1121,7 +1121,7 @@ were added."
 
 (defmacro generate-medium-recording-body (class-name args)
   (let ((arg-list (loop for arg in args
-                     nconc `(,(intern (symbol-name arg) :keyword) ,arg))))
+                        appending `(,(alexandria:make-keyword arg) ,arg))))
     `(progn
        (when (stream-recording-p stream)
          (let ((record
--- a/Core/clim-basic/extended-streams/recording.lisp
+++ b/Core/clim-basic/extended-streams/recording.lisp
@@ -1577,7 +1577,10 @@ were added."
     finally (return (values min-x min-y max-x max-y))))
 
 (def-grecording (draw-text :replay-fn nil) (gs-text-style-mixin gs-transformation-mixin)
-    (string start end point-x point-y align-x align-y
+    ((string (subseq string (or start 0) (or end (length string))))
+     (start  nil nil)
+     (end    nil nil)
+     point-x point-y align-x align-y
      toward-x toward-y transform-glyphs)
   ;; FIXME!!! Text direction.
   (let* ((transformation (medium-transformation medium))
@@ -1586,7 +1589,6 @@ were added."
     (multiple-value-bind (left top right bottom)
         (text-bounding-rectangle* medium string
                                   :align-x align-x :align-y align-y
-                                  :start start :end end
                                   :text-style text-style)
       (if transform-glyphs
           (enclosing-transform-polygon transformation (list (+ point-x left)  (+ point-y top)
@@ -1611,11 +1613,11 @@ were added."
     ((record draw-text-output-record) stream
      &optional (region +everywhere+) (x-offset 0) (y-offset 0))
   (declare (ignore x-offset y-offset region))
-  (with-slots (string point-x point-y start end align-x align-y toward-x
+  (with-slots (string point-x point-y align-x align-y toward-x
                toward-y transform-glyphs transformation)
       record
     (let ((medium (sheet-medium stream)))
-      (medium-draw-text* medium string point-x point-y start end align-x
+      (medium-draw-text* medium string point-x point-y 0 nil align-x
                          align-y toward-x toward-y transform-glyphs))))
 
 (defrecord-predicate draw-text-output-record
--- a/Core/clim-basic/extended-streams/recording.lisp
+++ b/Core/clim-basic/extended-streams/recording.lisp
@@ -1120,19 +1120,20 @@ were added."
         (sequence= my-coord-seq coord-seq #'coordinate=))))
 
 (defmacro generate-medium-recording-body (class-name args)
-  (let ((arg-list (loop for arg in args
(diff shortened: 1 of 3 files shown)


Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -1131,7 +1131,11 @@ were added."
                 ;; (arguably) less bad than untrasnforming the coords back to
                 ;; how they were.
                 (let (,@(when (member 'coord-seq args)
-                          `((coord-seq (copy-seq coord-seq)))))
+                          `((coord-seq (copy-seq coord-seq))))
+                      ,@(when (member 'string args)
+                          `((string (subseq string (or start 0) (or end (length string))))
+                            (start  0)
+                            (end    nil))))
```
dkochmanski, line 1138: I'd like rather medium-draw-text recording method be opencoded than picking up a single argument. Eventually, for this macro to scan all arguments for sequences and copy them without relying on the name.

That said this solution for strings is not complete, because output records may have strings added by means of the stream protocol, i.e add-string-output-to-text-record. When the string is styled, I think that it may be captured without "appending" to the existing string.

scymtym: > I'd like rather medium-draw-text recording method be opencoded than picking up a single argument.

Do you mean inlining? Or should the method be defined outside of the macro?

> Eventually, for this macro to scan all arguments for sequences and copy them without relying on the name.

I agree that looking at the name is pretty terrible, but I don't understand how the macro could decide which arguments will be sequences at runtime.

scymtym: > That said this solution for strings is not complete, because output records may have strings added by means of the stream protocol, i.e add-string-output-to-text-record. When the string is styled, I think that it may be captured without "appending" to the existing string.

I looked at `add-string-output-to-text-record` and it always copies the characters of the user-supplied string into an array owned by a `styled-string` so it should be fine. Are you aware of other places that keep references to user-supplied strings?

dkochmanski: you are right, in both arms of `if` we call replace. I'm not aware of other places.

dkochmanski: By open coding, I mean that the method should be defined outside of the macro. To inhibit the generation of the recording method there is a keyword argument ~medium-fn~ which needs to be specified as NIL.

I've meant that that the recording body could check the argument type at runtime and copy it on demand.

```
(defmethod medium-draw-text* ((medium recording) string)
  (let ((string (if (sequencep string) (copy-string …)))) …))
```
but that's indeed a terrible way to do things. So I think that the method should be written without using the macro. Draw-rectangle inhibits this method (notice, that it is defined as the ~around~ method to allow actual drawing).

dkochmanski: the call to the macro generate-medium-recording-body may be done inside the method of course, after copying the sequence. Since we are at it, we probably should do the same for coord-seq instead of relying on the argument name.

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -1392,12 +1410,12 @@ were added."
 
 (defrecord-predicate draw-polygon-output-record (closed filled)
   (and (if-supplied (closed)
-         (eql (slot-value record 'closed) closed))
+                    (eql (slot-value record 'closed) closed))
```
dkochmanski, line 1413: maybe we should introduce a dummy macro if-supplied (or rather, when supplied!) so the indentation is right and that arguments are hinted in the modeline. This indentation does not match the macro.

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -1150,42 +1151,50 @@ were added."
...
-                                                (x-offset 0) (y-offset 0))
-               (declare (ignore x-offset y-offset region))
-               (with-slots (,@args) record
-                 (let ((,medium (sheet-medium stream)))
-                   ;; Graphics state is set up in :around method.
-                   (,method-name ,medium ,@args)))))))))
-
-(def-grecording draw-point ((gs-line-style-mixin) point-x point-y) ()
+(defmacro def-grecording (name-and-options (&rest mixins) (&rest args)
+                          &body body)
```
dkochmanski, line 1155: I like this change of the macro lambda list, however it will require a longer comment explaining things.

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -1259,7 +1273,9 @@ were added."
        (if-supplied (point-y2 coordinate)
          (coordinate= (slot-value record 'point-y2) point-y2))))
 
-(def-grecording draw-lines ((coord-seq-mixin gs-line-style-mixin) coord-seq) ()
+;;; Regarding COORD-SEQ, see comment for, see comment for DRAW-POINTS.
```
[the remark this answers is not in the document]
scymtym: Fixed.

Core/clim-basic/extended-streams/recording.lisp
scymtym: Indentation is fixed. I would prefer to do further improvements in a separate pull request.

dkochmanski:
lgtm, thanks!

PR #1028, opened 2020-07-09
Presentation Types #1

dkochmanski:


nsrahmad:
There is one test related to `presentation-typep` that fails. Please see if it is actually required or redundant now.

Core/clim-core/presentations/presentation-type-functions.lisp
```diff
@@ -82,12 +82,17 @@ otherwise return false."
...
   (type-key parameters object type))
 
-;;; The default behavior is implemented in the function below. If we
-;;; end up here that means that the presentation parameters were
-;;; specified or that the presentation type does not correspond to a
-;;; class. In that case the method PRESENTATION-TYPEP must be
-;;; implemented by the programmer. -- jd 2020-07-02
-(define-default-presentation-method presentation-typep (object type)
+;;; The following two methods are defined for inheritance and combined
+;;; implement a default behavior of the presentation method
```
nsrahmad, line 86: "and combined, they implement the default behavior of presentation-typep."

dkochmanski: fixed, however, I've kept "of the presentation method", because otherwise this statement would be ambigous. they do not implement the default behavior of the function presentation-typep.

dkochmanski:
I've cherry picked the fix for this test from the "Presentation Types #2" PR, I'll remove and rebase that branch too. all fixed here.

PR #1032, opened 2020-07-13
recording: fix DEFRECORD-PREDICATE DRAW-TEXT-OUTPUT-RECORD again

scymtym:
The slots `start` and `end` have been removed from `draw-text-output-record` in ab85d944 but the methods generated by `(defrecord-predicate draw-text-output-record)` still tried to include them in the comparison.

To fix the issue, `defrecord-predicate` can now accept things that are keyword arguments but not slots and can also generate different comparison code for the slot and the argument cases.

### DIFF
--- a/Core/clim-basic/extended-streams/recording.lisp
+++ b/Core/clim-basic/extended-streams/recording.lisp
@@ -142,22 +142,15 @@ all of FUNCTION-ARGS as APPLY arguments."
 ;;; methods are very similar, hence this macro.  In order to exploit
 ;;; the similarities, it's necessary to treat the slots of the second
 ;;; record like variables, so for convenience the macro will use
-;;; SLOT-VALUE on both records.
-
-(defmacro defrecord-predicate (record-type slots &body body)
-  "Each element of SLOTS is either a symbol or (:initarg-name slot-name)."
-  (let* ((slot-names (mapcar #'(lambda (slot-spec)
-                                 (if (consp slot-spec)
-                                     (cadr slot-spec)
-                                     slot-spec))
-                             slots))
-         (supplied-vars (mapcar #'(lambda (slot)
-                                    (gensym (symbol-name
-                                             (symbol-concat slot '#:-p))))
+;;; WITH-SLOTS on the second record.
+(defmacro defrecord-predicate (record-type slot-names &body body)
+  "Each element of SLOT-NAMES is a symbol naming a slot."
+  (let* ((supplied-vars (mapcar (lambda (name)
+                                  (gensym (format nil "~A-~A" name '#:p)))
                                 slot-names))
-         (key-args (mapcar #'(lambda (slot-spec supplied)
-                               `(,slot-spec nil ,supplied))
-                           slots supplied-vars))
+         (key-args (mapcar (lambda (name supplied)
+                             `(,name nil ,supplied))
+                           slot-names supplied-vars))
          (key-arg-alist (mapcar #'cons slot-names supplied-vars)))
     `(progn
        (defmethod output-record-equal and ((record ,record-type)
@@ -991,9 +984,9 @@ were added."
         (with-drawing-options (stream :clipping-region (graphics-state-clip record))
           (call-next-method)))))
 
-(defrecord-predicate gs-clip-mixin ((:clipping-region clipping-region))
+(defrecord-predicate gs-clip-mixin (clipping-region)
   (if-supplied (clipping-region)
-    (region-equal (graphics-state-clip record) clipping-region)))
+    (region-equal (slot-value record 'clipping-region) clipping-region)))
 
 ;;; 16.3.2. Graphics Displayed Output Records
 (defclass standard-displayed-output-record (gs-clip-mixin gs-ink-mixin
--- a/Core/clim-basic/extended-streams/recording.lisp
+++ b/Core/clim-basic/extended-streams/recording.lisp
@@ -143,37 +143,41 @@ all of FUNCTION-ARGS as APPLY arguments."
 ;;; the similarities, it's necessary to treat the slots of the second
 ;;; record like variables, so for convenience the macro will use
 ;;; WITH-SLOTS on the second record.
-(defmacro defrecord-predicate (record-type slot-names &body body)
-  "Each element of SLOT-NAMES is a symbol naming a slot."
-  (let* ((supplied-vars (mapcar (lambda (name)
-                                  (gensym (format nil "~A-~A" name '#:p)))
-                                slot-names))
-         (key-args (mapcar (lambda (name supplied)
-                             `(,name nil ,supplied))
-                           slot-names supplied-vars))
-         (key-arg-alist (mapcar #'cons slot-names supplied-vars)))
+(defmacro defrecord-predicate (record-type slots &body body)
+  "Each element of SLOTS is either a symbol naming a slot
+or (SLOT-NAME SLOT-P)."
+  (multiple-value-bind (slot-names key-args key-arg-alist)
+      (loop for slot-spec in slots
+            for (name slot-p) = (if (atom slot-spec)
+                                    (list slot-spec t)
+                                    slot-spec)
+            for suppliedp = (gensym (format nil "~A-~A" name '#:p))
+            when slot-p collect name into slot-names
+            collect `(,name nil ,suppliedp) into key-args
+            collect (cons name suppliedp) into key-arg-alist
+            finally (return (values slot-names key-args key-arg-alist)))
     `(progn
        (defmethod output-record-equal and ((record ,record-type)
                                            (record2 ,record-type))
          (macrolet ((if-supplied ((var &optional (type t)) &body supplied-body)
-                      (declare (ignore var type))
-                      `(progn ,@supplied-body)))
+                      (declare (ignore type))
+                      (when (find var ',slot-names :test #'eq)
+                        `(progn ,@supplied-body))))
            (with-slots ,slot-names record2
              ,@body)))
        (defmethod match-output-records-1 and ((record ,record-type)
                                               &key ,@key-args)
          (macrolet ((if-supplied ((var &optional (type t)) &body supplied-body)
-                      (let ((supplied-var (cdr (assoc var ',key-arg-alist))))
-                        (unless supplied-var
-                          (error "Unknown slot ~S" var))
+                      (let ((supplied-var (or (cdr (assoc var ',key-arg-alist))
+                                              (error "Unknown argument ~S" var))))
                         `(or (null ,supplied-var)
-                             ,@(if (eq type t)
-                                   `((progn ,@supplied-body))
-                                   `((if (typep ,var ',type)
-                                         (progn ,@supplied-body)
-                                         (error 'type-error
-                                                :datum ,var
-                                                :expected-type ',type))))))))
+                             ,(if (eq type t)
+                                  `(progn ,@supplied-body)
+                                  `(if (typep ,var ',type)
+                                       (progn ,@supplied-body)
+                                       (error 'type-error
+                                              :datum ,var
+                                              :expected-type ',type)))))))
            ,@body)))))
 
 (defmacro with-output-recording-options ((stream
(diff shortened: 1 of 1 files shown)


Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -143,44 +143,47 @@ all of FUNCTION-ARGS as APPLY arguments."
...
+          do (push (cons name suppliedp) key-arg-alist))
     `(progn
        (defmethod output-record-equal and ((record ,record-type)
                                            (record2 ,record-type))
-         (macrolet ((if-supplied ((var &optional (type t)) &body supplied-body)
-                      (declare (ignore var type))
-                      `(progn ,@supplied-body)))
-           (with-slots ,slot-names record2
+         (macrolet ((if-supplied ((var &optional (type t))
+                                  slot-form &optional arg-form)
```
dkochmanski, line 167: I think that changing the if-supplied macro lambda list doesn't do much good (i.e it is less comprehensible). The test for the text-output record could be expressed as:

```
(if-supplied (string)
  (let ((start (or (if-supplied (start) start) 0))
        (end (or (if-supplied (end) end) nil)))
    (string= (slot-value record 'string) string
             :start2 start :end end)))
```

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -143,44 +143,47 @@ all of FUNCTION-ARGS as APPLY arguments."
...
+                      (declare (ignore type arg-form))
+                      (when (find var ',class-slots :test #'eq)
+                        slot-form)))
```
dkochmanski, line 170: ditto, `(progn ,@supplied-body)`

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -143,44 +143,47 @@ all of FUNCTION-ARGS as APPLY arguments."
...
-                             ,@(if (eq type t)
-                                   `((progn ,@supplied-body))
-                                   `((if (typep ,var ',type)
-                                         (progn ,@supplied-body)
-                                         (error 'type-error
-                                                :datum ,var
-                                                :expected-type ',type))))))))
+                                              &key ,@(nreverse key-args))
+         (macrolet ((if-supplied ((var &optional (type t))
+                                  slot-form &optional (arg-form slot-form))
```
dkochmanski, line 176: `&body supplied-body`

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -1622,17 +1625,22 @@ were added."
...
        (if-supplied (point-y coordinate)
          (coordinate= (slot-value record 'point-y) point-y))
+       (if-supplied (string)
+         (string= (slot-value record 'string) string)
+         (let ((start2 0)
+               (end2   nil))
+           (if-supplied (start) (setf start2 start))
+           (if-supplied (end) (setf end2 end))
+           (string= (slot-value record 'string) string
+                    :start2 start2 :end2 end2)))
```
dkochmanski, line 1643: ditto, no need for separate forms for existing and non-existing slots.

dkochmanski: of course, that requires somewhat intimate knowledge of the implementer what is a slot and what is not, but that requirement is true for both cases.

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -143,44 +143,47 @@ all of FUNCTION-ARGS as APPLY arguments."
...
+                      (let ((supplied-var (or (cdr (assoc var ',key-arg-alist))
+                                              (error "Unknown slot ~S" var))))
+                        `(when ,supplied-var
+                           ,(if (eq type t)
+                                arg-form
+                                `(if (typep ,var ',type)
+                                     ,arg-form
+                                     (error 'type-error
+                                            :datum ,var
+                                            :expected-type ',type)))))))
```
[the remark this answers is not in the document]
scymtym: I was actually wondering why check the type at all? I seems completely arbitrary.

Core/clim-basic/extended-streams/recording.lisp
scymtym: I'm not sure I understand this suggestion. Wouldn't that try to access `start` and `end` as slots in the method that compares two output record objects?

Core/clim-basic/extended-streams/recording.lisp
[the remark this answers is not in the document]
dkochmanski: when the type is known I think that it makes sense to check it. We should do it in more places though. Example:

(if-supplied (ink design)
    (design-equalp (slot-value record 'ink) ink))

if ink is not a design, then we will never call design-equalp (in code the type is not declared though). It seems to be mostly checked for coordinates – in this case that may improve the performance, because we know that we can inline some operations.

Core/clim-basic/extended-streams/recording.lisp
dkochmanski: in your code (find var ',class-slots :test #'eq) will yield nil, so the body won't be expanded,

(or (if-supplied (start) start) 0) will expand to
(or nil 0)

Core/clim-basic/extended-streams/recording.lisp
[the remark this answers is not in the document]
scymtym: I would prefer to just fix the regression for now. We can work on the type checking later.

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -143,44 +143,47 @@ all of FUNCTION-ARGS as APPLY arguments."
...
-                                   `((if (typep ,var ',type)
-                                         (progn ,@supplied-body)
-                                         (error 'type-error
-                                                :datum ,var
-                                                :expected-type ',type))))))))
+                                              &key ,@(nreverse key-args))
+         (macrolet ((if-supplied ((var &optional (type t))
+                                  slot-form &optional (arg-form slot-form))
+                      (let ((supplied-var (or (cdr (assoc var ',key-arg-alist))
+                                              (error "Unknown slot ~S" var))))
```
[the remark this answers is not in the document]
scymtym: Changed.

Core/clim-basic/extended-streams/recording.lisp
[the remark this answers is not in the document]
dkochmanski: I'm not asking you to add annotations to other if-supplied forms, only to change the expansion (and I doubt we will get back to it before 2025). It doesn't matter though, because the change would be cosmetic.

That said, I've spotted another problem here: you change the semantics of the `if-supplied` macro.

Previously: when the argument is not supplied, returns T
Currently: when the argument is not supplied returns NIL

The latter is incorrect, consider
```
(defrecord-predicate gs-ink-mixin (ink)
  (if-supplied (ink)
    (design-equalp (slot-value record 'ink) ink)))
```
when the ink is not supplied, then predicate should default to T. Please bring back the old form (especially if we agree about not changing the lambda list of the macro -- this change won't be necessary at all then.

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -142,45 +142,41 @@ all of FUNCTION-ARGS as APPLY arguments."
...
+                                              (error "Unknown argument ~S" var))))
+                        `(cond ((null ,supplied-var)
+                                t)
+                               ,@(unless (eq type t)
+                                   `(((not (typep ,var ',type))
+                                      (error 'type-error
+                                             :datum ,var
+                                             :expected-type ',type))))
+                               (t
+                                ,@supplied-body)))))
```
dkochmanski, line 179: I find the previous version of this block much more readable, please only change the error string.

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -1622,17 +1618,26 @@ were added."
...
        (if-supplied (point-y coordinate)
          (coordinate= (slot-value record 'point-y) point-y))
+       ;; START and END can be supplied as keyword arguments, but the
+       ;; output record does not store them in slots. For
+       ;; MATCH-OUTPUT-RECORDS-1, compare the designated subsequence
+       ;; of the STRING keyword argument to the entire string stored
+       ;; in the output record.
+       (if-supplied (string)
+         (let ((start2 0)
+               (end2   nil))
```
dkochmanski, line 1636: remove column alignment

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -142,45 +142,41 @@ all of FUNCTION-ARGS as APPLY arguments."
...
-         (supplied-vars (mapcar #'(lambda (slot)
-                                    (gensym (symbol-name
-                                             (symbol-concat slot '#:-p))))
-                                slot-names))
-         (key-args (mapcar #'(lambda (slot-spec supplied)
-                               `(,slot-spec nil ,supplied))
-                           slots supplied-vars))
-         (key-arg-alist (mapcar #'cons slot-names supplied-vars)))
+  "Each element of SLOTS is either a symbol naming a slot
+or (SLOT-NAME ARGUMENT-ONLY-P)."
```
dkochmanski, line 148: the previous default was more intuitive: by default, it is a slot, you need to explicitly specify, that it is not (by putting nil as a second argument)

so argument-only-p -> slotp and proper destructuring with &optional (slotp t)

Core/clim-basic/extended-streams/recording.lisp
```diff
@@ -143,44 +143,47 @@ all of FUNCTION-ARGS as APPLY arguments."
...
+        (class-slots '()))
+    (loop for spec in slots
+          for (keyword name) = (if (consp spec)
+                                   spec
+                                   (list (alexandria:make-keyword spec)
+                                         spec))
+          for suppliedp = (gensym (format nil "~A-P" keyword))
+          for slotp = (if (and (consp spec) (>= (length spec) 3))
+                          (third spec)
+                          t)
```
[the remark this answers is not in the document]
dkochmanski: (apply #'push-args (ensure-list spec))

Core/clim-basic/extended-streams/recording.lisp
scymtym: Changed.

Core/clim-basic/extended-streams/recording.lisp
scymtym: This is only used in one heavily commented place. I think keeping the macro as simple as possible makes more sense in this case.

> so argument-only-p -> slotp and proper destructuring with &optional (slotp t)

Destructuring is not needed beyond `for (name argument-only-p) = (alexandria:ensure-list slot-spec)`. I think your comment about the destructuring part somehow got attached to an outdated version of the code.

Core/clim-basic/extended-streams/recording.lisp
scymtym: Opposite for me but I change it back if you want.

scymtym: I would still change `,@(if … '((…)) '((…)))` to `,(if … '(…) '(…))` (backticks replaced with single quote for markdown compatibility) though, unless there is reason not to that I missed.

dkochmanski: I don't see why not either. you may also pull the error to let with or like you do already.

Core/clim-basic/extended-streams/recording.lisp
dkochmanski: foo means slot (the default), but
(foo t) means *not* slot
(foo nil) means *not* *not* a slot

confirming negation (i.e T stands for "true that it is not a slot") is very weird here

scymtym: I flipped the meaning from `argument-only-p` to `slot-p`.

Core/clim-basic/extended-streams/recording.lisp
scymtym: Done.

Core/clim-basic/extended-streams/recording.lisp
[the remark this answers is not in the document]
scymtym: With the keyword processing removed, the current `loop`-based version seems simple enough. In particular, it avoids a helper function and reversing of the collected lists before splicing.

dkochmanski:
lgtm, thanks

PR #1035, opened 2020-07-21
More cleanup and optimizations for raster graphics

scymtym:
* Improvements to ARGB representation to avoid shifting for each pixel

* Cache-friendly loops over raster image pixels

* Various minor cleanups

* Image transform demo shows transformed text which backends should be
  able to handle

* Remove obsolete copy of `menu-choose.lisp` in `Experimental`
  directory


Examples/image-transform-demo.lisp
```diff
@@ -111,19 +113,14 @@
...
-        (clim:draw-rectangle* stream 0 0 (clim:pattern-width image) (clim:pattern-height image) :filled nil :ink clim:+blue+)
-        ;; We don't display text here if using the Truetype font
-        ;; renderer, since other font renderers doesn't support text
-        ;; transform.
-        #+mcclim-ffi-freetype
+        (clim:draw-rectangle* stream 0 0 width height :filled nil :ink clim:+blue+)
         (clim:with-text-size (stream 60)
-          (clim:draw-text* stream "Foo abcdefgh" 100 100))))))
+          (clim:draw-text* stream "Can transform textr, too" 100 100
+                           :transform-glyphs t))))))
```
dkochmanski, line 123: s/textr/text/

Core/clim-basic/drawing/pattern.lisp
```diff
@@ -99,42 +99,50 @@ pattern, stencil, image etc)."))
...
        (multiple-value-bind (red green blue opacity)
            (color-rgba element)
-         (logior (ash (transform red)   24)
-                 (ash (transform green) 16)
-                 (ash (transform blue)   8)
-                 (transform opacity))))
+         (logior (ash (transform red)    24)
+                 (ash (transform green)  16)
+                 (ash (transform blue)    8)
+                 (ash (transform opacity) 0))))
```
dkochmanski: what's the point of calling ash here?

Core/clim-basic/drawing/pattern.lisp
```diff
@@ -99,42 +99,50 @@ pattern, stencil, image etc)."))
...
-                                   (float (/ (ldb (byte 8 00) elt) 255))))
+        (let* ((rgba-value (aref array y x))
+               (alpha (ldb (byte 8 0) rgba-value))
+               (color (make-rgb-color (/ (ldb (byte 8 24) rgba-value) 255.0)
+                                      (/ (ldb (byte 8 16) rgba-value) 255.0)
+                                      (/ (ldb (byte 8  8) rgba-value) 255.0))))
+          (case alpha
+            (0 +transparent-ink+)
+            (255 color)
+            (t (make-uniform-compositum color (/ alpha 255.0)))))
```
dkochmanski, line 145: alpha is known before the color is computed, so maybe

```
(let* ((rgba-value  …)
         (alpha …))
  (case alpha
    (0 +transaprent-ink+)
    (255 (make-color))
    (otherwise (make-uniform-compositum (make-color) (/ alpha 255.0))))
```
?

dkochmanski: another question: why do we avoid calling a function make-uniform-compositum? when alpha is opaque, then it returns the ink (but it accepts normalized alpha value).

```
(defun make-uniform-compositum (ink opacity-value)
  (cond ((= opacity-value 0)
         +transparent-ink+)
        ((= opacity-value 1)
         ink)
        (t
         (make-instance 'uniform-compositum
           :ink ink
           :mask (make-opacity opacity-value)))))
```

n.b since it is an internal function, we could define also `make-uniform-compositum*` which takes values [0,255] instead of [0,1] and call that for patterns.

Backends/CLX-fb/mirror.lisp
```diff
@@ -21,7 +21,7 @@
 	    height h)
       (setf xlib-image (make-array (list height width)
                                    :element-type '(unsigned-byte 32)
-                                   :initial-element #x00FFFFFF))
+                                   :initial-element #xffffffff))
```
dkochmanski, line 24: shouldn't that be #ffffff00 ?

Examples/image-transform-demo.lisp
scymtym: Fixed.

Core/clim-basic/drawing/pattern.lisp
scymtym: Readability. The call surely gets optimized away.

Core/clim-basic/drawing/pattern.lisp
scymtym: Not calling `make-uniform-compositum` avoids computing `(/ alpha 255.0)` for the (presumably common) cases which would throw it away immediately.

I'm not sure making `make-uniform-compositum*` is worth it. Especially if this is the only call site.

scymtym: Only computing the color if needed seems like a good idea.

Backends/CLX-fb/mirror.lisp
scymtym: If I understand correctly, X only uses the color channels, that is `(byte 0 24)` and the color is only visible if the backend fail to draw all pixels of the mirror. So maybe something recognizable like `#xc080e0a0` would be good? The screenshot shows the suggested color.
screenshot9

PR #1037, opened 2020-07-24
Major performance improvements in `fill-image`

scymtym:
This gets the drawing benchmark in "red rectangle" mode to 466 on my machine.

Change details:

* Package cleanup

* New types `image-{dimension,index-displacement}`

* Declare types.

* Disregard `clip-region`, in particular the per-pixel `region-contains-position-p` check, if the image region is entirely contained in it.

* Instead of unconditionally and fully updating the alpha value and the ink-based color values once per pixel, only perform the necessary computations and do not recompute unchanged quantities.

* Depending on the flipping and blending mode, skip unnecessary flipping and blending computations.

### DIFF
--- a/Backends/CLX-fb/package.lisp
+++ b/Backends/CLX-fb/package.lisp
@@ -1,73 +1,49 @@
-;;; -*- Mode: Lisp; Package: COMMON-LISP-USER -*-
+;;; ---------------------------------------------------------------------------
+;;;   License: LGPL-2.1+ (See file 'Copyright' for details).
+;;; ---------------------------------------------------------------------------
+;;;
+;;;  (c) copyright 2016 Alessandro Serra <gas2serra@gmail.com>
+;;;  (c) copyright 2017-2020 Daniel Kochmański <daniel@turtleware.eu>
+;;;  (c) copyright 2020 Jan Moringen <jmoringe@techfak.uni-bielefeld.de>
+;;;
+;;; ---------------------------------------------------------------------------
+;;;
+;;; Package definition for the clx-fb module.
 
-(in-package :common-lisp-user)
-
-(defpackage :clim-clx-fb
-    (:use :clim :clim-lisp :clim-backend :clim-clx :mcclim-render-extensions)
-  (:import-from :climi
-                #:+alt-key+
-                ;;
-                #:port-text-style-mappings
+(defpackage #:clim-clx-fb
+  (:use #:clim
+        #:clim-lisp
+        #:clim-backend
+        #:clim-clx
+        #:mcclim-render-extensions)
+  (:import-from #:climi
                 #:port-lookup-mirror
                 #:port-register-mirror
-                #:port-event-process
                 #:port-grafts
-                #:%%sheet-native-transformation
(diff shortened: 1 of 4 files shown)


Extensions/render/image.lisp
```diff
@@ -116,43 +116,117 @@
...
+                         (octet-blend-function* r    g    b    source-a
+                                                r.bg g.bg b.bg a.bg))))))
+            (:copy
+             (setf (aref dst-array j i) source-rgba))
+            (:blend
+             (let-rgba ((r.bg g.bg b.bg a.bg) (aref dst-array j i))
+               (setf (aref dst-array j i)
+                     (octet-blend-function*
+                      source-r source-g source-b source-a
+                      r.bg     g.bg     b.bg     a.bg)))))))))
```
dkochmanski, line 229: non-consistent line break with flipping/blend

Extensions/render/image.lisp
```diff
@@ -116,43 +116,117 @@
...
+                                           nil
+                                           :blend))))))))
+      (do-regions ((src-j j y y y2)
+                   (src-i i x x x2))
+        (when (or (null clip-region)
+                  (region-contains-position-p clip-region src-i src-j))
+          (when stencil-array
+            (update-alpha i j))
+          (update-ink i j)
+          (case mode ; do nothing if MODE is NIL
```
dkochmanski, line 211: [optional] maybe add empty otherwise clause instead of a comment?

Extensions/render/image.lisp
```diff
@@ -116,43 +116,117 @@
...
+                   ((typep ink 'standard-flipping-ink)
+                    (setf source-rgba (let ((d1 (slot-value ink 'climi::design1))
+                                            (d2 (slot-value ink 'climi::design2)))
+                                        (logand #x00ffffff
+                                                (logxor (climi::%rgba-value d1)
+                                                        (climi::%rgba-value d2)))))
+                    (if (= alpha 255)
+                        (setf mode :flipping)
+                        (setf source-a alpha
+                              mode     :flipping/blend)))
```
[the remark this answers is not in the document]
scymtym: I think I removed most.

Extensions/render/image.lisp
scymtym: Changed.

Extensions/render/image.lisp
scymtym: I would use an `otherwise` clause if I wanted to emphasize the resulting value. In the case, I want to emphasize that nothing is done, so not having a clause seems better to me.

Backends/CLX-fb/package.lisp
```diff
@@ -1,73 +1,49 @@
...
-                #:get-transformation
                 ;;
-                #:medium-miter-limit
-                ;; classes:
-                #:mirrored-pixmap
-                #:window-destroy-event
-                #:device-font-text-style
+                #:frame-managers        ; used as slot
+                #:height                ; used to access a slot
+                #:width                 ; dito
```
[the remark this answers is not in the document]
scymtym: Fixed.

PR #851, opened 2019-10-02
Improve click to focus

dkochmanski:

PR #1038, opened 2020-07-27
menu: don't close on first pointer-button-release-event

glv2:
This allows navigating the menus without having to keep the pointer button
constantly pressed. A click on a menu item selects it, and a click on the title
of an open menu closes it.


dkochmanski:
Thanks, this change is something that I want to add to McCLIM. That said, in this PR menu does not close when I click somewhere else in the application.

In this PR the menu does not close when I click somewhere else in the application. Even worse, it persists even after I close the frame. The proper solution will require some refactor in menus, but the crucial thing is the input focus -- when the menu will lose focus it will change. See #851. I'm closing this PR for a time being.

PR #1044, opened 2020-08-07
seos: do not assume stream-char-height after a new line

dkochmanski:
Consider the following snippet:

    (with-text-size (stream :huge)
      (format stream "HEY1!"))
    (fresh-line stream)
    (format stream "HEY2!~%")
    (format stream "HEY3!~%")
    (format stream "HEY4!~%")

When the above code is executed, then FRESH-LINE caused advancing the cursor
vertically by the STREAM-CHAR-HEIGHT of the line with the size :HUGE), and did
not reset the STREAM-CHAR-HEIGHT, so after the line "HEY2!" a vertical
advancement was once again the STREAM-CHAR-HEIGHT line with the size :HUGE.

    (with-text-size (stream :huge)
      (format stream "HEY1!~%"))
    (format stream "HEY2!~%")
    (format stream "HEY3!~%")

The situation is similar, however it is slightly different: the newline is
called when a text style with the size :HUGE is still the context of
drawing. That illustrates, that we should not set STREAM-CHAR-HEIGHT to the
current text-style height after the line break, but rather wait for the first
character written.

The new approach is to reset the STREAM-CHAR-HEIGHT to 0 after each line break
and call MAXF on it only when the text is drawn. After this change the
vertical advancement of the cursor after the new line remains the same (that
is the "old line" height is used), only the initial baseline is set to 0.

### DIFF
--- a/Core/clim-basic/extended-streams/stream-output.lisp
+++ b/Core/clim-basic/extended-streams/stream-output.lisp
@@ -195,7 +195,6 @@
                          :y-position y-start)))
   (setf (cursor-active (stream-text-cursor stream)) t))
 
-
 (defmethod stream-cursor-position ((stream standard-extended-output-stream))
   (cursor-position (stream-text-cursor stream)))
 
@@ -291,7 +290,9 @@
       (let* ((medium       (sheet-medium stream))
              (text-style   (medium-text-style medium))
              (new-baseline (text-style-ascent text-style medium))
-             (new-height   (text-style-height text-style medium)))
+             (new-height   0))
+        ;; For new lines we reset the char height to 0 in case of the text
+        ;; style change after the line break. -- jd 2020-08-07
         (maybe-end-of-page-action stream (+ updated-cy new-height))
         (setf (slot-value stream 'baseline) new-baseline
               (%stream-char-height stream)  new-height)))))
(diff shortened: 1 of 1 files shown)


Core/clim-basic/extended-streams/stream-output.lisp
```diff
@@ -350,7 +351,9 @@
                                              :margin margin
                                              :break-strategy break
                                              :start start :end end))
-              do (stream-write-output stream string start split)
+              do (maxf (%stream-char-height stream)
+                       (text-style-height text-style medium))
```
scymtym, line 355: Doing this within the loop seems at odds with the comment "Also text-style doesn't change until the end of this function. -- jd 2019-01-10" a few lines above. Given that the text style is stored in a lexical variable and, as far as I can tell, not changed, that comment seems to be right.

dkochmanski: right, I've bound text-style-height once. fixed.

we need to call maxf repeatedly after each line because we don't want to set the stream-char-height for a line to which we do not write (i.e `"foo\nbar\n[no-further-output]"`), and to do that we need to zero the stream-char-height after each newline.

PR #1048, opened 2020-08-11
Optionally update frame instances when class is redefined

scymtym:
* Minor cosmetic changes

* New module for frame-related code

* Changes to `accept` as well as gadget to make them work better with frame class redefinition

* The actual frame class redefinition changes in `frames/frames.lisp`, `frames/define-application-frame.lisp` and the new file `frames/redefinition.lisp`

* A section on frame class redefinition in the manual

* An example for frame class redefinition


Documentation/Manual/Texinfo/extensions.texi
```diff
@@ -321,3 +321,59 @@ image. Valid values are @code{:png}, @code{:jpg}, @code{:jpeg},
...
+selected if the @code{(:update-instances-on-redefinition t)} option is
+supplied in the @code{clim:define-application-frame} form in
+question@footnote{Note that it is currently not possible to change this
+option after the initial definition of an application frame class.  This
+may be desirable anyway since changing this option after creating
+instances of the class would have unclear semantics.}.  When an
+application frame class which has been defined with this option is
+redefined by re-evaluating the @code{clim:define-application-frame} form
+or reinitializing the class in some other way, the following changes are
+applied to each existing instance of that class:
```
dkochmanski, line 348: one could imagine, that semantics of that are as follows:
instances which are defined while the frame class has this flag are updated on redefinition
instances which were defined when this flag was nil are not redefined no matter what

this is just a suggestion

dkochmanski:
I'm closing this for now.

Issue #1045, opened 2020-08-08
Layout protocol


admich:
I think that McCLIM implementation of layout protocol have some problem. Here some issue:

1. `change-space-requirements` is called from `redisplay-frame-pane`
   on `clim-stream-pane` unless one of `pane-user-{width,height...}`
   is `:compute`. But the space requirement of a clim-pane change only
   if one of the `pane-user-{width,height...}` is :compute, exactly
   the opposite.
2. `change-space-requirements` calls `change-space-requirements` on
   the parent, but a pane can not know if the space requirements of
   its parent will change or not, this is must be a decision of the
   parent. So `change-space-requirements` must only call
   `note-space-requirements-changed` and then the parent decide what
   to do.
3. Due to 1. and 2 `layout-frame` is called too much, e.g. in the Data
   Graph Toy examples it is called each time a new point is drawn
   (once each second)
4. :resize-frame initargs. This is a problem in the specification,
   this argument is cited in `change-space-requirements` description
   as a define-application-frame keywords but then it is not described
   there.
5. `restraining-pane` is defined but not implemented. Also Franz and
   LispWorks CLIM didnn't define it, but Lucid CLIM yes (its written
   in LispWorks CLIM manual, so inside Lispworks Ltd there are two
   closed source CLIM implementation LispWors and Lucid, I didn't
   known).
6. pane slot `space-requirement` and its accessor
   `pane-space-requirement` is defined two times, in `basic-pane` and
   in `layout-protocol-mixin`.
7. I don't understand why there are `pane-current-width` and
   `pane-current-height` to known the actual dimension on a pane there
   is already `sheet-region`.

admich:
Another problem in the layout protocol is the following:
8. If I change the layout of an application with `(setf
   (frame-current-layout))` (e.g. n Clouseau with CTRL-i ) the
   application frame is resized (in a stacking wm). I don't think this
   is what a user want.

dkochmanski:
I've recently noticed some problems which partially overlap with your remarks, most notably regarding frequent changes of the space requirements.

scymtym:
#1050 has been integrated without technically merging it.

PR #1050, opened 2020-08-12
Layout protocol

admich:
Try to improve the layout protocol fixing some issue explained on #1045.

### DIFF
--- a/Core/clim-core/frames/frames.lisp
+++ b/Core/clim-core/frames/frames.lisp
@@ -121,7 +121,9 @@ input focus. This is a McCLIM extension."))
    (panes-for-layout :initform nil :accessor frame-panes-for-layout
                      :documentation "alist of names and panes
                                      (as returned by make-pane)")
-
+   (resize-frame :initarg :resize-frame
+                 :initform nil
+                 :accessor frame-resize-frame)
    (output-pane :initform nil
                 :accessor frame-standard-output
                 :accessor frame-error-output)
@@ -279,9 +281,16 @@ documentation produced by presentations.")
 (defmethod (setf frame-current-layout) :around (name (frame application-frame))
   (unless (eql name (frame-current-layout frame))
     (call-next-method)
-    (alexandria:when-let ((fm (frame-manager frame)))
-      (generate-panes fm frame)
-      (layout-frame frame)
+    (when-let ((fm (frame-manager frame)))
+      (if-let ((tls (and (frame-resize-frame frame)
+                         (frame-top-level-sheet frame))))
+        (multiple-value-bind (width height)
+            (bounding-rectangle-size tls)
+          (generate-panes fm frame)
+          (layout-frame frame width height))
+        (progn
+          (generate-panes fm frame)
+          (layout-frame frame)))
       (signal 'frame-layout-changed :frame frame))))
 
 (defmethod (setf frame-command-table) :around (new-command-table frame)
--- a/Core/clim-core/menu-choose.lisp
+++ b/Core/clim-core/menu-choose.lisp
@@ -310,7 +310,8 @@ maximum size according to `frame')."
     (when-let ((viewport (pane-viewport menu)))
       (multiple-value-bind (viewport-width viewport-height)
           (menu-size viewport *application-frame*)
-        (let ((scroller (pane-scroller menu)))
+        (multiple-value-bind (scroller-width scroller-height)
+            (bounding-rectangle-size (pane-scroller menu))
           (change-space-requirements scroller
                                      ;; HACK: How are you supposed to
                                      ;; change the size of the viewport?
@@ -323,11 +324,9 @@ maximum size according to `frame')."
                                      ;; difference (to make room for
                                      ;; scroll bars).
                                      :width (+ menu-width
-                                               (- (pane-current-width scroller)
-                                                  viewport-width))
+                                               (- scroller-width viewport-width))
                                      :height (+ menu-height
-                                                (- (pane-current-height scroller)
-                                                   viewport-height))
+                                                (- scroller-height viewport-height))
                                      :resize-frame t))))
 
     ;; Modify the size and location of the frame as well.
--- a/Core/clim-core/panes/composition.lisp
+++ b/Core/clim-core/panes/composition.lisp
@@ -279,13 +279,14 @@
                  *changed-space-requirements*)))
         (t
          (let ((frame (pane-frame pane)))
-           ;; ### we miss the :resize-frame option
            (cond (resize-frame
                   (layout-frame frame))
                  (t
-                  (layout-frame frame
-                                (bounding-rectangle-width pane)
-                                (bounding-rectangle-height pane))))))))
+                  (if (frame-resize-frame frame)
+                      (layout-frame frame)
+                      (multiple-value-bind (width height)
+                          (bounding-rectangle-size pane)
+                        (layout-frame frame width height)))))))))
 
 (defmethod compose-space ((pane top-level-sheet-pane) &key width height)
   (declare (ignore width height))
--- a/Core/clim-core/panes/composition.lisp
+++ b/Core/clim-core/panes/composition.lisp
@@ -190,6 +190,14 @@
   ()
   (:documentation "protocol class"))
 
+;; When a composite-pane receives the note-space-requirements-changed
+;; from one of its children, the normal behaviour is to change its
+;; space-requirements. For the moment exception are restraining-pane
+;; and viewport-pane. -- admich 2020-08-11
+(defmethod note-space-requirements-changed  ((pane composite-pane) client)
+  (declare (ignore client))
+  (change-space-requirements pane))
+
 (defmethod spacing-value-to-device-units ((pane extended-output-stream) x)
   (etypecase x
     (real x)
@@ -1152,72 +1160,33 @@
         (make-space-requirement))))
 
 (defmethod allocate-space ((pane viewport-pane) width height)
-  (with-slots (hscrollbar vscrollbar) (sheet-parent pane)
-    (let* ((child            (sheet-child pane))
-           (child-space      (compose-space child))
-           (child-width      (space-requirement-width child-space))
-           (child-min-width  (space-requirement-min-width child-space))
-           (child-height     (space-requirement-height child-space))
-           (child-min-height (space-requirement-min-height child-space)))
+  (let* ((parent           (sheet-parent pane))
+         (child            (sheet-child pane))
+         (child-space      (compose-space child))
+         (child-width      (space-requirement-width child-space))
+         (child-min-width  (space-requirement-min-width child-space))
+         (child-height     (space-requirement-height child-space))
+         (child-min-height (space-requirement-min-height child-space)))
+    (with-slots (hscrollbar vscrollbar) parent
       (move-and-resize-sheet child
                              (if hscrollbar (- (gadget-value hscrollbar)) 0)
                              (if vscrollbar (- (gadget-value vscrollbar)) 0)
                              (max child-width  width)
-                             (max child-height height))
+                             (max child-height height)))
                                         ; move-and-resize-sheet does not allocate space for the sheet...
                                         ; so we do it manually for this case, which may be wrong - CHECKME
                                         ; if this is the right place, reusing the above calculation might be a good idea
-      (allocate-space child
-                      (max child-min-width child-width  width)
-                      (max child-min-height child-height height)))))
+    (allocate-space child
+                    (max child-min-width child-width  width)
+                    (max child-min-height child-height height))
+    (scroller-pane/update-scroll-bars parent)))
 
 (defmethod note-input-focus-changed ((pane viewport-pane) state)
   (note-input-focus-changed (sheet-child pane) state))
 
-;;; This method ensures that when the child changes size, the viewport
-;;; will move its focus so that it will not display a region outside
-;;; of `child' (if at all possible, this ideal can be circumvented by
-;;; creating a child sheet that is smaller than the viewport). I do
-;;; not believe having a viewport look at "empty" space is ever
-;;; useful.
 (defmethod note-space-requirements-changed ((pane viewport-pane) child)
-  (let ((viewport-width (bounding-rectangle-width pane))
-        (viewport-height (bounding-rectangle-height pane))
-        (child-width (bounding-rectangle-width child))
-        (child-height (bounding-rectangle-height child)))
-    (destructuring-bind (horizontal-scroll vertical-scroll)
-        (mapcar #'- (multiple-value-list
-                     (transform-position (sheet-transformation child) 0 0)))
-      ;; XXX: We cannot use `scroll-extent', because McCLIM ignores it
-      ;; unless the scrollee happens to be drawing. Very weird, should
-      ;; be fixed.
-
-      ;; It's not a bug, it's a feature. This requires further thought. -Hefner
-      (move-sheet child
-                  (- (if (> (+ horizontal-scroll viewport-width)
-                            child-width)
-                         (- child-width viewport-width)
-                         horizontal-scroll))
-                  (- (if (> (+ vertical-scroll viewport-height)
-                            child-height)
-                         (- child-height viewport-height)
-                         vertical-scroll)))
-      (scroller-pane/update-scroll-bars (sheet-parent pane)))))
-
-;;; Accounting for changed space requirements
-
-(defmethod change-space-requirements ((pane viewport-pane) &rest options)
-  (declare (ignore options))
-  (let* ((client (sheet-child pane))
-         (sr (compose-space client))
-         (width  (max (bounding-rectangle-width pane)
-                      (space-requirement-width sr)))
-         (height (max (bounding-rectangle-height pane)
-                      (space-requirement-height sr))))
-    (resize-sheet client width height)
-    (allocate-space client width height)
-    (scroller-pane/update-scroll-bars (sheet-parent pane))))
-
+  (multiple-value-bind (width height) (bounding-rectangle-size pane)
+    (allocate-space pane width height)))
 
 ;;; SCROLLER PANE
 
--- a/Core/clim-core/panes/layout-protocol.lisp
+++ b/Core/clim-core/panes/layout-protocol.lisp
@@ -42,8 +42,7 @@
 ;;;
 ;;; NOTE-SPACE-REQUIREMENTS-CHANGED
 ;;;
-;;;   Called by CLIM when the space requirements of a pane have
-;;;   changed. Not called to layout a pane; This is only a kind of signal.
+;;;   Called by CLIM when the space requirements of a pane have changed.
 ;;;
 ;;; LAYOUT-FRAME
 ;;;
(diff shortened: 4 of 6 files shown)


dkochmanski:
While preparing for this review, I wrote this document:
https://gist.github.com/dkochmanski/d1dcacabcfa35886000301a8882f0b2c

Core/clim-core/panes/layout-protocol.lisp
```diff
@@ -473,13 +469,30 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
...
+        ;; the layout protocol calling LAYOUT-FRAME. The rationale of
+        ;; this is:
+        ;; 1. we can't call (LAYOUT-FRAME (PANE-FRAME pane)) on a
+        ;;   menu because with the actual implementation of menu it
+        ;;   will layout the main application and not the menu frame.
+        ;; 2. we automatically clear the cached values of
+        ;;    space-requirements for the involved panes.
+        ;; -- admich 2020-08-11
+        (if (top-level-sheet-pane-p pane)
+            (note-space-requirements-changed (sheet-parent pane) pane)
```
dkochmanski, line 483: top-levle-sheet's parent is the graft, is that right that we call the function on the graft? Isn't this supposed to be
(layout-frame frame)?
or even
```
(if (top-level-sheet-pane-p pane)
   (if resize-frame (layout-frame pane) (layout-frame pane width height))
   (note-space-requirements-changed parent pane))
```

I don't think that conditions here are right.

Core/clim-core/panes/layout-protocol.lisp
```diff
@@ -473,13 +469,30 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
 
 (defmethod change-space-requirements ((pane layout-protocol-mixin)
                                       &key resize-frame &allow-other-keys)
-  (when-let ((parent (sheet-parent pane)))
-    (change-space-requirements parent :resize-frame resize-frame)))
+  ;; do nothing here
+  nil)
 
 (defmethod change-space-requirements :after ((pane layout-protocol-mixin)
                                              &key resize-frame &allow-other-keys)
```
dkochmanski, line 466: this logic for the top-level-sheet-pane should be implemented in the top-level-sheet-pane *primary* method. Notice, that doing that in the after method violates the changing-space-requirements macro specification.

Core/clim-core/panes/layout-protocol.lisp
```diff
@@ -435,24 +435,20 @@
...
 ;;; know the details of the space requirement cache and the
 ;;; note-space-requirements-changed notifications.
 ;;;
-;;; The calls to change-space-requirements travel all the way up to
-;;; the top-level-sheet-pane which then invokes the layout protocol
-;;; calling layout-frame.
+;;; If :resize-frame t the calls to change-space-requirements travel
+;;; all the way up to the top-level-sheet-pane which then invokes the
+;;; layout protocol calling layout-frame.
 ;;;
```
dkochmanski, line 440: there is also a comment at the very top of the file which describes the layout protocol, maybe it should be amended too?

Core/clim-core/panes/composition.lisp
```diff
@@ -1180,55 +1188,16 @@
                                         ; if this is the right place, reusing the above calculation might be a good idea
       (allocate-space child
                       (max child-min-width child-width  width)
-                      (max child-min-height child-height height)))))
+                      (max child-min-height child-height height))
+      (scroller-pane/update-scroll-bars (sheet-parent pane)))))
```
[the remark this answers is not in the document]
admich: I do it

Core/clim-core/panes/composition.lisp
```diff
@@ -1180,55 +1188,16 @@
...
-                      (space-requirement-width sr)))
-         (height (max (bounding-rectangle-height pane)
-                      (space-requirement-height sr))))
-    (resize-sheet client width height)
-    (allocate-space client width height)
-    (scroller-pane/update-scroll-bars (sheet-parent pane))))
-
+  (allocate-space pane
+                  (bounding-rectangle-width pane)
+                  (bounding-rectangle-height pane)))
```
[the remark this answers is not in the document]
admich: I also think that this is right and your interpretation of allocate-space as small "layout-frame" for subtree (but without the compose-space step) seems right.

Core/clim-core/panes/layout-protocol.lisp
```diff
@@ -473,13 +469,30 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
...
+        ;; 1. we can't call (LAYOUT-FRAME (PANE-FRAME pane)) on a
+        ;;   menu because with the actual implementation of menu it
+        ;;   will layout the main application and not the menu frame.
+        ;; 2. we automatically clear the cached values of
+        ;;    space-requirements for the involved panes.
+        ;; -- admich 2020-08-11
+        (if (top-level-sheet-pane-p pane)
+            (note-space-requirements-changed (sheet-parent pane) pane)
+            (change-space-requirements parent :resize-frame t))
+        (note-space-requirements-changed (sheet-parent pane) pane))))
```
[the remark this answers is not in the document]
admich: I do it

Core/clim-core/panes/layout-protocol.lisp
admich: Note that this branch is executed when :resize-frame is t. It would be correct to simply call (layout-frame (pane-frame frame) here for all the pane, but in McCLIM for the menu frame (pane-frame frame) return the main application and not the menu. I explain this in the comment.

When change-space-requirements on a top-level-sheet it is correct to call note-space-requirements-changed because then the parent (the graft) ask space to the window manager.

The layout-frame function is called in the primary method of change-space-requirements on top-level-sheet.

Core/clim-core/panes/layout-protocol.lisp
admich: I don't understant but maybe is the same prolem of above.

Core/clim-core/panes/layout-protocol.lisp
admich: The comment at the top is almost right, I only remove the sentence that note-space-requirements-changed is only a kind of signal.

dkochmanski:
I've rebased your changes against master (the file frames.lisp has moved), see the branch admich-layout-protocol in this repository. Rebase was clean, so you may either pull that branch or rebase yourself. I'm going to review the branch again now.

Core/clim-core/panes/layout-protocol.lisp
```diff
@@ -520,13 +523,17 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
...
                              (layout-frame frame))
                             (t
-                             (layout-frame frame
-                                           (bounding-rectangle-width pane)
-                                           (bounding-rectangle-height pane))))))))))))
+                             (if (frame-resize-frame frame)
+                                 (layout-frame frame)
+                                 (layout-frame frame
+                                               (bounding-rectangle-width pane)
+                                               (bounding-rectangle-height pane)))))))))))))
```
dkochmanski, line 539: please use bounding-rectangle-size to gather both values at once.

Core/clim-core/panes/layout-protocol.lisp
```diff
@@ -466,20 +454,35 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
...
+        ;; the way up to the top-level-sheet-pane which then invokes
+        ;; the layout protocol calling LAYOUT-FRAME. The rationale of
+        ;; this is:
+        ;; 1. we can't call (LAYOUT-FRAME (PANE-FRAME pane)) on a
+        ;;   menu because with the actual implementation of menu it
+        ;;   will layout the main application and not the menu frame.
+        ;; 2. we automatically clear the cached values of
+        ;;    space-requirements for the involved panes.
+        ;; -- admich 2020-08-11
+        (if (top-level-sheet-pane-p pane)
```
dkochmanski, line 482: this function is forward-referenced from the file composition.lisp (it is defined after the first use)

Core/clim-core/panes/layout-protocol.lisp
dkochmanski: all calls to change-space-requirements which appear in a dynamic scope of `changing-space-requirements` should be simply recorded for later execution, while this after method triggers a traversal of a whole tree eagerly. Our implementation of `changing-space-requirements` is wrong either way, so I suppose that could be fixed /then/. It is not related to the previous question.

That said, please provide an example code that illustrates the issue with menus and explain what in the current implementation of menus makes it non-viable to call layout-frame directly (when resize-frame is given).

Core/clim-core/panes/layout-protocol.lisp
```diff
@@ -466,20 +454,35 @@ which changed during the current execution of CHANGING-SPACE-REQUIREMENTS.
...
+        ;; call directly LAYOUT-FRAME, we call
+        ;; CHANGE-SPACE-REQUIREMENTS on the parent and it travels all
+        ;; the way up to the top-level-sheet-pane which then invokes
+        ;; the layout protocol calling LAYOUT-FRAME. The rationale of
+        ;; this is:
+        ;; 1. we can't call (LAYOUT-FRAME (PANE-FRAME pane)) on a
+        ;;   menu because with the actual implementation of menu it
+        ;;   will layout the main application and not the menu frame.
+        ;; 2. we automatically clear the cached values of
+        ;;    space-requirements for the involved panes.
```
dkochmanski, line 480: do I understand correctly, that previously they were not invalidated and allocate-space was using obsolete results of compute-space?

assuming that we call layout-frame (i.e because the menu implementation is fixed to work better), how should that work? layout-protocol should invalidate all cached values at one go?  or should it be a responsibility of note-sheet-region-changed?

Core/clim-core/panes/layout-protocol.lisp
```diff
@@ -43,7 +43,7 @@
 ;;; NOTE-SPACE-REQUIREMENTS-CHANGED
 ;;;
 ;;;   Called by CLIM when the space requirements of a pane have
-;;;   changed. Not called to layout a pane; This is only a kind of signal.
+;;;   changed.
```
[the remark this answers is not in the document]
dkochmanski: also there is another comment in the same file which states:
`;;; - NOTE-SPACE-REQUIREMENTS-CHANGED is solely for the user.`

Core/clim-core/panes/composition.lisp
```diff
@@ -1163,72 +1173,34 @@
...
```
dkochmanski, line 1203: please use bounding-rectangle-size

Core/clim-core/panes/composition.lisp
```diff
@@ -201,6 +201,14 @@
...
   (:documentation "protocol class"))
 
+;; When a composite-pane receives the note-space-requirements-changed
+;; from one of its children, the normal behaviour is to change its
+;; space-requirements. For the moment exception are restraining-pane
+;; and viewport-pane. -- admich 2020-08-11
+(defmethod note-space-requirements-changed  ((pane composite-pane) client)
+  (declare (ignore client))
+  (change-space-requirements pane))
+
```
dkochmanski, line 211: I suppose that i.e vertical pane should call compose-space on the child and if its new space requirements may be fulfilled by the current size, then change-space-requirements doesn't need to be called. Is that right?

Core/clim-core/panes/layout-protocol.lisp
admich: It is the `layout-frame` call that its recorded for later execution.
I think that the most straight way will be something like:
```
(defmethod change-space-requirements :after ((pane layout-protocol-mixin)
                                             &key resize-frame &allow-other-keys)
  (when-let ((parent (sheet-parent pane)))
    (if resize-frame
        (if *changing-space-requirements*
            (unless (find pane *changed-space-requirements* :key #'second)
              (push (list (pane-frame pane) pane resize-frame)
                    *changed-space-requirements*))
            (layout-frame (pane-frame pane)))
        (note-space-requirements-changed parent pane))))
```
but we can't because `pane-frame` when its called on a menu pane (like Listener completion menu or presentation menu- right click-) return the main frame and not the menu-frame. So it will be the main frame and not the menu to be resized.

Core/clim-core/panes/layout-protocol.lisp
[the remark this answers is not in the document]
admich: done

Core/clim-core/frames.lisp
```diff
@@ -266,8 +268,14 @@ documentation produced by presentations.")
...
-      (generate-panes fm frame)
-      (layout-frame frame)
+      (let ((old-width (and (frame-top-level-sheet frame)
+                            (bounding-rectangle-width (frame-top-level-sheet frame))))
+            (old-height (and (frame-top-level-sheet frame)
+                             (bounding-rectangle-height (frame-top-level-sheet frame)))))
+        (generate-panes fm frame)
+        (if (and old-width old-height (not (frame-resize-frame frame)))
+            (layout-frame frame old-width old-height)
+            (layout-frame frame)))
```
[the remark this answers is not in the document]
admich: done

Core/clim-core/menu-choose.lisp
```diff
@@ -323,10 +323,10 @@ maximum size according to `frame')."
...
                                      ;; scroll bars).
                                      :width (+ menu-width
-                                               (- (pane-current-width scroller)
+                                               (- (bounding-rectangle-width scroller)
                                                   viewport-width))
                                      :height (+ menu-height
-                                                (- (pane-current-height scroller)
+                                                (- (bounding-rectangle-height scroller)
                                                    viewport-height))
                                      :resize-frame t))))
```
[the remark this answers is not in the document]
admich: done

Core/clim-core/panes/composition.lisp
```diff
@@ -286,9 +294,11 @@
            (cond (resize-frame
                   (layout-frame frame))
                  (t
-                  (layout-frame frame
-                                (bounding-rectangle-width pane)
-                                (bounding-rectangle-height pane))))))))
+                  (if (frame-resize-frame frame)
```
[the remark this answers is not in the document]
admich: done

Core/clim-core/panes/composition.lisp
```diff
@@ -286,9 +294,11 @@
...
+                      (layout-frame frame)
+                      (layout-frame frame
+                                    (bounding-rectangle-width pane)
+                                    (bounding-rectangle-height pane)))))))))
```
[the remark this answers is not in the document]
admich: done

Core/clim-core/panes/layout-protocol.lisp
admich: Yes, it's true. I don't have a solution, moving here the definition doesn't seem right. 

Maybe the right things it is to change the menu implementation before this PR and then use what I have written before:
```
(defmethod change-space-requirements :after ((pane layout-protocol-mixin)
                                             &key resize-frame &allow-other-keys)
  (when-let ((parent (sheet-parent pane)))
    (if resize-frame
        (if *changing-space-requirements*
            (unless (find pane *changed-space-requirements* :key #'second)
              (push (list (pane-frame pane) pane resize-frame)
                    *changed-space-requirements*))
            (layout-frame (pane-frame pane)))
        (note-space-requirements-changed parent pane))))
```

Core/clim-core/panes/layout-protocol.lisp
admich: We have to think about that, maybe it is necessary to change when cache the values and when invalidate them. I wrote something in the comment to your gist about layout-protocol https://gist.github.com/dkochmanski/d1dcacabcfa35886000301a8882f0b2c

Core/clim-core/panes/composition.lisp
admich: done

Core/clim-core/panes/composition.lisp
admich: I think you are right and that can be implemented.

Core/clim-core/panes/composition.lisp
```diff
@@ -1152,72 +1161,33 @@
...
-         (width  (max (bounding-rectangle-width pane)
-                      (space-requirement-width sr)))
-         (height (max (bounding-rectangle-height pane)
-                      (space-requirement-height sr))))
-    (resize-sheet client width height)
-    (allocate-space client width height)
-    (scroller-pane/update-scroll-bars (sheet-parent pane))))
-
+  (multiple-value-bind (width height) (bounding-rectangle-size pane)
+    (allocate-space pane width height)))
```
dkochmanski, line 1190: allocate-space is part of the layout-protocol, so if this function is invoked in the context of `changing-space-requirements`, this operation should be batched instead of being eagerly invoked.

Core/clim-core/frames/frames.lisp
```diff
@@ -121,7 +121,9 @@ input focus. This is a McCLIM extension."))
    (panes-for-layout :initform nil :accessor frame-panes-for-layout
                      :documentation "alist of names and panes
                                      (as returned by make-pane)")
-
+   (resize-frame :initarg :resize-frame
+                 :initform nil
+                 :accessor frame-resize-frame)
```
dkochmanski, line 126: this slot is taken into account only in the change-space-requirements specialization for the top-level-sheet-pane -- appropriate code should be added also to the specialization of unmanaged-top-level-sheet-pane

dkochmanski (changes requested):
I've pushed a few cosmetic fixes to the branch admich-layout-protocol in the McCLIM repository. I have two remarks for now:
- allocate-space should not be called when changing-space-requirements is in action
- resize-frame slot is not taken into account for unamanged-top-level-sheet-pane

I'll experiment a little with a solution, where unmanaged sheets have a correct frame set.

Core/clim-core/panes/composition.lisp
dkochmanski: but this is not a regression, I suppose it could be addressed when changing-space-requirements is improved

Core/clim-core/frames/frames.lisp
dkochmanski: taking that into account would require handing menu-frame correcty as a frame too, so that will wait as well.

Core/clim-core/panes/composition.lisp
dkochmanski: I'll treat it as a possible improvement.

Core/clim-core/menu-choose.lisp
```diff
@@ -310,7 +310,8 @@ maximum size according to `frame')."
     (when-let ((viewport (pane-viewport menu)))
       (multiple-value-bind (viewport-width viewport-height)
           (menu-size viewport *application-frame*)
-        (let ((scroller (pane-scroller menu)))
+        (multiple-value-bind (scroller-width scroller-height)
+            (bounding-rectangle-size (pane-scroller menu))
           (change-space-requirements scroller
```
dkochmanski, line 315: you still need to bind scroller for its used in the form below.

dkochmanski (changes requested):

dkochmanski:
I have merged the pull request manually (with few corrections). thanks!

admich:
ok, great , thanks

PR #1051, opened 2020-08-17
Fix and reactivate stopwatch demo

scymtym:
* Slightly prettier output

* Re-center output after frame geometry changes

* Update every 10 ms instead of every 100 ms

* Cosmetic changes


dkochmanski:
lgtm, thanks

PR #1053, opened 2020-08-25
Better keyword checking and processing in table formatting

scymtym:
* All `formatting-*` macros are specified to accept a `record-type` keyword argument and to allow other keyword arguments. This is almost certainly intended as a mechanism for passing initargs in case non-default record types are used. To make this mechanism work consistently across all macros:

  * Use `apply` with `invoke-with-new-output-record` and the passed initargs in all `invoke-with-*` functions

  * Use `&allow-other-keys` in all macros.

  * Use `gen-invoke-trampoline` in all macro.

* Check initargs which are passed as keyword argument in `formatting-*` at macroexpansion-time when possible.

### DIFF
--- a/Core/clim-core/table-formatting.lisp
+++ b/Core/clim-core/table-formatting.lisp
@@ -66,38 +66,41 @@
   (:default-initargs
    :align-x :left :align-y :baseline :min-width 0 :min-height 0))
 
-(defgeneric invoke-formatting-cell
-    (stream cont &key align-x align-y min-width min-height record-type &allow-other-keys)
-  (:method (stream continuation &key
-                                  (align-x :left)
-                                  (align-y :baseline)
-                                  (min-width 0)
-                                  (min-height 0)
-                                  (record-type 'standard-cell-output-record)
-                                  &allow-other-keys)
-    (invoke-with-new-output-record
-     stream
-     (lambda (stream record)
-       (declare (ignore record))
-       (with-temporary-margins (stream :left '(:absolute 0))
-         (letf (((stream-cursor-position stream) (values 0 0)))
-           (funcall continuation stream))))
-     record-type
-     :align-x align-x
-     :align-y align-y
-     :min-width (parse-space stream min-width :horizontal)
-     :min-height (parse-space stream min-height :vertical))))
+(defun invoke-formatting-cell
+    (stream continuation &rest initargs
+                         &key (record-type 'standard-cell-output-record)
+                              (align-x :left)
+                              (align-y :baseline)
+                              (min-width 0)
+                              (min-height 0)
+                         &allow-other-keys)
+  (apply #'invoke-with-new-output-record
+         stream
+         (lambda (stream record)
+           (declare (ignore record))
+           (with-temporary-margins (stream :left '(:absolute 0))
+             (letf (((stream-cursor-position stream) (values 0 0)))
+               (funcall continuation stream))))
+         record-type
+         :align-x align-x
+         :align-y align-y
+         :min-width (parse-space stream min-width :horizontal)
+         :min-height (parse-space stream min-height :vertical)
+         (alexandria:remove-from-plist
+          initargs :record-type :align-x :align-y :min-width :min-height)))
 
 (defmacro formatting-cell ((&optional (stream t)
                             &rest args
-                            &key align-x align-y min-width min-height record-type)
+                            &key (record-type ''standard-cell-output-record)
+                                 align-x align-y min-width min-height
+                            &allow-other-keys)
                            &body body)
-  (declare (ignore align-x align-y min-width min-height record-type))
-  (setq stream (stream-designator-symbol stream '*standard-output*))
-  (with-gensyms (continuation)
-    `(flet ((,continuation (,stream) ,@body))
-       (declare (dynamic-extent #',continuation))
-       (invoke-formatting-cell ,stream #',continuation ,@args))))
+  (declare (ignore record-type align-x align-y min-width min-height))
+  (gen-invoke-trampoline
+   'invoke-formatting-cell
+   (list (stream-designator-symbol stream '*standard-output*))
+   args
+   body))
 
 
 ;;; Generic block formatting
@@ -155,24 +158,27 @@ to a table cell within the row."))
                                (row-record standard-row-output-record))
   (map-over-block-cells function row-record))
 
-(defmacro formatting-row ((&optional (stream t)
-                                     &rest more
-                                     &key (record-type ''standard-row-output-record))
-                          &body body)
-  (setf stream (stream-designator-symbol stream '*standard-output*))
-  (with-gensyms (record)
-    (with-keywords-removed (more (:record-type))
-      `(with-new-output-record (,stream ,record-type ,record ,@more)
-         ,@body))))
-
-(defgeneric invoke-formatting-row (stream cont record-type &rest initargs))
-
-(defmethod invoke-formatting-row (stream cont record-type &rest initargs)
+(defun invoke-formatting-row
+    (stream continuation &rest initargs
+                         &key (record-type 'standard-row-output-record)
+                         &allow-other-keys)
   (apply #'invoke-with-new-output-record stream
-         (lambda (s r)
-           (declare (ignore r))
-           (funcall cont s))
-         record-type initargs))
+         (lambda (stream record)
+           (declare (ignore record))
+           (funcall continuation stream))
+         record-type (alexandria:remove-from-plist initargs :record-type)))
+
+(defmacro formatting-row ((&optional (stream t)
+                           &rest args
+                           &key (record-type ''standard-row-output-record)
+                           &allow-other-keys)
+                          &body body)
+  (declare (ignore record-type))
+  (gen-invoke-trampoline
+   'invoke-formatting-row
+   (list (stream-designator-symbol stream '*standard-output*))
+   args
+   body))
 
 
 ;;; Column formatting
@@ -192,22 +198,27 @@ corresponding to a table cell within the column."))
 (defmethod map-over-column-cells (function (column-record standard-column-output-record))
   (map-over-block-cells function column-record))
 
+(defun invoke-formatting-column
+    (stream continuation &rest initargs
+                         &key (record-type 'standard-column-output-record)
+                         &allow-other-keys)
+  (apply #'invoke-with-new-output-record stream
+         (lambda (stream record)
+           (declare (ignore record))
+           (funcall continuation stream))
+         record-type (alexandria:remove-from-plist initargs :record-type)))
+
 (defmacro formatting-column ((&optional (stream t)
-                                        &rest more
-                                        &key (record-type
-                                              ''standard-column-output-record))
+                              &rest args
+                              &key (record-type ''standard-column-output-record)
+                              &allow-other-keys)
                              &body body)
-  (setf stream (stream-designator-symbol stream '*standard-output*))
-  (with-gensyms (record)
-    (with-keywords-removed (more (:record-type))
-      `(with-new-output-record (,stream ,record-type ,record ,@more)
-         ,@body))))
-
-(defgeneric invoke-formatting-column (stream cont record-type &rest initargs))
-
-(defmethod invoke-formatting-column (stream cont record-type &rest initargs)
-
-  (apply #'invoke-with-new-output-record stream (lambda (s r) r (funcall cont s)) record-type initargs))
+  (declare (ignore record-type))
+  (gen-invoke-trampoline
+   'invoke-formatting-column
+   (list (stream-designator-symbol stream '*standard-output*))
+   args
+   body))
 
 
 ;;; Table formatting
@@ -233,7 +244,11 @@ skips intervening non-table output record structures."))
    ;; adjust-multiple-columns
    (widths)
    (heights)                            ;needed?
-   (rows)))
+   (rows))
+  (:default-initargs
+   :multiple-columns nil
+   :multiple-columns-x-spacing nil
+   :equalize-column-widths nil))
 
 (defmethod initialize-instance :after ((table standard-table-output-record)
                                        &rest initargs)
@@ -259,33 +274,15 @@ skips intervening non-table output record structures."))
                 (replay-output-record record stream region x-offset y-offset))
             (nreverse other-records)))))
 
-(defmacro formatting-table ((&optional (stream t)
-                                       &rest args
-                                       &key x-spacing y-spacing
-                                       multiple-columns
-                                       multiple-columns-x-spacing
-                                       equalize-column-widths (move-cursor t)
-                                       (record-type ''standard-table-output-record)
-                                       &allow-other-keys)
-                            &body body)
-  (declare (ignore x-spacing y-spacing multiple-columns
-                   multiple-columns-x-spacing
-                   equalize-column-widths move-cursor record-type))
-  (gen-invoke-trampoline 'invoke-formatting-table
-                         (list (stream-designator-symbol stream
-                                                         '*standard-output*))
-                         args
-                         body))
-
 (defun invoke-formatting-table
-    (stream continuation
-     &key x-spacing y-spacing
-       multiple-columns
-       multiple-columns-x-spacing
-       equalize-column-widths
-       (move-cursor t)
-       (record-type 'standard-table-output-record)
-       &allow-other-keys)
+    (stream continuation &rest initargs
+                         &key (record-type 'standard-table-output-record)
+                              (move-cursor t)
+                              x-spacing y-spacing
+                              multiple-columns
+                              multiple-columns-x-spacing
+                              equalize-column-widths
+                         &allow-other-keys)
   (setq x-spacing (parse-space stream (or x-spacing #\Space) :horizontal))
   (setq y-spacing (parse-space stream (or y-spacing
                                           (stream-vertical-spacing stream))
@@ -294,31 +291,54 @@ skips intervening non-table output record structures."))
         (if multiple-columns-x-spacing
             (parse-space stream multiple-columns-x-spacing :horizontal)
             x-spacing))
-  (with-new-output-record
-      (stream record-type table
-              :x-spacing x-spacing
-              :y-spacing y-spacing
-              :multiple-columns multiple-columns
-              :multiple-columns-x-spacing multiple-columns-x-spacing
-              :equalize-column-widths equalize-column-widths)
-    (multiple-value-bind (cursor-old-x cursor-old-y)
-        (stream-cursor-position stream)
-      (with-output-recording-options (stream :record t :draw nil)
-        (funcall continuation stream)
-        (force-output stream))
-      (setf (stream-cursor-position stream)
-            (values cursor-old-x cursor-old-y))
-      (with-output-recording-options (stream :record nil :draw nil)
-        (adjust-table-cells table stream)
-        (when multiple-columns (adjust-multiple-columns table stream))
-        (tree-recompute-extent table))
-      (replay table stream)
-      (if move-cursor
-          (setf (stream-cursor-position stream)
-                (values (bounding-rectangle-max-x table)
-                        (bounding-rectangle-max-y table)))
-          (setf (stream-cursor-position stream)
-                (values cursor-old-x cursor-old-y))))))
+  (apply #'invoke-with-new-output-record stream
+         (lambda (stream table)
+           (multiple-value-bind (cursor-old-x cursor-old-y)
+               (stream-cursor-position stream)
+             (with-output-recording-options (stream :record t :draw nil)
+               (funcall continuation stream)
+               (force-output stream))
+             (setf (stream-cursor-position stream)
+                   (values cursor-old-x cursor-old-y))
+             (with-output-recording-options (stream :record nil :draw nil)
+               (adjust-table-cells table stream)
+               (when multiple-columns (adjust-multiple-columns table stream))
+               (tree-recompute-extent table))
+             (replay table stream)
+             (setf (stream-cursor-position stream)
+                   (if move-cursor
+                       (values (bounding-rectangle-max-x table)
+                               (bounding-rectangle-max-y table))
+                       (values cursor-old-x cursor-old-y)))))
+         record-type
+         :x-spacing x-spacing
+         :y-spacing y-spacing
+         :multiple-columns multiple-columns
+         :multiple-columns-x-spacing multiple-columns-x-spacing
+         :equalize-column-widths equalize-column-widths
+         (alexandria:remove-from-plist
+          initargs :record-type :move-cursor
+                   :x-spacing :y-spacing :multiple-columns
+                   :multiple-columns-x-spacing :equalize-column-widths)))
+
+(defmacro formatting-table ((&optional (stream t)
+                             &rest args
+                             &key (record-type ''standard-table-output-record)
+                                  (move-cursor t)
+                                  x-spacing y-spacing
+                                  multiple-columns
+                                  multiple-columns-x-spacing
+                                  equalize-column-widths
+                             &allow-other-keys)
+                            &body body)
+  (declare (ignore record-type move-cursor
+                   x-spacing y-spacing multiple-columns
+                   multiple-columns-x-spacing equalize-column-widths))
+  (gen-invoke-trampoline
+   'invoke-formatting-table
+   (list (stream-designator-symbol stream '*standard-output*))
+   args
+   body))
 
 ;;; Think about rewriting this using a common superclass for row and
 ;;; column records.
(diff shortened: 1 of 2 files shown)


Core/clim-core/table-formatting.lisp
```diff
@@ -54,6 +54,22 @@
...
+    (when (and (constantp record-type)
+               (eq (eval record-type) standard-record-type))
+      (loop for (key value) on arguments by #'cddr
+            when (constantp key)
+            do (let ((key (eval key)))
+                 (unless (or (eq key :record-type)
+                             (member key valid-keywords))
+                   (with-current-source-form (arguments)
+                     (error "~@<~S is not a valid initarg for ~S.~@:>"
+                            key standard-record-type))))))))
```
dkochmanski, line 71: since we are doing this, maybe it is worth also verifying, whether the record type has expected protocol class?

Core/clim-core/table-formatting.lisp
```diff
@@ -66,38 +82,57 @@
...
-  (declare (ignore align-x align-y min-width min-height record-type))
-  (setq stream (stream-designator-symbol stream '*standard-output*))
-  (with-gensyms (continuation)
-    `(flet ((,continuation (,stream) ,@body))
-       (declare (dynamic-extent #',continuation))
-       (invoke-formatting-cell ,stream #',continuation ,@args))))
+  (declare (ignore . #.+standard-cell-output-record-initargs+))
+  ;; If RECORD-TYPE is STANDARD-CELL-OUTPUT-RECORD, make sure that all
+  ;; keyword arguments with constant keys correspond to known
+  ;; initargs.
```
dkochmanski, line 126: wrap the last line.

Core/clim-core/table-formatting.lisp
```diff
@@ -66,38 +82,57 @@
   (:default-initargs
    :align-x :left :align-y :baseline :min-width 0 :min-height 0))
 
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (alexandria:define-constant +standard-cell-output-record-initargs+
+      '(align-x align-y min-width min-height)
+    :test #'equal))
```
dkochmanski, line 88: these initargs are initargs of the protocol class `cell-output-record`, not the `standard-cell-output-record`, please rename the constant.

Core/clim-core/table-formatting.lisp
```diff
@@ -66,38 +82,57 @@
...
-       (declare (ignore record))
-       (with-temporary-margins (stream :left '(:absolute 0))
-         (letf (((stream-cursor-position stream) (values 0 0)))
-           (funcall continuation stream))))
-     record-type
-     :align-x align-x
-     :align-y align-y
-     :min-width (parse-space stream min-width :horizontal)
-     :min-height (parse-space stream min-height :vertical))))
+    (stream continuation
```
dkochmanski, line 91: please move generic functions invoke-* to decls.lisp

add them under 17.2 with a comment, that they are not part of the specification. initargs should be manually mentioned there.

Core/clim-core/table-formatting.lisp
dkochmanski: I gather that these are not keywords in order to include them in defgeneric as arguments - please define initargs as keywords and use the constant without #.(map 'list …) later.

Mention arguments in functions and methods "manually".

Core/clim-core/table-formatting.lisp
```diff
@@ -66,38 +82,57 @@
...
+               (letf (((stream-cursor-position stream) (values 0 0)))
+                 (funcall continuation stream))))
+           record-type
+           :align-x align-x
+           :align-y align-y
+           :min-width (parse-space stream min-width :horizontal)
+           :min-height (parse-space stream min-height :vertical)
+           (alexandria:remove-from-plist
+            initargs :record-type . #.(map 'list #'alexandria:make-keyword
+                                           +standard-cell-output-record-initargs+)))))
```
dkochmanski, line 115: ditto, use a list of keywords in the constant, so there is no need to call make-keyword.

Core/clim-core/table-formatting.lisp
```diff
@@ -66,38 +82,57 @@
...
 
 (defmacro formatting-cell ((&optional (stream t)
                             &rest args
-                            &key align-x align-y min-width min-height record-type)
+                            &key (record-type ''standard-cell-output-record)
+                                 . #.`(,@+standard-cell-output-record-initargs+
+                                       &allow-other-keys))
```
dkochmanski, line 121: ditto, just list arguments in plain text.

Core/clim-core/table-formatting.lisp
```diff
@@ -155,24 +190,32 @@ to a table cell within the row."))
                                (row-record standard-row-output-record))
   (map-over-block-cells function row-record))
 
+(defgeneric invoke-formatting-row (stream continuation
+                                   &rest initargs
+                                   &key record-type &allow-other-keys)
```
dkochmanski, line 195: move to decls

Core/clim-core/table-formatting.lisp
```diff
@@ -66,38 +82,57 @@
...
-       (declare (dynamic-extent #',continuation))
-       (invoke-formatting-cell ,stream #',continuation ,@args))))
+  (declare (ignore . #.+standard-cell-output-record-initargs+))
+  ;; If RECORD-TYPE is STANDARD-CELL-OUTPUT-RECORD, make sure that all
+  ;; keyword arguments with constant keys correspond to known
+  ;; initargs.
+  (maybe-check-constant-keyword-arguments
+   record-type 'standard-cell-output-record
+   args '#.(map 'list #'alexandria:make-keyword
+                +standard-cell-output-record-initargs+))
```
dkochmanski, line 130: use a list of keywords without read-time sorcery.

Core/clim-core/table-formatting.lisp
```diff
@@ -259,33 +312,18 @@ skips intervening non-table output record structures."))
...
-  (gen-invoke-trampoline 'invoke-formatting-table
-                         (list (stream-designator-symbol stream
-                                                         '*standard-output*))
-                         args
-                         body))
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (alexandria:define-constant +standard-table-output-record-initargs+
+      '(x-spacing y-spacing multiple-columns
+        multiple-columns-x-spacing equalize-column-widths)
+    :test #'equal))
```
dkochmanski, line 319: these are initargs for table-output-record, so rename the constant.

n.b multiple-columns is not mentioned in the spec, howerver it seems to be used and is included in the Franz CLIM guide:

> multiple-columns Either nil,t, or an integer. If it istor an integer, the table rows are broken up into multiple columns. If it ist, CLIM will determine the optimal number of columns. If it is an integer, it will be interpreted as the desired number of columns.

since you are at it, please add taht entry to the spec as table-output-record initarg. use keywords, not internal symbols (ditto).

Core/clim-core/table-formatting.lisp
```diff
@@ -259,33 +312,18 @@ skips intervening non-table output record structures."))
...
-       multiple-columns-x-spacing
-       equalize-column-widths
-       (move-cursor t)
-       (record-type 'standard-table-output-record)
-       &allow-other-keys)
+    (stream continuation &rest initargs
+                         &key (record-type 'standard-table-output-record)
+                              (move-cursor t)
+                              . #.`(,@+standard-table-output-record-initargs+
+                                    &allow-other-keys))
```
dkochmanski, line 326: mention all arguments manually and for consistency define as a generic function (and add it to decls)

Core/clim-core/table-formatting.lisp
```diff
@@ -294,31 +332,55 @@ skips intervening non-table output record structures."))
...
+                       (values cursor-old-x cursor-old-y)))))
+         record-type
+         :x-spacing x-spacing
+         :y-spacing y-spacing
+         :multiple-columns multiple-columns
+         :multiple-columns-x-spacing multiple-columns-x-spacing
+         :equalize-column-widths equalize-column-widths
+         (alexandria:remove-from-plist
+          initargs :record-type :x-spacing :y-spacing :multiple-columns
+                   :multiple-columns-x-spacing :equalize-column-widths)))
```
dkochmanski, line 362: use the constant list of initargs (which are keywords).

*** 

Instead of removing intiargs, this should be

(with-keywords-removed (initargs (:record-type))
  (apply stream lambda record-type initargs))

because otherwise when some other method verifies, whether the argument was supplied (for whatever reason), it will always conclude that yes, it was.

Core/clim-core/table-formatting.lisp
```diff
@@ -294,31 +332,55 @@ skips intervening non-table output record structures."))
...
+                                        &allow-other-keys))
+                            &body body)
+  (declare (ignore move-cursor . #.+standard-table-output-record-initargs+))
+  ;; If RECORD-TYPE is STANDARD-TABLE-OUTPUT-RECORD, make sure that
+  ;; all keyword arguments with constant keys correspond to known
+  ;; initargs.
+  (maybe-check-constant-keyword-arguments
+   record-type 'standard-table-output-record
+   args '(:move-cursor . #.(map 'list #'alexandria:make-keyword
+                            +standard-table-output-record-initargs+)))
```
dkochmanski, line 378: use keywords in the list so you don't have to map at read time but simply use the constant.

Core/clim-core/table-formatting.lisp
```diff
@@ -373,60 +435,69 @@ skips intervening non-table output record structures."))
...
-                                &key x-spacing y-spacing n-columns n-rows
-                                  max-width max-height
-                                  initial-spacing (row-wise t) (move-cursor t)
-                                  (record-type
-                                   'standard-item-list-output-record))
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (alexandria:define-constant +standard-item-list-output-record-initargs+
+      '(x-spacing y-spacing n-columns n-rows max-width max-height
+        initial-spacing row-wise stream-width stream-height)
+    :test #'equal))
```
dkochmanski, line 453: these are item-list-output-record initiargs, which is a protocol class - rename a constant. also please use keywords, not internal symbols.

Core/clim-core/table-formatting.lisp
```diff
@@ -373,60 +435,69 @@ skips intervening non-table output record structures."))
...
+
+(defun invoke-format-item-list
```
dkochmanski, line 455: for consistency make this a generic function and declare in decls.

Core/clim-core/table-formatting.lisp
```diff
@@ -373,60 +435,69 @@ skips intervening non-table output record structures."))
...
+         :y-spacing y-spacing
+         :n-columns n-columns
+         :n-rows n-rows
+         :max-width max-width
+         :max-height max-height
+         :initial-spacing initial-spacing
+         :row-wise row-wise
+         (alexandria:remove-from-plist
+          initargs :record-type :x-spacing :y-spacing :n-columns :n-rows
+                   :max-width :max-height :initial-spacing :row-wise)))
```
dkochmanski, line 494: Instead of removing intiargs, this should be

(with-keywords-removed (initargs (:record-type))
  (apply stream lambda record-type initargs))

because otherwise when some other method verifies, whether the argument was supplied (for whatever reason), it will always conclude that yes, it was.

Core/clim-core/table-formatting.lisp
```diff
@@ -67,37 +67,42 @@
...
+             (with-temporary-margins (stream :left '(:absolute 0))
+               (letf (((stream-cursor-position stream) (values 0 0)))
+                 (funcall continuation stream))))
+           record-type
+           :align-x align-x
+           :align-y align-y
+           :min-width (parse-space stream min-width :horizontal)
+           :min-height (parse-space stream min-height :vertical)
+           (alexandria:remove-from-plist
+            initargs :record-type :align-x :align-y :min-width :min-height))))
```
dkochmanski: Instead of removing intiargs, this should be
```
(with-keywords-removed (initargs (:record-type))
  (apply stream lambda record-type initargs))
```

because otherwise when some other method verifies, whether the argument was supplied (for whatever reason), it will always conclude that yes, it was.

Core/clim-core/table-formatting.lisp
```diff
@@ -155,24 +160,28 @@ to a table cell within the row."))
...
+(defgeneric invoke-formatting-row (stream continuation
+                                   &rest initargs
+                                   &key record-type &allow-other-keys)
+  (:method (stream continuation &rest initargs
+                                &key (record-type 'standard-row-output-record))
+    (apply #'invoke-with-new-output-record stream
+           (lambda (stream record)
+             (declare (ignore record))
+             (funcall continuation stream))
+           record-type (alexandria:remove-from-plist initargs :record-type))))
```
dkochmanski, line 202: with-keyword-removed

Core/clim-core/table-formatting.lisp
```diff
@@ -192,22 +201,28 @@ corresponding to a table cell within the column."))
...
+(defgeneric invoke-formatting-column (stream continuation
+                                      &rest initargs
+                                      &key record-type &allow-other-keys)
+  (:method (stream continuation &rest initargs
+                                &key (record-type 'standard-column-output-record))
+    (apply #'invoke-with-new-output-record stream
+           (lambda (stream record)
+             (declare (ignore record))
+             (funcall continuation stream))
+           record-type (alexandria:remove-from-plist initargs :record-type))))
```
dkochmanski, line 247: with-keywords-removed

dkochmanski (changes requested):
First pass of the review. Remarks boil down to:
- when you have constants with initargs, use keywords
- write lambda lists manually (without unsplicing at read time)
- don't pass unsupplied initargs to apply
- use with-keywords-removed
- put defgenerics for standard macros in decls.lisp
- make all invoke-* functions generic

Core/clim-core/table-formatting.lisp
scymtym: That seems out of scope for a macro. Somebody might use the macro with a custom record type and define that record type class after using the macro.

Core/clim-core/table-formatting.lisp
scymtym: You mean after changing this to a generic function? This code would be the default method, right?

I assumed the argument processing was done to inject default values for unsupplied keywords which would work. But in the function lambda list, all keyword parameters default to `nil` if unsupplied, so I'm not sure what the intention was. Given that, I tend to agree that just leaving everything but `record-type` alone is probably best.

Core/clim-core/table-formatting.lisp
scymtym: See previous comment. But in this case I would argue that the intention was to default `row-wise` to `t`. Leaving the others alone seems like the right thing to do.

Core/clim-core/table-formatting.lisp
scymtym: See previous comments. But in this case, I don't see how the initargs could just be applied unmodified. That would forego the defaulting as well as the `parse-space` calls.

Core/clim-core/table-formatting.lisp
scymtym: Instead of `remove-from-plist`?

Core/clim-core/table-formatting.lisp
scymtym: The keyword is mentioned in the description of `formatting-table` (with a similar description to the one you quote) so it seems just the initarg is missing.

dkochmanski:
re macros, I was thinking about something in this spirit (mind, that move-cursor is not handled at all in this snippet):

```lisp
(defmacro define-formatting (name class args
                                   &body body)
  (let* ((fname (alexandria:symbolicate 'invoke- name))
         (keywords (map 'list
                        (alexandria:compose #'alexandria:make-keyword
                                            #'alexandria:ensure-car)
                        args)))
    `(progn
       (defmacro ,name ((&optional (stream t)
                         &rest args
                         &key (record-type '',class) ,@args
                         &allow-other-keys)
                        &body body)
         (declare (ignore ,@args))
         (maybe-check-constant-keyword-arguments record-type ',class
                                                 args ',keywords)
         (gen-invoke-trampoline
          ',fname (list (stream-designator-symbol stream '*standard-output*))
          args body))
       (defmethod ,fname (stream continuation
                          &rest initargs
                          &key (record-type ',class) ,@args
                          &allow-other-keys)
         (declare (ignorable ,@args))
         (with-keywords-removed (initargs (:record-type))
           ,@body)))))

(define-formatting formatting-cell standard-cell-output-record
    ((align-x :left) (align-y :baseline) (min-width 0) (min-height 0))
  (with-keywords-removed (initargs (:align-x :align-y :min-width :min-height))
    (apply #'invoke-with-new-output-record
           stream
           (lambda (stream record)
             (declare (ignore record))
             (with-temporary-margins (stream :left '(:absolute 0))
               (letf (((stream-cursor-position stream) (values 0 0)))
                 (funcall continuation stream))))
           record-type
           :align-x align-x
           :align-y align-y
           :min-width (parse-space stream min-width :horizontal)
           :min-height (parse-space stream min-height :vertical)
           initargs)))

(define-formatting formatting-row standard-row-output-record
    ()
  (apply #'invoke-with-new-output-record stream
         (lambda (stream record)
           (declare (ignore record))
           (funcall continuation stream))
         record-type
         initargs))

;;; define-formatting in this case have also argument move-cursor, so either
;;; define a separate macro, or add the conditional part to define-formatting.
(define-formatting formatting-table standard-table-output-record
    (x-spacing y-spacing multiple-columns
     multiple-columns-x-spacing equalize-column-widths)
  (setq x-spacing (parse-space stream (or x-spacing #\Space) :horizontal))
  (setq y-spacing (parse-space stream (or y-spacing
                                          (stream-vertical-spacing stream))
                               :vertical))
  (setq multiple-columns-x-spacing
        (if multiple-columns-x-spacing
            (parse-space stream multiple-columns-x-spacing :horizontal)
            x-spacing))
  (apply #'invoke-with-new-output-record stream
         (lambda (stream table)
           (multiple-value-bind (cursor-old-x cursor-old-y)
               (stream-cursor-position stream)
             (with-output-recording-options (stream :record t :draw nil)
               (funcall continuation stream)
               (force-output stream))
             (setf (stream-cursor-position stream)
                   (values cursor-old-x cursor-old-y))
             (with-output-recording-options (stream :record nil :draw nil)
               (adjust-table-cells table stream)
               (when multiple-columns (adjust-multiple-columns table stream))
               (tree-recompute-extent table))
             (replay table stream)
             (setf (stream-cursor-position stream)
                   (if move-cursor
                       (values (bounding-rectangle-max-x table)
                               (bounding-rectangle-max-y table))
                       (values cursor-old-x cursor-old-y)))))
         record-type
         :x-spacing x-spacing
         :y-spacing y-spacing
         :multiple-columns multiple-columns
         :multiple-columns-x-spacing multiple-columns-x-spacing
         :equalize-column-widths equalize-column-widths
         initargs))
```

scymtym:
Haha, I was thinking the same but thought it would be too extreme for your taste :)

Core/clim-core/table-formatting.lisp
```diff
@@ -54,6 +54,94 @@
...
+                   &rest initargs
+                   &key (record-type ',standard-output-record-class)
+                        ,@keyword-arguments/defaults)
+           (apply #'invoke-with-new-output-record
+                  stream
+                  (lambda (,stream-var ,record-var) ,@body)
+                  record-type
+                  ,@transformed-keyword-arguments
+                  (alexandria:remove-from-plist
+                   initargs :record-type ,@remove-keyword-arguments))))
```
dkochmanski, line 126: use with-keywords-removed

Core/clim-core/table-formatting.lisp
```diff
@@ -54,6 +54,94 @@
...
+            do (let ((key (eval key)))
+                 (unless (or (eq key :record-type)
+                             (member key valid-keywords))
+                   (with-current-source-form (arguments)
+                     (error "~@<~S is not a valid initarg for ~S.~@:>"
+                            key standard-record-type))))))))
+
+;;; Helper macro for defining FORMATTING-* macros in the corresponding
+;;; INVOKE-FORMATTING-* functions.
+
```
dkochmanski, line 77: explain in the comment binding optinons (bind, default, transform and passp)

Core/clim-core/table-formatting.lisp
```diff
@@ -54,6 +54,94 @@
...
+(defmacro define-formatting-macro ((name standard-output-record-class)
+                                   initargs
```
dkochmanski, line 79: name this variable either macro-args, or simply args. all are initargs.

dkochmanski (changes requested):
Spec change is OK to me. Few remarks regarding macro included. The macro interface is more complicated than I would wish for, but it is similar in spirit to def-grecording, so I have no objections.

Core/clim-core/table-formatting.lisp
dkochmanski: when value needs processing, it needs to be passed explicitly - yes. for other cases, the default method should have its own defaults (and macro can have these defaults listed for a sake of editor hints when writing the macro).

Core/clim-core/table-formatting.lisp
dkochmanski: I've answered in the previous comment, closing this thread.

Core/clim-core/table-formatting.lisp
dkochmanski: ditto, closing

Core/clim-core/table-formatting.lisp
dkochmanski: yes, closing (because this remark is in the new review as well)

Core/clim-core/table-formatting.lisp
dkochmanski: mentioned in the later review.

Core/clim-core/table-formatting.lisp
dkochmanski: mentioned later.

Core/clim-core/table-formatting.lisp
```diff
@@ -54,6 +54,94 @@
...
+                  (alexandria:appendf transformed-keyword-arguments
+                                      `(,keyword ,(if transformp
+                                                      transform
+                                                      name))))
+                (when (or defaultp transformp (not passp))
+                  (alexandria:appendf remove-keyword-arguments (list keyword))))))
+          initargs)
+    `(progn
+       (defgeneric ,invoke-name
+           (stream continuation &key record-type ,@keyword-arguments
```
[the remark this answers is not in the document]
scymtym: I'm fine with ordinary functions. The only halfway sensible specialization would be to the stream class, but I can't imagine any uses cases for that.

Core/clim-core/table-formatting.lisp
```diff
@@ -373,101 +390,77 @@ skips intervening non-table output record structures."))
...
+           stream
+           (lambda (stream)
+             (map nil (lambda (item)
+                        (formatting-cell (stream :align-x cell-align-x
+                                                 :align-y cell-align-y)
+                          (funcall printer item stream)))
+                  items))
+           (alexandria:remove-from-plist
+            args :stream :printer :presentation-type
+                 :cell-align-x :cell-align-y))))
```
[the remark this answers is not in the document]
scymtym: Why? I think `remove-from-plist` is appropriate since the modified plist is not referenced in other places. And even if it were, `let` with `remove-from-plist` (or our own `remove-keywords`) seems more flexible and clearer to me than a specialized binding macro.

Core/clim-core/table-formatting.lisp
```diff
@@ -54,6 +54,94 @@
 
 (in-package :clim-internals)
 
+;;; Macro argument checking
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  ;; FORMAT-ITEMS uses FORMATTING-CELL, so we need this function at
+  ;; compile-time.
+  (defun maybe-check-constant-keyword-arguments
+      (record-type standard-record-type arguments valid-keywords)
```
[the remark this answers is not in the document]
scymtym: How about renaming it to `%maybe-check-constant-keyword-arguments`? It would rather not compile and keep around the same function five times.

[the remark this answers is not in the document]
dkochmanski: it may stay as it is in that case. I don't have a strong opinion. The current name is longish though.

Core/clim-core/table-formatting.lisp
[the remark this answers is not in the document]
dkochmanski: with-keywords-removed is our internal macro to do exactly that, I don't see a reason why not to use it. That said, if you really have a strong desire to use remove-from-plist anyway, then bind args in a let over apply.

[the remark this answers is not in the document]
scymtym: >  That said, if you really have a strong desire to use remove-from-plist anyway

Thanks. Long-term, I would like to remove our `remove-keywords` and `with-keywords-removed` and switch to the alexandria versions. Just like we do with `minf`, `lerp`, etc.

> then bind args in a let over apply.

You mean binding the intermediate value instead of computing it within the `apply` form?

Core/clim-core/table-formatting.lisp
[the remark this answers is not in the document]
scymtym: Changed.

Core/clim-core/table-formatting.lisp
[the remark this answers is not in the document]
scymtym: Renamed.

Core/clim-core/table-formatting.lisp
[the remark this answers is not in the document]
dkochmanski: yes, imo it is easy to miss when put within apply form

Core/clim-core/table-formatting.lisp
scymtym: Done.

Core/clim-core/table-formatting.lisp
scymtym: Renamed.

Core/clim-core/table-formatting.lisp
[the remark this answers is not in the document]
scymtym: Changed.

Core/clim-core/table-formatting.lisp
scymtym: They are ordinary functions now.

Core/clim-core/table-formatting.lisp
```diff
@@ -54,6 +54,113 @@
...
+;;; ARGS is a list of argument specifications of the form
+;;;
+;;;   (name &key bindp default transform passp)
+;;;
+;;; where NAME specifies the name of the keyword argument in the
+;;; generated function and the generated macro.
+;;;
+;;; BINDP controls whether
+;;; NAME should be bound when BODY is evaluated even if the argument
+;;; is just "passed through".
```
dkochmanski, line 87: wrap a line

scymtym: Fixed.

Core/clim-core/table-formatting.lisp
```diff
@@ -259,66 +325,35 @@ skips intervening non-table output record structures."))
...
+    ((move-cursor                :default   t :passp nil)
+     (x-spacing                  :transform (parse-space stream (or x-spacing #\Space) :horizontal))
+     (y-spacing                  :transform (parse-space stream (or y-spacing
+                                                                    (stream-vertical-spacing stream))
+                                                         :vertical))
+     (multiple-columns           :bindp t)
+     (multiple-columns-x-spacing :transform (if multiple-columns-x-spacing
+                                                (parse-space stream multiple-columns-x-spacing :horizontal)
+                                                x-spacing))
+     (equalize-column-widths))
```
dkochmanski, line 338: don't align forms in columns (applies to other definitions too). If you want to have them aligned, start each argument from a new line, like
```
(move-cursor
 :default t
 :passp nil)
```

or simply remove extra spaces.

scymtym: Changed.

dkochmanski:
lgtm, thanks!

PR #1054, opened 2020-08-26
Cleanup frames

dkochmanski:
This is a "cleanup" part o the PR #1048.


dkochmanski:
This branch was reviewed as part of 1048, and there are no regression test-suite wise; so I'm merging it as is.

PR #1055, opened 2020-08-26
define-application-frame: check option types during macroexpansion

scymtym:


Core/clim-core/frames/define-application-frame.lisp
```diff
@@ -91,11 +91,11 @@
...
                  (:layouts               * :conflicts (:pane))
-                 (:command-table         1)
-                 (:command-definer       1)
-                 (:menu-bar              ensure-1)
+                 (:command-table         1 :type (cons symbol list))
+                 (:command-definer       1 :type symbol)
+                 (:menu-bar              ensure-1 :type (or symbol list))
                  (:disabled-commands     *)
-                 (:top-level             1)
+                 (:top-level             1 :type (cons (or symbol cons) list))
```
dkochmanski, line 98: > :top-level form, where form is a list whose first element is the name of a function to be called to execute the top-level loop. The function must take at least one argument, the frame. The rest of the list consists of additional arguments to be passed to the function. The default function is default-frame-top-level.


why is there (or symbol cons)?

Core/clim-core/frames/define-application-frame.lisp
```diff
@@ -91,11 +91,11 @@
                  (:pane                  * :conflicts (:panes :layouts))
                  (:panes                 * :conflicts (:pane))
                  (:layouts               * :conflicts (:pane))
-                 (:command-table         1)
-                 (:command-definer       1)
-                 (:menu-bar              ensure-1)
+                 (:command-table         1 :type (cons symbol list))
+                 (:command-definer       1 :type symbol)
+                 (:menu-bar              ensure-1 :type (or symbol list))
```
dkochmanski, line 96: What does ensure-1 mean (I find it hard to read this macro)?
menu-bar is either
t, nil, symbol or a list of lists

consider :menu-bar (foo bar qux), which is neither, but ensure car will take foo and conclude, that it is a symbol and make it pass.

dkochmanski (changes requested):

Core/clim-core/frames/define-application-frame.lisp
scymtym: The spec says

> `:top-level` *form,* where *form* is a list whose first element is **the name of a function** to be called to execute the top-level loop.

To acommodate extended function names like `(setf NAME)`, `cons` must be allowed.

Core/clim-core/frames/define-application-frame.lisp
scymtym: Options are considered to be of the form `(name . values)`, i.e. a list of values. The is used for `:panes` or the `:default-initargs` `defclass` option.

So `ensure-1` operates on the list of values, not the value, and is needed to support `(:menu-bar . t)` as well as `(:menu-bar t)`. Accepting the former is probably wrong, but I did that to preserve the old behavior (see line 847 of the left side of https://github.com/McCLIM/McCLIM/commit/2632804c7a21d69e343ea9ff8b3dc8e360a49f44#).

dkochmanski:
lgtm, thanks

PR #1056, opened 2020-08-28
cleanup input editing

dkochmanski:


Core/clim-core/presentations/translators.lisp
```diff
@@ -289,14 +289,25 @@ and used to ensure that presentation-translators-caches are up to date.")
...
+  (labels ((rec (type-specifier)
+             (with-presentation-type-decoded (name parameters)
+                 type-specifier
+               (cond ((member name '(or and))
+                      (list* name (mapcar #'rec parameters)))
+                     ((and include-parameters-p parameters)
+                      (cons name parameters))
+                     (t
+                      name)))))
+    (rec type-specifier)))
```
dkochmanski, line 302: this commit made me thinking, why do we take only the type name from to-type. I've made a small table which illustrates, that to-type parameters are important to the translator applicability, hence it influences how many translators are included in to translator cache entry, because the condition for including a translator is

```
(and (stupid-subtypep from-type (from-type tr))
                      (stupid-subtypep (to-type tr) to-type))
```

that leads to the following example table:
```
foo       -> bar
foo       -> (bar 3 4)
(foo 1 2) -> (bar 3 4)
(foo 1 2) -> bar

| from-type | context   | tr from-type | tr to-type | applicable? |
|-----------+-----------+--------------+------------+-------------|
| foo       | bar       | foo          | bar        | t           |
|           |           | foo          | (bar 3 4)  | t           |
|           |           | (foo 1 2)    | (bar 3 4)  | nil         |
|           |           | (foo 1 2)    | bar        | nil         |
|-----------+-----------+--------------+------------+-------------|
| foo       | (bar 3 5) | foo          | bar        | nil         |
|           |           | foo          | (bar 3 4)  | t           |
|           |           | (foo 1 2)    | (bar 3 4)  | nil         |
|           |           | (foo 1 2)    | bar        | nil         |
|-----------+-----------+--------------+------------+-------------|
| (foo 1)   | (bar 3 5) | foo          | bar        | t           |
|           |           | foo          | (bar 3 4)  | t           |
|           |           | (foo 1 2)    | (bar 3 4)  | t           |
|           |           | (foo 1 2)    | bar        | t           |
|-----------+-----------+--------------+------------+-------------|
| (foo 1)   | bar       | foo          | bar        | t           |
|           |           | foo          | (bar 3 4)  | t           |
|           |           | (foo 1 2)    | (bar 3 4)  | t           |
|           |           | (foo 1 2)    | bar        | t           |
|-----------+-----------+--------------+------------+-------------|
```

in other words, include-parameters-p should be always t I think.

dkochmanski: also, I think that cache correctness could be added to regression tests.

dkochmanski: hm, what I've said is not right, from the comment of stupid-subtypep (which I have co-authored ^_^):
```
;;; This is to implement the requirement on presentation translators for doing
;;; subtype calculations without reference to type parameters. We are generous
;;; in that we return T when we are unsure, to give translator testers a
;;; chance to accept or reject the translator. This is essentially
;;;
;;;   (multiple-value-bind (yesp surep)
;;;       (presentation-subtypep maybe-subtype maybe-supertype)
;;;     (or yesp (not surep)))
;;;
;;; except faster. The above relation should not be taken literally. The
;;; actual relation between predicates is
;;;
;;;   (unless (stupid-subtypep maybe-subtype maybe-supertype)
;;;     (multiple-value-bind (yesp surep)
;;;         (presentation-subtypep maybe-subtype maybe-supertype)
;;;       (assert (and (null yesp) surep))))
;;;
;;; for example the first relation is not preserved with
;;;
;;;   (presentation-subtypep '(integer 1 2) '(integer 2 4)) ; -> (nil t)
;;;   (stupid-subtypep       '(integer 1 2) '(integer 2 4)) ; -> t
;;;
````

in other words, we should take from-name and to-name as a key unless the corresponding from/to type are either or, and, completion, subset-completion or sequence. So it is the other way around, include-parameters-p should be always nil, and the corresponding `cond` for meta-types should be appropriately extended.

adding also parameters to hash table keys (include-parameters-p always t) won't affect the semantics but will make the hash table larger.

Comment (dkochmanski) [reply Core/clim-core/presentations/translators.lisp:302]: alternatively, to not put the burden of testing on testers, we could always include parameters and call presentation-subtypep; I think that this may make actually more sense.

Comment (dkochmanski) [reply Core/clim-core/presentations/translators.lisp:302]: another part of the question is that since we cache only names, then should we treat metatypes differently? given that we check for the actual subtypep in the tester, we could significantly simplify the function stupid-subtypep at a cost of having more /potentially/ applicable translators for meta types (per cache entry).

Comment (dkochmanski) [reply Core/clim-core/presentations/translators.lisp:302]: stupid-subtypep would look like this:
```
(defun stupid-subtypep (maybe-subtype maybe-supertype)
  "Return T if MAYBE-SUBTYPE is a presentation subtype of
MAYBE-SUPERTYPE, regardless of parameters."
  (when (or (eq maybe-subtype nil) (eq maybe-supertype t))
    (return-from stupid-subtypep t))
  (when (eql maybe-subtype maybe-supertype)
    (return-from stupid-subtypep t))
  (let ((maybe-subtype-name (presentation-type-name maybe-subtype))
        (maybe-supertype-name (presentation-type-name maybe-supertype)))
    ;; Check for meta types.
    (when (or (eq maybe-subtype-name 'or)
              (eq maybe-supertype-name 'or)
              (eq maybe-subtype-name 'and)
              (eq maybe-supertype-name 'and)
              (eq maybe-subtype-name 'completion)
              (eq maybe-supertype-name 'completion)
              (eq maybe-subtype-name 'subset-completion))
      (return-from stupid-sbutypep t))
    ;; Check for normal types.
    (let ((subtype-meta (get-ptype-metaclass maybe-subtype-name))
          (maybe-supertype-meta (get-ptype-metaclass maybe-supertype-name)))
      (when (and subtype-meta maybe-supertype-meta)
        (map-over-ptype-superclasses (lambda (super)
                                       (when (eq maybe-supertype-meta super)
                                         (return-from stupid-subtypep t)))
                                     maybe-subtype-name))
      nil)))
```

dkochmanski:
rebased against master (removed translator changes).

scymtym:
Looks good to me.

PR #1057, opened 2020-08-30
Replace 606

dkochmanski:
This PR is a replacement for #606. Slider improvements were merged earlier, so they are not backported, and scroll-bar handle-event method is not backported because I don't personally think it is a good idea -- it is not consistent with existing toolkits.

- recognize wheel-left and wheel-right pointer gestures
- add a pointer-scroll gesture
- add named gestures scroll-up and scroll-down
- clim-stream-pane does not scroll the sheet if the gesture has associated translator
- mouse-wheel-scroll-mixin forwards the pointer-scroll-event if it has modifiers

### DIFF
--- a/Core/clim-basic/extended-streams/pointer-tracking.lisp
+++ b/Core/clim-basic/extended-streams/pointer-tracking.lisp
@@ -154,12 +154,12 @@
          (modifier-state))
     (flet ((track-pointer-event (event)
              (multiple-value-call #'track-event state event
-               (let ((sheet (event-sheet event)))
-                 (get-pointer-position (sheet event)
-                   (if (not transformp)
-                       (values x y)
-                       (with-sheet-medium (medium sheet)
-                         (transform-position (medium-transformation medium) x y))))))))
+               (let ((x (pointer-event-x event))
+                     (y (pointer-event-y event)))
+                 (if (not transformp)
+                     (values x y)
+                     (with-sheet-medium (medium (event-sheet event))
+                       (transform-position (medium-transformation medium) x y)))))))
       ;; Synthesize a pointer motion event for the current pointer
       ;; position so that appropriate handlers are called even if no
       ;; event immediately follows the INVOKE-TRACKING-POINTER call.
--- a/Core/clim-basic/extended-streams/stream-input.lisp
+++ b/Core/clim-basic/extended-streams/stream-input.lisp
@@ -119,7 +119,8 @@
         (input-buffer (stream-input-buffer stream)))
     (flet ((process-gesture (raw-gesture)
              (when (and pointer-button-press-handler
-                        (typep raw-gesture 'pointer-button-press-event))
+                        (typep raw-gesture '(or pointer-button-press-event
+                                                pointer-scroll-event)))
                (funcall pointer-button-press-handler stream raw-gesture))
              (when-let ((gesture (stream-process-gesture stream raw-gesture t)))
                (return-from stream-read-gesture gesture))))
@@ -429,17 +430,20 @@ known gestures."
                   (symbolp device-name))
              (setq device-name (or (cdr (assoc device-name +name-to-char+))
                                    device-name)))
-            ((and (member type '(:pointer-button
-                                 :pointer-button-press
-                                 :pointer-button-release)
-                          :test #'eq))
+            ((member type '(:pointer-button
+                            :pointer-button-press
+                            :pointer-button-release
+                            :pointer-scroll)
+                     :test #'eq)
              (let ((real-device-name
                      (case device-name
-                       (:left       +pointer-left-button+)
-                       (:middle     +pointer-middle-button+)
-                       (:right      +pointer-right-button+)
-                       (:wheel-up   +pointer-wheel-up+)
-                       (:wheel-down +pointer-wheel-down+)
+                       (:left        +pointer-left-button+)
+                       (:middle      +pointer-middle-button+)
+                       (:right       +pointer-right-button+)
+                       (:wheel-up    +pointer-wheel-up+)
+                       (:wheel-down  +pointer-wheel-down+)
+                       (:wheel-left  +pointer-wheel-left+)
+                       (:wheel-right +pointer-wheel-right+)
                        (t (error "~S is not a known button" device-name)))))
                (setq device-name real-device-name))))
       (values type device-name modifier-state))))
@@ -496,12 +500,21 @@ known gestures."
              (eql type :pointer-button))
          (eql (pointer-event-button event) device-name)
          (eql (event-modifier-state event) modifier-state)))
+  (:method ((event pointer-scroll-event)
+            type
+            device-name
+            modifier-state)
+    (and (or (eql type :pointer-scroll)
+             (eql type :pointer-button))
+         (eql (pointer-event-button event) device-name)
+         (eql (event-modifier-state event) modifier-state)))
   (:method ((event pointer-button-event)
             type
             device-name
             modifier-state)
     (and (or (eql type :pointer-button-press)
              (eql type :pointer-button-release)
+             (eql type :pointer-scroll)
              (eql type :pointer-button))
          (eql (pointer-event-button event) device-name)
          (eql (event-modifier-state event) modifier-state)))
@@ -563,6 +576,11 @@ known gestures."
 (define-gesture-name :menu :pointer-button-press (:right))
 (define-gesture-name :edit :pointer-button-press (:left :meta))
 (define-gesture-name :delete :pointer-button-press (:middle :shift))
+
+(define-gesture-name :scroll-up :pointer-scroll (:wheel-up))
+(define-gesture-name :scroll-down :pointer-scroll (:wheel-down))
+(define-gesture-name :scroll-left :pointer-scroll (:wheel-left))
+(define-gesture-name :scroll-right :pointer-scroll (:wheel-right))
 
 ;;; Define so we have a gesture for #\newline that we can use in
 ;;; *standard-activation-gestures*
--- a/Core/clim-basic/windowing/events.lisp
+++ b/Core/clim-basic/windowing/events.lisp
@@ -273,35 +273,6 @@
 (defconstant +hyper-key+             #x1000)
 (defconstant +alt-key+               #x2000)
 
-(defmacro key-modifier-state-match-p (button modifier-state &body clauses)
-  (let ((button-names '((:left       . +pointer-left-button+)
-                        (:middle     . +pointer-middle-button+)
-                        (:right      . +pointer-right-button+)
-                        (:wheel-up   . +pointer-wheel-up+)
-                        (:wheel-down . +pointer-wheel-down+)))
-        (modifier-names '((:shift . +shift-key+)
-                          (:control . +control-key+)
-                          (:meta . +meta-key+)
-                          (:super . +super-key+)
-                          (:hyper . +hyper-key+)))
-        (b (gensym))
-        (m (gensym)))
-    (labels ((do-substitutes (c)
-               (cond
-                 ((null c)
-                  nil)
-                 ((consp c)
-                  (cons (do-substitutes (car c)) (do-substitutes (cdr c))))
-                 ((assoc c button-names)
-                  (list 'check-button (cdr (assoc c button-names))))
-                 ((assoc c modifier-names)
-                  (list 'check-modifier (cdr (assoc c modifier-names))))
-                 (t
-                  c))))
-      `(flet ((check-button (,b) (= ,button ,b))
-              (check-modifier (,m) (not (zerop (logand ,m ,modifier-state)))))
-         (and ,@(do-substitutes clauses))))))
-
 ;; Key names are a symbol whose value is port-specific. Key names
 ;; corresponding to the set of standard characters (such as the
 ;; alphanumerics) will be a symbol in the keyword package.
--- a/Core/clim-basic/windowing/events.lisp
+++ b/Core/clim-basic/windowing/events.lisp
@@ -120,6 +120,12 @@
    (key-character :initarg :key-character :reader keyboard-event-character
                   :initform nil)))
 
+(defmethod print-object ((event keyboard-event) stream)
+  (print-unreadable-object (event stream :type t :identity nil)
+    (format stream "NAME ~s CHARACTER ~s"
+            (keyboard-event-key-name event)
+            (keyboard-event-character event))))
+
 (define-event-class key-press-event (keyboard-event)
   ())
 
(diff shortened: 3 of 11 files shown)


dkochmanski:
A throwaway test code:

```
(in-package clim-user)

(defun display-foo (frame pane)
  (let ((counter (counter frame))
        (what-we-want (what-we-want frame)))
    (setf (stream-cursor-position pane)
          (values 200 200))
    (present (what-we-want frame))
    (terpri)
    (with-drawing-options
        (pane :ink (if (= counter what-we-want)
                       +dark-green+
                       +red+))
      (present (counter frame)))
    (setf (stream-cursor-position pane)
          (values 200 400))
    (print "HI")))

(define-application-frame foo ()
  ((counter :initform 0 :accessor counter)
   (what-we-want :initform 123 :accessor what-we-want))
  (:pointer-documentation t)
  (:pane :application
   :display-function 'display-foo
   :text-style (make-text-style nil nil :huge)
   :text-margins '(:left 20 :top 10)
   :END-OF-PAGE-ACTION :ALLOW))

(define-foo-command incf-foo ((arg integer :gesture :scroll-up))
  (declare (ignore arg))
  (incf (counter *application-frame*)))

(define-foo-command decf-foo ((arg integer :gesture :scroll-down))
  (declare (ignore arg))
  (decf (counter *application-frame*)))

(define-gesture-name :scroll-up-much
  :pointer-scroll (:wheel-up :control))

(define-gesture-name :scroll-very-up-much
  :pointer-scroll (:wheel-up :control :meta))

(define-gesture-name :scroll-very-down-much
  :pointer-scroll (:wheel-down :control :meta))

(define-gesture-name :scroll-down-much
  :pointer-scroll (:wheel-down :control))

(define-foo-command incf-foo1 ((arg integer :gesture :scroll-up-much))
  (declare (ignore arg))
  (incf (counter *application-frame*) 10))

(define-foo-command decf-fo1o ((arg integer :gesture :scroll-down-much))
  (declare (ignore arg))
  (decf (counter *application-frame*) 10))

(define-foo-command incf-qux ((arg integer :gesture :scroll-very-up-much))
  (declare (ignore arg))
  (incf (counter *application-frame*) 100))

(define-foo-command decf-qux ((arg integer :gesture :scroll-very-down-much))
  (declare (ignore arg))
  (decf (counter *application-frame*) 100))

(define-foo-command zerom ((arg integer :gesture :select))
  (declare (ignore arg))
  (setf (counter *application-frame*) 0)
  (setf (what-we-want *application-frame*) (random 1000)))

(run-frame-top-level (make-application-frame 'foo))

```

dkochmanski:
adding a wip tag, because I've found an unresolved issue, namely that when the scroll event coordinates /after/ scrolling the sheet are over the presentation with translator accepting the scrolling gesture, the translator is invoked (i.e command is executed in the example above). I'm attaching a video illustrating the issue.
bug

scymtym (changes requested):
A few comments for the first commit.

Core/clim-basic/windowing/events.lisp
```diff
@@ -274,33 +274,38 @@
 (defconstant +alt-key+               #x2000)
 
 (defmacro key-modifier-state-match-p (button modifier-state &body clauses)
```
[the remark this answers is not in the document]
dkochmanski: yes, done

Core/clim-basic/extended-streams/stream-input.lisp
```diff
@@ -440,15 +441,18 @@ known gestures."
                                    device-name)))
             ((and (member type '(:pointer-button
                                  :pointer-button-press
-                                 :pointer-button-release)
+                                 :pointer-button-release
+                                 :pointer-scroll)
                           :test #'eq))
```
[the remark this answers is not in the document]
dkochmanski: yes, fixed

Core/clim-basic/extended-streams/stream-input.lisp
```diff
@@ -573,6 +586,9 @@ known gestures."
 (define-gesture-name :edit :pointer-button-press (:left :meta))
 (define-gesture-name :delete :pointer-button-press (:middle :shift))
 
+(define-gesture-name :scroll-up :pointer-scroll (:wheel-up))
+(define-gesture-name :scroll-down :pointer-scroll (:wheel-down))
```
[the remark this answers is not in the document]
dkochmanski: well, why not (not much use for me for I have only one scroller)

dkochmanski:
I'm taking out the [wip] tag because I've resolved the problem with coordinates.

Core/clim-basic/windowing/ports.lisp
```diff
@@ -256,8 +256,8 @@ is a McCLIM extension.")
              (new-event (shallow-copy-object event event-class)))
         (when (typep new-event 'pointer-event)
           (get-pointer-position (target-sheet new-event)
-            (setf (slot-value new-event 'x) x
-                  (slot-value new-event 'y) y)))
+            (setf (slot-value new-event 'sheet-x) x
+                  (slot-value new-event 'sheet-y) y)))
```
scymtym, line 260: (I wanted to attach this to the correct commit, but not necessarily to this code location)

I think that several places would have to be adapted to this change:
* One is `throw-object-ptype` which produced an event with unbound `sheet-x` slot when i tried this branch.
* Backend methods on `synthesize-pointer-motion-event`. I'm not sure whether those can work unmodified or not.
* Are there other places that directly construct pointer events?

Alternatively, can sensible default values for the new slots be computed? Maybe at least default values good enough for synthesized events?

Core/clim-basic/windowing/events.lisp
```diff
@@ -102,17 +103,40 @@
          (defmethod event-type ((event ,name))
            ',type)))))
 
+;;; We have three pairs of coordinates. sheet-x and sheet-y are the sheet's
```
scymtym, line 106: Maybe
```common-lisp
"
* SHEET-X and SHEET-Y are the position of the pointer event in the coordinate system of the sheet.
* X and Y are the position of the pointer event in the coordinate system of the mirror of the sheet.
* GRAFT-X and GRAFT-Y are the position of the pointer event in the screen coordinate system.
"
```
Could also write one comment above each pair of slots.

Core/clim-basic/windowing/events.lisp
```diff
@@ -69,6 +69,7 @@
    ;; This slot is pretty much required in order to call handle-event. Some
    ;; events have something other than a sheet in this slot, which is gross.
    (sheet :initarg :sheet
+          :initform nil
           :reader event-sheet)))
```
scymtym, line 73: Can we add `(:default-initargs :sheet (alexandria:required-argument :sheet))` sheet here and do `:sheet nil` as a workaround in EAS?

scymtym: If possible as a separate commit since both changes are rather severe and separating them is better for bisecting, CI, reviewing, etc.

scymtym (changes requested):
Remarks in commit comments.

Core/clim-basic/windowing/events.lisp
dkochmanski: I'm going to make it
```
;;; We have three pairs of the pointer event coordinates in different
;;; coordinate systems:
;;;
;;; * (X Y) - native coordinates (the mirror)
;;; * (SHEET-X SHEET-Y) - sheet coordinates
;;; * (GRAFT-X GRAFT-Y) - graft coordinates (the screen)
```

"sheet coordinates" and "native coordinates" are terms which may be found in the spec.

Core/clim-basic/windowing/ports.lisp
dkochmanski: how could that be? we initialize sheet-x and sheet-y in initialize-instance based on x/y initargs:
```
(defmethod initialize-instance :after ((event device-event) &key x y)
  (when-let ((sheet (event-sheet event)))
    (multiple-value-bind (sheet-x sheet-y)
        (untransform-position (sheet-native-transformation sheet) x y)
      (setf (slot-value event 'sheet-x) (floor sheet-x)
            (slot-value event 'sheet-y) (floor sheet-y)))))
```
could you provide a test case with which I could reproduce the issue?

dkochmanski: I've made them default to X/Y, so if the sheet is not supplied when making the event instance it will fallback to that:

```
(defmethod initialize-instance :after ((event device-event) &key x y)
  (if-let ((sheet (event-sheet event)))
    (multiple-value-bind (sheet-x sheet-y)
        (untransform-position (sheet-native-transformation sheet) x y)
      (setf (slot-value event 'sheet-x) (floor sheet-x)
            (slot-value event 'sheet-y) (floor sheet-y)))
    (setf (slot-value event 'sheet-x) x
          (slot-value event 'sheet-y) y)))
```

Core/clim-basic/windowing/events.lisp
dkochmanski: this will break i.e throw-object-ptype, so I'd rather keep it as it is now (so we change a default behavior as little as possible) -- this is mostly to address the issue of incorrect coordinates for scrolling, not to fix the "gross" behavior of putting something what is not a sheet in the event.

Core/clim-basic/windowing/ports.lisp
scymtym: Why are the `floor` calls necessary?

Core/clim-basic/windowing/events.lisp
scymtym: If the sheet is required by the spec (and code unprepared to handle `event-sheet` returning `nil`) I don't think we should further accommodate not supplying the `:sheet` initarg. That is, I would prefer not to add this default or the conditional in the `initialize-instance` method for `device-event`.

If there are things besides the one instance in ESA that would definitely break, things may be different, but do we know of any?

dkochmanski: checking for nil sheet is still necessary in initialize-instance, because some functions simply pass supplied sheet argument, i.e throw-ptype (not to mention ESA). I've made the initarg required and kept the test.

Core/clim-basic/windowing/ports.lisp
dkochmanski: beats me, fixed.

dkochmanski:
resolved all issues. the failure in CI was due to a test not supplying the sheet initarg, this is fixed now.

Core/clim-core/panes/stream-panes.lisp
```diff
@@ -83,6 +83,20 @@
                          (event window-manager-focus-event))
   (setf (port-keyboard-input-focus (port sheet)) sheet))
 
+;; This method is defined to prevent the sheet scrolling defined for the
+;; mouse-wheel-scroll-mixin when the event activates a command. In other
+;; words, we scroll the clim-stream-pane only when scrolling does not match
+;; the current input context. -- jd 2020-08-29
```
scymtym, line 89: Please use three semicolons or move into method body.

Core/clim-basic/windowing/events.lisp
```diff
@@ -134,6 +140,12 @@
    (graft-x :reader pointer-event-native-graft-x)
    (graft-y :reader pointer-event-native-graft-y) ))
 
+(defmethod print-object ((event pointer-event) stream)
+  (print-unreadable-object (event stream :type t :identity nil)
+    (format stream "X: ~d, Y: ~d"
```
scymtym, line 173: I know this is cosmetic but please make the format consistent (labels as keywords or with following colons, comma between fields or not).

Core/clim-basic/extended-streams/pointer-tracking.lisp
```diff
@@ -154,12 +154,12 @@
          (modifier-state))
     (flet ((track-pointer-event (event)
              (multiple-value-call #'track-event state event
-               (let ((sheet (event-sheet event)))
-                 (get-pointer-position (sheet event)
-                   (if (not transformp)
-                       (values x y)
-                       (with-sheet-medium (medium sheet)
-                         (transform-position (medium-transformation medium) x y))))))))
+               (let ((x (pointer-event-x event))
```
scymtym, line 157: Commit message: "where coordinates w~h~ere computed"

Core/clim-basic/windowing/events.lisp
```diff
@@ -146,40 +176,6 @@
...
-  (get-pointer-position ((event-sheet event) event) x))
-
-(defmethod pointer-event-y ((event pointer-event))
-  (get-pointer-position ((event-sheet event) event) y))
-
-(defgeneric pointer-event-position* (pointer-event))
-
-(defmethod pointer-event-position* ((event pointer-event))
-  (get-pointer-position ((event-sheet event) event)
-    (values x y)))
```
scymtym, line 159: This has been removed, right? If the removal was on purpose, please mention it in the commit message (preferably with a justification).

Core/clim-basic/windowing/events.lisp
```diff
@@ -63,23 +63,20 @@
...
-    (with-lock-held (*last-timestamp-lock*)
-      (if timestamp
-          (maxf *last-timestamp* timestamp)
-          (setf (slot-value event 'timestamp)
-                (incf *last-timestamp*))))))
+  (with-lock-held (*last-timestamp-lock*)
+    (if-let ((timestamp (event-timestamp event)))
+      (maxf *last-timestamp* timestamp)
+      (setf (slot-value event 'timestamp)
+            (incf *last-timestamp*)))))
```
scymtym, line 79: I think the previous code tried to hold the lock as briefly as possible. We should probably use atomic operations here.

scymtym (changes requested):
Requested minor changes in commit comments.

Core/clim-basic/windowing/events.lisp
dkochmanski: As discussed on IRC - I agree, however, atomic operations have poor coverage among implementations. Our own clim-sys:atomic-incf errors on MP environments (that's something we could improve). Either way, out of the scope of this PR.

Core/clim-basic/windowing/events.lisp
dkochmanski: fixed. I took hint from regions.lisp, where
- if only pair of coordinates is printed, it is "~s ~s"
- when name val are intervened, it is "NAME VAL NAME VAL"
- when name coord are intervened, it is "START ~s:~s END ~s:~s"

Core/clim-core/panes/stream-panes.lisp
dkochmanski: fixed

Core/clim-basic/extended-streams/pointer-tracking.lisp
dkochmanski: fixed

Core/clim-basic/windowing/events.lisp
dkochmanski: fixed

I've added to the commit message the following:

```
We also remove a generic function POINTER-EVENT-POSITION* which was returning
the event coordinates in sheet coordinate system /at the time of calling it/.
This function is not used in McCLIM codebase and is not exported.
```

dkochmanski:
all remarks have been addressed, the PR is rebased against current master branch.

dkochmanski:
I've rebased this branch against current master.

scymtym (approved):

PR #1059, opened 2020-09-01
Improve commands

dkochmanski:
This PR will be followed by another one with more semantically involved changes and does not itself contain anything controversial.


Core/clim-basic/extended-streams/stream-input.lisp
```diff
@@ -393,6 +393,10 @@ keys read."))
 (defmethod stream-unread-char ((stream standard-extended-input-stream) char)
   (stream-unread-gesture stream char))
 
+(defmethod stream-peek-char ((stream standard-extended-input-stream))
+  (let ((ch (stream-read-gesture stream :peek-p t)))
+    (and (characterp ch) ch)))
+
 
 ;;; stream-read-gesture on string strings. Needed for accept-from-string.
```
scymtym, line 401: Looks like two empty lines.

The following comment (from a different commit) looks wrong. Should that have been "string streams"?

Core/clim-core/standard-presentations.lisp
```diff
@@ -682,8 +681,10 @@
         (t (getf (cdr element) :value))))
 
 (defun member-alist-doc-key (element)
-  (if (and (consp element) (consp (cdr element)) (consp (cddr element)))
-      (getf (cdr element) :documentation)))
+  (when (and (consp element)
+             (consp (cdr element))
+             (consp (cddr element)))
```
scymtym, line 686: How about `(typep element '(cons t (cons t cons)))`? (Not that I understand what this check does)

Core/clim-core/commands/commands.lisp
```diff
@@ -673,50 +673,50 @@ examine the type of the command menu item to see if it is
...
+                        ,translator-name
+                        (,(eval ptype) ,command-name ,command-table
+                         :gesture ,gesture
+                         ,@(unless (getf translator-options :documentation)
+                             `(:documentation ,default-documentation))
+                         ,@translator-options)
+                        (object)
+                      (list ,@command-args)))))))
+      (loop for arg in args
+            for arg-index from 0
```
scymtym, line 736: Maybe destructure here: `for (name ptype options) in args`? Then `(getf options :gesture)`, keep the gesture and pass it to `make-define-gesture-translator` instead of the `destructuring-bind`?

Core/clim-core/commands/commands.lisp
```diff
@@ -406,6 +407,25 @@ designator) inherits menu items."
...
                        &allow-other-keys)
+  (ecase type
+    (:command
+     ;; This is specified to be a cons, but McCLIM is more permissive
+     ;; (and we don't want to break the backward compatibility).
+     (unless (consp value)
+       (setf value (list value))))
+    (:function
+     ;; A function of two arguments (funcalled).
+     (check-type value (or function symbol)))
```
scymtym, line 418: Lacks `(cons (eql setf) (cons symbol null))`. We should probably add a `deftype function-designator` for that.

Libraries/ESA/esa.lisp
```diff
@@ -1008,10 +1008,6 @@ used.")
      (find-keystroke-item event table :errorp nil))))
 
 (defun set-key (command table gestures)
-  ;; WTF?
-  #-(and)
-  (unless (consp command)
-    (setf command (list command)))
```
scymtym, line 1014: Should we replace this with `(check-type command cons)`?

Core/clim-core/commands/parsers.lisp
```diff
@@ -234,9 +237,11 @@
...
         collect `(,arg (progn
                          (write-char ,seperator ,stream)
-                         (present (car ,command-args) ,ptype-form
-                                  :stream ,stream)
-                         (pop ,command-args)))
+                         (let ((value (pop ,command-args)))
+                           (if (unsupplied-argument-p value)
+                               (with-text-face (,stream :italic)
+                                 (write-string "<unsupplied>" ,stream))
+                               (present value ,ptype-form :stream ,stream)))))
```
scymtym: Since this appears in generated code and is repeated in this file, I think defining a function `(defun present-argument-value (value ptype stream))` would make sense.

Core/clim-core/commands/parsers.lisp
```diff
@@ -247,8 +252,10 @@
                           ,seperator
                           ,(keyword-arg-name-from-symbol arg)
                           ,seperator)
-                  (present ,key-arg-val ,ptype-form
-                           :stream ,stream))
+                  (if (unsupplied-argument-p ,key-arg-val)
+                      (with-text-face (stream :italic)
```
scymtym, line 257: Missing unquote for `stream`? Only relevant if no function is defined.

scymtym (changes requested):
Suggested changes in commit comments.

Core/clim-basic/extended-streams/stream-input.lisp
dkochmanski: comment fixed. re empty lines, it is page break (in emacs it shows as ^L)

Core/clim-core/standard-presentations.lisp
dkochmanski: this was just stylistic line break change and when->if. a function above also uses cddr etc, so having it this way is more consistent.

Core/clim-core/commands/commands.lisp
dkochmanski: fixed

Core/clim-core/commands/commands.lisp
dkochmanski: fixed

Libraries/ESA/esa.lisp
dkochmanski: we do ensure-list in make-menu-item from now on, so this is not a problem.

I gather that the person who originally wrote the code consulted the spec and knew, that the command is a list and wanted to allow atoms, while next person did not, and assumed that this is bogus. First I've started to correct all places in McCLIM to pass lists (not symbols) where appropriate, however I've decided that it may break code which was already written, so if the command is atom, then we coerce it to a list in make-menu-item.

Core/clim-core/commands/parsers.lisp
dkochmanski: fixed.

Core/clim-core/commands/parsers.lisp
dkochmanski: fixed and squashed.

dkochmanski:
all concerns are addressed

dkochmanski:
I've pushed next commit which introduces a final split of the commands file

Core/clim-core/commands/commands.lisp
```diff
@@ -673,50 +673,47 @@ examine the type of the command menu item to see if it is
...
+  (let* ((readable-command-name
+           (command-name-from-symbol command-name)) ; XXX or :NAME
+         (default-documentation
+           `((object stream)
+             (orf stream *standard-output*)
+             (format stream "~A " ,readable-command-name)
+             (present object (presentation-type-of object) ; type?
+                      :stream stream
+                      :acceptably nil
+                      :sensitive nil))))
```
scymtym, line 135: Could move this into the `flet` as a function to communicate the fact that it is not always needed. Also reduces consing a bit in some case, but I don't know whether we expand enough commands for that to matter.

Core/clim-core/commands/commands.lisp
```diff
@@ -406,6 +407,25 @@ designator) inherits menu items."
...
                        &allow-other-keys)
+  (ecase type
+    (:command
+     ;; This is specified to be a cons, but McCLIM is more permissive
+     ;; (and we don't want to break the backward compatibility).
+     (unless (consp value)
+       (setf value (list value))))
+    (:function
+     ;; A function of two arguments (funcalled).
+     (check-type value (or function symbol (cons (eql setf) null))))
```
scymtym, line 96: `(cons (eql setf) null)` looks wrong, wouldn't that be just `(setf)`. I think it should be `(cons (eql setf) (cons symbol null))`.

scymtym (changes requested):
Minor comments.

scymtym:
This branch produces a new warning:
```
WARNING:
   redefining DISPLAY-COMMAND-TABLE-MENU (#<STANDARD-CLASS CLIM:STANDARD-COMMAND-TABLE>
                                          #<STANDARD-CLASS TRIVIAL-GRAY-STREAMS:FUNDAMENTAL-OUTPUT-STREAM>) in DEFMETHOD
```
I think this is due to accidentally duplicated code:
```
find . -type f -exec grep --color -nH --null -e display-command-table-menu \{\} +
./commands/processor.lisp:180:(defmethod display-command-table-menu ((command-table standard-command-table)
./commands/processor.lisp:206:                    (apply #'display-command-table-menu
./gadgets/menu.lisp:399:(defmethod display-command-table-menu ((command-table standard-command-table)
./gadgets/menu.lisp:419:                    (apply #'display-command-table-menu
./gadgets/menu.lisp:437:    (apply #'display-command-table-menu command-table stream args)))
```

Core/clim-core/commands/commands.lisp
dkochmanski: fixed

Core/clim-core/commands/commands.lisp
dkochmanski: fixed by adding a type to utils

dkochmanski:
should be fine now

scymtym (approved):

scymtym:
Looks good to me.

PR #1060, opened 2020-09-02
seis: remove character input methods

dkochmanski:
This commit solves the regression where reading sequences were not working.

### DIFF
--- a/Core/clim-core/builtin-commands.lisp
+++ b/Core/clim-core/builtin-commands.lisp
@@ -351,11 +351,11 @@
     (if (or subform-read auto-activate)
         (values object ptype)
         (loop
-           for c = (read-char stream)
-           until (or (activation-gesture-p c) (delimiter-gesture-p c))
-           finally
-             (when (delimiter-gesture-p c)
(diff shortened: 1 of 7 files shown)


Core/clim-core/builtin-commands.lisp
```diff
@@ -351,11 +351,11 @@
     (if (or subform-read auto-activate)
         (values object ptype)
         (loop
-           for c = (read-char stream)
+           for c = (read-gesture :stream stream)
```
scymtym, line 354: Maybe change `c` to `gesture` or similar since that name was probably chosen based on the (incorrect) assumption that the code would operate on characters.

scymtym (changes requested):
Small suggestion in commit comment.

Core/clim-core/builtin-commands.lisp
dkochmanski: fixed

scymtym (approved):

scymtym:
Looks good to me.

PR #968, opened 2020-02-20
FIx command menu bug

heshrobe:
this fixes #955
this fixes #967
In file clim-core/commands.lisp

Add an :initform nil for the docuemtation field of %menu-item.  This
is necessary to avoid an error if the mouse moves over a command menu
item.  The problem is that if a command-menu-item takes arguments, when
the mouse moves over the item an error will signaled rather than
querying for the rest of the arguments; this is because the
documentation was not being initialized.

In file clim-core/gadgets/menu.lisp

Introduce a new presentation type command-menu-item to be used in command menus diplayed in appllication-frames.
Add a translator from the command-menu-item to command.
Use the new presentation-type when displaying command table menu

In file clim-core/commands.lisp:
In the ACCEPT presentation-method for COMMAND:

1. Use WITH-INPUT-CONTEXT for presentation-type 'command so that this method
    can handle mouse clicks on objects that translate to (or are presented as) commands

2. In the handler for WITH-INPUT-CONTEXT, check if the command is a partial command and
    if so call the partial command parser and note that the input needs to be replaced

3. If the input needs to be replaced, then explictly provide the :RESCAN NIL option to
    PRESENTATION-REPLACE-INPUT.
    This has a different effect than not providing :rescan at all (which seems to default to T).

The reason for 3 is that the command returned by clicking on an object may contain arguments that print (or present) unreadably. PRESENTATION-REPLACE-INPUT presents that command into the input buffer and if :RESCAN NIL is not provided, then ACCEPT-1 will call the ACCEPT presentation method, invoking the command-parser again.  However, this time the input buffer contains the unreadable string for the argument in the command and the command argument parser will then signal an error.  This was observed in Clouseau.

The reason for 2 is that a command-menu in an application might contain a command with missing arguments.  Without this fix, clicking on the item will cause and error message saying that the command is incomplete.

PR #967, opened 2020-02-14
Fix problems with command-menu items that take commands

heshrobe:
…In file clim-core/commands.lisp add an :initform nil for the docuemtation field of %menu-item.  This is necessary to avoid an error if the mouse moves over a command menu item
In file clim-core/gadgets/menu.lisp

- Introduce a new presentation type command-menu-item to be used in command menus diplayed in appllication-frames.
- add a translator from the command-menu-item to command
- use the new presentation-type when displaying command table menu

The problem is that if a command-menu-item takes arguments, when the mouse moves over the item an error will signaled rather than querying for the rest of the arguments;  this is because the documentation was not being
initialized.  

The  approach is to change the object (to the full command-menu-item rather than just a field of it) and the presentation type used to present items in the command menu, and then to provide a presentation-translator to command.  This causes the documentation in the pointer-documentation pane to just show the command name.

The second change is to the accept method for command.   It wraps the call to the command-processor in a with-input-context form, allowing it to distingish between a menu-click and a typed command.  This then lets it invoke the partial-command-parser for a menu item with arguments that was clicked on.

These changes are necessary to fix a bug describbed  in #955, and followed up with pull request #957.  Both #955 and #957 can be closed

PR #1061, opened 2020-09-03
Replace 968

### LINKED ISSUE #955: Commands with arguments in menus
This issue concerns how the define-command macro (and similar command defining macros created by define-application-frame) handle commands that go into a menu (i.e. :menu t is specified).  clim:Define-command expands into clim-internals::%define-command which then expands into a variety of function and method definitions as well as a call to clim:add-command-to-command-table.  This call has a keyword argument :menu-command whose value is a list of the command name followed by as many occurrences of clim:*unsupplied-argument-marker* as there are arguments to the command.  For example (from an application of mine) a command with one argument
```
(define-aplan-command (com-load-model :name t :menu t)
    ((pathname 'clim:pathname)) 
...)
```
expands into:
```
(clim:add-command-to-command-table 'com-load-model 'aplan :name t :menu 't
                                    :keystroke 'nil :errorp nil :menu-command
                                    (list 'com-load-model
                                          clim:*unsupplied-argument-marker*))
```
the value of clim:*unsupplied-argument-marker* is CLIM-INTERNALS::%UNSUPPLIED-ARGUMENT-MARKER%.

This has two consequences one annoying and one serious:
1) If the menu item is clicked on rather than prompting for the arguments, it tells you you're missing arguments.
2) If the menu is in an application frame with a pointer-documentation pane then just moving the mouse over the menu item will cause an error.  The reason for this is that the pointer documentation pane will attempt to present the command, which in turn attempts to present the arguments.  But the present method for the argument is passed the symbol CLIM-INTERNALS::%UNSUPPLIED-ARGUMENT-MARKER% as the object and the presentation type of the argument (in my example above clim:pathname).   In this example, this means that the presentation method for pathnames is invoked but is passed a symbol as the object.  This can easily cause an error.

None of this occurs in the existing implementation of CLIM (e.g. the ACL one).  In the existing implementations, the call to add-command-to-command-table would supply just (list 'com-load-model) as the :menu-command argument.  As a result the pointer-documentation pane just displays the command name.  The command parser in the ACL implementation prompts for the remaining arguments.  Making the macro expansion do what the ACL implementation does would be easy and would prevent an error from being signaled, solving issue 2, but I don't understand the command parser well enough to solve issue 1.  Anybody have some guidance to offer?

dkochmanski:
fixes #955
fixes #967 
fixes #968

### DIFF
--- a/Core/clim-core/commands/processor.lisp
+++ b/Core/clim-core/commands/processor.lisp
@@ -361,14 +361,31 @@
 
 (define-presentation-method accept
     ((type command) stream (view textual-view) &key)
-  (if-let ((command (funcall *command-parser* command-table stream)))
-    (values (if (partial-command-p command)
-                (funcall *partial-command-parser*
-                         command-table stream command
-                         (position *unsupplied-argument-marker* command))
-                command)
-            type)
-    (simple-parse-error "Empty command")))
+  (let ((command-table (find-command-table command-table))
+        (start-position (and (getf options :echo t)
+                             (input-editing-stream-p stream)
+                             (stream-scan-pointer stream))))
+    (flet ((maybe-replace-input (input)
+             (when start-position
+               ;; Rescan is nil in case the clicked on thing returned
+               ;; something unparseable. -- heshrobe 2020-05-15
+               (presentation-replace-input stream input type view
+                                           :buffer-start start-position
+                                           :rescan nil))
+             input)
+           (handle-command (command)
+             (if (partial-command-p command)
+                 (funcall *partial-command-parser*
+                          command-table stream command
+                          (position *unsupplied-argument-marker* command))
+                 command)))
+      (with-input-context (type)
+          (object type)
+          (if-let ((command (funcall *command-parser* command-table stream)))
+            (values (maybe-replace-input (handle-command command)) type)
+            (simple-parse-error "Empty command"))
+        (command
+         (values (maybe-replace-input (handle-command object)) type))))))
 
 (define-presentation-type command-or-form
     (&key (command-table (frame-command-table *application-frame*)))
--- a/Core/clim-core/commands/processor.lisp
+++ b/Core/clim-core/commands/processor.lisp
@@ -38,9 +38,15 @@
         (when (and delimiter (delimiter-gesture-p delimiter))
           (read-gesture :stream stream))))
     (with-delimiter-gestures (*command-argument-delimiters* :override t)
-      (setq command-args (funcall (parser (gethash command-name
-                                                   *command-parser-table*))
-                                  stream)))
+      ;; The NULL context is a barrier which prevents other commands from
+      ;; being selected during reading arguments and cancelling the current
+      ;; command composition. When the argument's type is COMMAND, then a new
+      ;; context will be estabilished by the parser and it will be possible to
+      ;; select commands. -- jd 2020-09-03
+      (with-input-context ('null :override t) ()
+          (setq command-args (funcall (parser (gethash command-name
+                                                       *command-parser-table*))
+                                      stream))))
     (cons command-name command-args)))
 
 (defun command-line-command-unparser (command-table stream command)
--- a/Core/clim-core/commands/processor.lisp
+++ b/Core/clim-core/commands/processor.lisp
@@ -361,14 +361,14 @@
 
 (define-presentation-method accept
     ((type command) stream (view textual-view) &key)
-  (let ((command (funcall *command-parser* command-table stream)))
-    (cond ((null command)
-           (simple-parse-error "Empty command"))
-          ((partial-command-p command)
-           (funcall *partial-command-parser*
-                    command-table stream command
-                    (position *unsupplied-argument-marker* command)))
-          (t (values command type)))))
+  (if-let ((command (funcall *command-parser* command-table stream)))
+    (values (if (partial-command-p command)
+                (funcall *partial-command-parser*
+                         command-table stream command
+                         (position *unsupplied-argument-marker* command))
+                command)
+            type)
+    (simple-parse-error "Empty command")))
 
 (define-presentation-type command-or-form
     (&key (command-table (frame-command-table *application-frame*)))
--- a/Core/clim-core/presentations/typed-input.lisp
+++ b/Core/clim-core/presentations/typed-input.lisp
@@ -52,16 +52,14 @@
     (frame-input-context-track-pointer frame input-context sheet event)))
 
 (defmacro with-input-context ((type &key override)
-                              (&optional (object-var (gensym))
-                                 (type-var (gensym))
-                                 event-var
-                                 options-var)
+                              (&optional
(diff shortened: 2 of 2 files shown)


Core/clim-core/commands/processor.lisp
```diff
@@ -361,14 +361,14 @@
...
-                    (position *unsupplied-argument-marker* command)))
-          (t (values command type)))))
+  (if-let ((command (funcall *command-parser* command-table stream)))
+    (values (if (partial-command-p command)
+                (funcall *partial-command-parser*
+                         command-table stream command
+                         (position *unsupplied-argument-marker* command))
+                command)
+            type)
+    (simple-parse-error "Empty command")))
```
scymtym: Can we use the new `accept.smoke` test to ensure this requirement (that is returning two values)?

Core/clim-core/commands/processor.lisp
```diff
@@ -361,14 +361,31 @@
...
+                          command-table stream command
+                          (position *unsupplied-argument-marker* command))
+                 command)))
+      (with-input-context (type)
+          (object type)
+          (if-let ((command (funcall *command-parser* command-table stream)))
+            (values (maybe-replace-input (handle-command command)) type)
+            (simple-parse-error "Empty command"))
+        (command
+         (values (maybe-replace-input (handle-command object)) type))))))
```
scymtym, line 393: Can we use the `accept.smoke` test by feeding a suitable event to the fake extended input stream?

Core/clim-core/commands/processor.lisp
```diff
@@ -38,9 +38,14 @@
...
     (with-delimiter-gestures (*command-argument-delimiters* :override t)
-      (setq command-args (funcall (parser (gethash command-name
-                                                   *command-parser-table*))
-                                  stream)))
+      ;; The null context is a barrier which prevents other commands from
+      ;; being selected during reading arguments. -- jd 2020-09-03
+      (with-input-context ('null :override t) ()
+          (setq command-args (funcall (parser (gethash command-name
+                                                       *command-parser-table*))
+                                      stream))
```
scymtym, line 46: Will this also prevent commands from being selected as arguments? If not, the comment could be clarified a bit.

Core/clim-core/commands/processor.lisp
```diff
@@ -38,9 +38,14 @@
...
+        ;; Impossible clause.
+        (t (error "U Can't Touch This"))))
```
scymtym, line 48: Why is this needed? If it is actually reachable (or in any case, really), the error message should provide information and explain the situation.

scymtym (changes requested):
Questions and suggestions in commit comments.

Core/clim-core/commands/processor.lisp
dkochmanski: It is not reachable, if it is reached then we have a bug in input context processing. it is similar to "You lose" errors in other parts of the code.

Core/clim-core/commands/processor.lisp
dkochmanski: if an argument type is command, then it will be put as the context by the command parser. In other words barrier cancels all input contexts from above.

Core/clim-core/commands/processor.lisp
dkochmanski: do you mean something like that?


```
(test accept.smoke
  (mapc
   (lambda (case)
     (destructuring-bind (input accept-type expected-value expected-type) case
       (flet ((test-stream (stream)
                (multiple-value-bind (value type)
                    (accept accept-type :stream stream :prompt nil)
                  (is (equal expected-value value)
                      "~@<ACCEPTing ~S from a ~S with type ~S resulted ~
                       in value ~S, not ~S~@:>"
                      input (type-of stream) accept-type value expected-value)
                  (is (equal expected-value value)
                      "~@<ACCEPTing ~S from a ~S with type ~S resulted ~
                       in type ~S, not ~S~@:>"
                      input (type-of stream) accept-type type expected-type)
                  (is (not (null type))
                      "~@<ACCEPTing ~S from a ~S with type ~S resulted ~
                       in type ~s, which is not a valid."))))
         ;; Test with extended input stream.
         (let ((stream (make-instance 'test-stream)))
           (map nil (alexandria:curry #'climi::stream-append-gesture stream)
                input)
           (climi::stream-append-gesture stream #\Return)
           (test-stream stream))
         ;; Test with string stream.
         (let* ((string (format nil "~A~C" input #\Return))
                (stream (make-string-input-stream string)))
           (test-stream stream)))))
   '(("1"       integer            1         integer)
     ("1,2,3,4" (sequence integer) (1 2 3 4) (sequence integer)))))
```

or do you mean, that we should define a command table and a command, then try to parse it? that won't work for partial command parsers (and that's where we were bitten), because they invoke accepting-values dialog.

scymtym: If I understand the changed code correctly, we would have to add cases that replace feeding `"1,2,3,4"` to the stream by feeding a simulated presentation-click event to it.

Check the second return value can be done in all cases it seems.

Core/clim-core/commands/processor.lisp
scymtym: So maybe "The null context is a barrier which presents other commands from being selected and cancelling the current command while reading arguments. Selecting commands as arguments to the current command still works."?

Core/clim-core/commands/processor.lisp
scymtym: Maybe "An error an error in McCLIM caused code to be executed that should never be reached. Please report this as a bug."?

Core/clim-core/commands/processor.lisp
scymtym: Too difficult with the current test infrastructure.

Core/clim-core/commands/processor.lisp
scymtym: Too difficult with the current test infrastructure.

Core/clim-core/commands/processor.lisp
dkochmanski: removed the clause whatsoever.

Core/clim-core/commands/processor.lisp
dkochmanski: updated the comment to

```
(with-delimiter-gestures (*command-argument-delimiters* :override t)
      ;; The null context is a barrier which prevents other commands from
      ;; being selected during reading arguments and cancelling the current
      ;; command while reading arguments. When the argument's type is command
      ;; then this context will be estabilished by the parser and it will be
      ;; possible to select commands. -- jd 2020-09-03
      (with-input-context ('null :override t) ()
          (setq command-args (funcall (parser (gethash command-name
                                                       *command-parser-table*))
                                      stream))))
```

dkochmanski: updated again.

scymtym (approved):

scymtym:
Looks good to me.

PR #1067, opened 2020-09-04
Fix box-pane and rack layouts with +FILL+ and spacing

scymtym:
The two `box-layout-mixin/xically-allocate-space` methods lay out child panes sequentially, keeping track of the offset in the major dimension. This offset is increased according to the allotted widths of children as well as the spacing between children. Before this change, a `+fill+` client (which does not have a pane and thus does not need spacing) caused one additional unit of spacing to be added to the offset, potentially pushing the final "real" client out of the allocated region.

The following screenshot shows the issue when adding `:x-spacing 8` to the `horizontally` containing the push buttons in the file selector demo:
screenshot2


dkochmanski:
lgtm, thanks!

PR #1068, opened 2020-09-05
[wip] command parsers

dkochmanski:
This is still incomplete proposal (but it works fine except for the issue mentioned later in the thread) - changing the function `accept-1` and interpretation of what is the interactive stream requires more thought.

That said this PR introduces a functionality, which allows reading commands without interactor present  by implementing menu parsers. Moreover translators which return incomplete commands (that is commands with /less/ arguments than number of requried arguments) are coerced to a partial command hence may be further read (instead of erroring).


dkochmanski:
The attached example code may be used for testing. Notice, that when we select graph and node twice, drei will signal an error originating from handle-empty-input. That requries some attetion to read-command and what it should do (and down-the-road, why drei butts in here).

```
;;;;  Copyright (c) 2020 Daniel Kochmański
;;;;
;;;;    License: BSD-2-Clause.

(defpackage #:clim-demo.graph-demo
  (:use #:clim-lisp)
  (:export #:graph-demo #:run))
(in-package #:clim-demo.graph-demo)

(defclass node ()
  ((name :initarg :name :reader name)
   (children :initarg :children :reader children))
  (:default-initargs :children nil))

(defun make-node (&rest args)
  (apply #'make-instance 'node args))

(defun show-node (object stream)
  (clim:present object 'node :stream stream))

(defun display (frame stream)
  (clim:present frame 'graph-demo :stream stream :single-box t))

(defun run ()
  (clim:run-frame-top-level
   (clim:make-application-frame 'graph-demo)))

(clim:define-application-frame graph-demo ()
  ((root-nodes :initarg :root-nodes :reader root-nodes)
   (graph-args :initarg :graph-args :accessor graph-args))
  (:panes (app :application :display-function #'display
               :scroll-bars nil :borders nil)
          ;(int  :interactor)
          )
  (:layouts (default (clim:horizontally ()
               (clim:outlining (:thickness 10 :background clim:+white+)
                 app)
               ;int
               )))
  (:default-initargs :root-nodes
                     (let* ((2a (make-node :name "2A"))
                            (2b (make-node :name "2B"))
                            (2c (make-node :name "2C"))
                            (1a (make-node :name "1A" :children (list 2a 2b)))
                            (1b (make-node :name "1B" :children (list 2b 2c)))) 
                       (list (make-node :name "0" :children (list 1a 1b))))
                     :graph-args nil))

(clim:define-presentation-method clim:present
    (object (type node) stream view &key)
  (write-string (name object) stream))

(clim:define-presentation-method clim:present
    (object (type graph-demo) stream view &key)
  (apply #'clim:format-graph-from-roots (root-nodes object)
         #'show-node #'children :stream stream (graph-args object)))

(define-graph-demo-command com-change-orientation
    ((graph 'graph-demo :gesture :select)
     (node 'node))
  (format *debug-io* "~s vs ~s~%" graph node))
```

dkochmanski:
superseded in new-parsers-branch

PR #1070, opened 2020-09-16
stream-input: drop STRING-STREAM hack for SBCL and CMUCL

scymtym:
I verified with CMUCL 21D Unicode and SBCL 2.0.6 that `string-stream` is now a direct superclass of the respective string input stream class.


dkochmanski:
lgtm, thanks

PR #1072, opened 2020-09-18
Commands processor respects :ECHO option returned by translator

scymtym:
Before this change, the `accept` method for the `command` presentation type, when `maybe-replace-input` is invoked via a presentation click, did not check whether the translator returned the :echo nil option in its third argument.

Test program:

``` common-lisp
(clim:define-presentation-type thing ()
  :inherit-from t)

(clim:define-application-frame test ()
  ()
  (:pane
   (clim:vertically ()
     (clim:make-pane :application :display-function (lambda (frame pane)
                                                      (clim:present 1 'thing)))
     (clim:make-pane :interactor))))

(define-test-command com-doit
    ((thing thing :gesture (:select :echo ni))))

(clim:run-frame-top-level (clim:make-application-frame 'test))
```

### DIFF
--- a/Core/clim-core/commands/processor.lisp
+++ b/Core/clim-core/commands/processor.lisp
@@ -371,8 +371,8 @@
         (start-position (and (getf options :echo t)
                              (input-editing-stream-p stream)
                              (stream-scan-pointer stream))))
-    (flet ((maybe-replace-input (input)
-             (when start-position
+    (flet ((maybe-replace-input (input options)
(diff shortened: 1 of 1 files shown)


scymtym:
I think this can be merged. It is true that the interactor starts a new line and prints an additional prompt despite `:echo nil`, but that is a separate problem and almost certainly not a regression.

dkochmanski:
looks good to me. I've investigated the part with "empty" newlines but they are not related, so merging.

PR #1075, opened 2020-09-28
Fix graph formatting regression

scymtym:
* After the refactoring in 69188296e, `generate-graph-nodes` no longer
  stopped the traversal at already seen nodes.

* Less importantly, `generate-graph-nodes` called the `duplicate-key`
  function irregardless of whether the key was actually needed.

* Least importantly, `generate-graph-nodes` performed an unnecessary
  hash-table write for already seen nodes.


dkochmanski:
lgtm, thanks

PR #1079, opened 2020-10-01
Make clim stream pane

admich:


Core/clim-core/panes/stream-panes.lisp
```diff
@@ -410,16 +410,24 @@ current background message was set."))
                           nil)))
       (values pane stream))))
 
-(defun make-clim-interactor-pane (&rest options)
+(defun make-clim-interactor-pane (&rest options &key (scroll-bar t scroll-bar-p) (scroll-bars scroll-bar scroll-bars-p) &allow-other-keys)
+  (unless (or scroll-bar-p scroll-bars-p)
+    (setf (getf options :scroll-bars) :vertical))
   (apply #'make-clim-stream-pane :type 'interactor-pane options))
```
scymtym, line 418: This results in a warning because `scroll-bars` is not used.

I also find a non-mutating version easier to understand:
```lisp
(apply #'make-clim-stream-pane :type 'interactor-pane
       (if (or scroll-bar-p scroll-bars-p)
           options
           (list* :scroll-bars :vertical options)))
```
But that may just be my taste.

Core/clim-core/panes/stream-panes.lisp
```diff
@@ -79,6 +77,15 @@
...
     extended input and output, and output recording protocols."))
 
+(defmethod initialize-instance :after ((instance clim-stream-pane) &rest initargs)
+  (declare (ignore initargs))
+  (with-slots (scroll-bars label) instance
+    (when (or (not (eql :obsolete scroll-bars))
+              (not (eql :obsolete label)))
+      (warn "LABEL and SCOLL-BARS slots in CLIM-STREAM-PANE are
+      obsolete, don't use them but use the keywords :LABEL and
+      :SCROLL-BARS in function MAKE-CLIM-STREAM-PANE"))))
```
scymtym, line 87: Please split this into two warnings so it is clear which one was supplied.

Core/clim-core/panes/composition.lisp
```diff
@@ -1002,12 +1002,6 @@
 (defun spacing-p (pane)
   (typep pane 'spacing-pane))
 
-(defmethod initialize-instance :after ((spacing spacing-pane) &key thickness contents &allow-other-keys)
-  (declare (ignorable thickness contents))
-  (with-slots (user-width user-min-width user-max-width
-               user-height user-min-height user-max-height)
-      spacing))
-
```
scymtym, line 1010: Please describe in the commit message which methods are removed. There are lots of dumb methods in the code base :)

Core/clim-core/panes/stream-panes.lisp
```diff
@@ -365,49 +365,50 @@ current background message was set."))
 ;;;
 ;;; CONSTRUCTORS
 ;;;
-
 (defun make-clim-stream-pane (&rest options
                               &key (type 'clim-stream-pane)
+                                label
+                                label-alignment
+                                (display-after-commands nil display-after-commands-p)
```
scymtym, line 374: Please add `(check-type display-after-commands (member nil t :no-clear))` in case the option is supplied.

scymtym, line 374: Should an error be signaled if both `display-after-commands` and `display-time` are supplied?

Core/clim-core/panes/stream-panes.lisp
```diff
@@ -365,49 +365,50 @@ current background message was set."))
...
-                            :contents (list pane)
-                            user-sr)))
-        (values pane stream)))))
+  (with-keywords-removed (options (:type :scroll-bar :scroll-bars :borders :label
+                                         :label-alignment :display-after-commands))
+    ;; See Spec 29.4.3 for :display-after-commands keyword.
+    (when (and display-after-commands-p (not (member :display-time options)))
+      (if (eq t display-after-commands)
+          (nconc options (list :display-time :command-loop))
+          (nconc options (list :display-time display-after-commands))))
```
scymtym, line 385: The fact that the result of `nconc` isn't used seems wrong.

How about this?
```lisp
(let* ((other-options (alexandria:remove-from-plist options :type …))
       ;; See Spec 29.4.3 for :display-after-commands keyword.
       (final-options (if (or (member :display-time options) (not display-after-commands-p))
                          other-options
                          (list* :display-time
                                 (if (eq display-after-commands t) 
                                     :command-loop 
                                     display-after-commands)
                                 other-options))
   …)

File: Core/clim-core/panes/stream-panes.lisp [hunk 6]
```diff
@@ -365,49 +365,50 @@ current background message was set."))
...
+                          ;; argument and the cdr is a list of keyword
+                          ;; argument pairs to be used as options of
+                          ;; the scroller-pane
+                          (if (consp scroll-bars)
+                              `(:scroll-bar ,@scroll-bars)
+                              `(:scroll-bar ,scroll-bars)))))
+      (when label
+        (setq pane (apply #'make-pane 'label-pane
+                          :label label
+                          :label-alignment label-alignment
```
scymtym, line 403: This could pass `:label-alignment nil` which seems forbidden:

> … label-alignment may be either :bottom or :top, …

scymtym (changes requested):
Remarks and requested changes are in individual commit comments.

Core/clim-core/panes/stream-panes.lisp
admich: done

Core/clim-core/panes/stream-panes.lisp
admich: done

Core/clim-core/panes/composition.lisp
admich: Done and I alse remove an exported symbol of an old function that was removed from the code.

Core/clim-core/panes/stream-panes.lisp
admich: done

admich: I have added the error

Core/clim-core/panes/stream-panes.lisp
admich: I have used your code with the difference that because now an error is signaled if both `:display-after-commands` and `:display-time` are supplied I can trust that if `display-after-command-p` is true than there isn't `:display-time` in the options

Core/clim-core/panes/stream-panes.lisp
admich: I have added label-alignment-p  to test if the keyword is supplied.

Core/clim-core/panes/composition.lisp
scymtym: Sorry, I mainly meant in the first line of the commit message. Currently the commit appears as "Cleanup code" in a compact history listing which is not informative. If you want, I can rebase and fix these cosmetic things.

Core/clim-core/panes/stream-panes.lisp
```diff
@@ -79,6 +77,18 @@
...
+(defmethod initialize-instance :after ((instance clim-stream-pane) &rest initargs)
+  (declare (ignore initargs))
+  (with-slots (scroll-bars label) instance
+    (when (not (eql :obsolete scroll-bars))
+      (warn "SCOLL-BARS slot in CLIM-STREAM-PANE is obsolete, don't
+      use it but use the keyword :SCROLL-BARS in function
+      MAKE-CLIM-STREAM-PANE"))
+    (when (not (eql :obsolete label))
+      (warn "LABEL slot in CLIM-STREAM-PANE is obsolete, don't use it
+      but use the keywords :LABEL in function
```
scymtym, line 89: the keyword~s~ :LABEL

Core/clim-core/panes/stream-panes.lisp
```diff
@@ -368,61 +370,88 @@ current background message was set."))
...
+      (setq pane (apply #'make-pane 'outlined-pane
+                        :thickness (if (not (numberp borders))
+                                       1
+                                       borders)
+                        :contents (list pane)
+                        nil)))
+    (values pane stream)))
+
+(defun make-clim-interactor-pane (&rest options &key (scroll-bar t scroll-bar-p) (scroll-bars scroll-bar scroll-bars-p) &allow-other-keys)
+  (declare (ignorable scroll-bars))
```
scymtym, line 429: Please use `ignore` instead of `ignorable`. The latter is for macros which include unknown code in the expansion that might or might not use a given variable.

Core/clim-core/panes/composition.lisp
scymtym: Also we, usually prefix the commit message with the changed function (e.g. `formatting-table: validate keyword arguments`), file (e.g. `stream-input: untabify`) or subsystem (e.g. `clx: fix mirror class name computation`).

scymtym (changes requested):
More minor requested changes in commit comments.

admich:
For me it is ok if you rebase and fix this cosmetic things by yourself. I think it is less work for me but also for you because otherwise you need to check again my (maybe not perfect) modification.

scymtym:
I have made the cosmetic changes and rebased.

But I have discovered a problem: the stopwatch demo tries to configure its interactor pane with `:height 100`. This worked previously because the `border-pane` surrounding the scroller, viewport and interactor "got" the user height 100 and realized it correctly. Now, the interactor pane itself "gets" the user height 100 and stores it in its `user-height` slot, but the space requirements and later the assigned region do not reflect that (see screenshot).
screenshot

scymtym:
@admich do you have any ideas for fixing the above layout problem?

admich:
Yes, the problem it is that I change the meaning of the `:height` and `:width`.  Before they are passed to the external container of the stream, in my PR they are always passed to the stream. I think this is better and it is what clim-tos does (my idea is that `:width` and `:height` are the width and height of the stream). If a client want set the width  and hight of the container it can use `:scroll-bars` keyword. For example in stopwatch example define the commands interactor pane as:
```
(commands :interactor :scroll-bars '(:vertical :height 100))
```
So we have two possibilities:

1) use my version of `make-clim-stream-pane` and change all the client like stopwatch example.
2) revert the old mcclim behaviour and pass width and hight also to the scoller.

Ofcourse with choice 1) some external application could have some layout problem if they are written with the old mcclim behaviour in mind.

I think its better also to ask @dkochmanski what he thinks it's better.

dkochmanski:
imo when the user passes (application-pane :scroll-bars t :width 100 :height 100) then whole thing should take 100x100 (so the old behavior is better in my eyes, because it is least surprising).

admich:
but in this way when there are scroll-bar you can't configure the dimension of the stream like:
`(application-pane :application :width 800 :height 400 :scroll-bars '(t :width 100 :height 100))`
(ofcourse you can using scroll-bars nil and then scrolling macro in the layout)

Anyway if you both agree I revert the old behavior. @scymtym have you put your changing/rebase in same branch that I can modify?

admich:
If you agree  I will try to modify the function in such way that if :scroll-bars is a cons we have the new behavior otherwise we have the old behavior

dkochmanski:
I think that this would be quite irregular and may lead to a headache when debugging in the future. If you do so though, then add a lengthy comment in the source code

scymtym:
>  @scymtym have you put your changing/rebase in same branch that I can modify?

In the `fix-stream-pane` branch in this repository.

scymtym (approved):

scymtym:
Looks good to me. Merged. Thank you.

PR #1081, opened 2020-10-06
drei: stream-process-gesture: ignore pointer events

### LINKED ISSUE #1080: Presentation to command translators sometimes don't execute the command
When an interactor pane is present (the problem may also occur without that, but the interactor pane makes it easier to observe), clicking a presentation for which a presentation to command translator is applicable (and has the highest priority, etc.), the command is sometimes echoed in the interactor and executed. In some cases however, the command is echoed in the interactor (or inserted for input editing to be more precise) but not executed.

An easy way to reproduce the issue is the tab layout demo. Clicking a tab header only sometimes switches to the tab as described above.

Based on bisection, the bad commit seems to be ec2225c06 ("accept command: provide input context to click on a command").

dkochmanski:
This method is used to process input editing commands and is not prepared to
work with pointer events. Most notably esa-read-gesture which is called from
read-gestures-and-act does not recognize them as a proper gesture and waits
until some key is pressed (indefinetely). Fixes #1080.

### DIFF
--- a/Libraries/Drei/input-editor.lisp
+++ b/Libraries/Drei/input-editor.lisp
@@ -616,6 +616,10 @@ if stuff is inserted after the insertion pointer."
 
 (defmethod stream-process-gesture ((stream drei-input-editing-mixin)
 				   gesture type)
+  ;; Drei is not interested in the pointer gestures.
+  (when (typep gesture 'pointer-event)
+    (return-from stream-process-gesture
+      (values nil 'null)))
   ;; If some other command processor has taken control, we do not want
   ;; to assume that an activation gesture really is an activation
   ;; gesture. For example, #\Newline should not cause input activation
--- a/Libraries/Drei/drei-clim.lisp
+++ b/Libraries/Drei/drei-clim.lisp
@@ -312,6 +312,10 @@ modifier key."))
       (display-drei drei :redisplay-minibuffer t)
       (propagate-changed-value drei))))
 
+;;; We don't want drei gadgets to append gestures to the input-buffer.
+(defmethod climi::stream-append-gesture ((stream drei-gadget-pane) gesture)
+  (declare (ignore stream gesture)))
+
 ;;; This is the method that functions as the entry point for all Drei
 ;;; gadget logic.
 (defmethod handle-event ((gadget drei-gadget-pane) (event key-press-event))


dkochmanski:
I've pushed a commit that fixes another regression with text fields.

scymtym (changes requested):
Requested changes maybe in commit comments. GitHub confused itself a bit during the review.

Core/clim-basic/extended-streams/stream-input.lisp
```diff
@@ -340,11 +340,11 @@ keys read."))
...
 
-(defmethod handle-event :after
+(defmethod handle-event
     ((client standard-extended-input-stream) (event key-press-event))
   (stream-append-gesture client event))
 
-(defmethod handle-event :after
+(defmethod handle-event
     ((client standard-extended-input-stream) (event pointer-event))
   (stream-append-gesture client event))
```
scymtym, line 349: Doesn't this require subclasses to `(call-next-method)` where previously they did not have to?

scymtym (changes requested):
More comments.

Core/clim-basic/extended-streams/stream-input.lisp
dkochmanski: Only when they define their own handle-event method on the pointer-event or key-press-event. this does not occur, and for clients outside McCLIM that would be mixing two different abstractions.

Libraries/Drei/input-editor.lisp
```diff
@@ -616,6 +616,9 @@ if stuff is inserted after the insertion pointer."
 
 (defmethod stream-process-gesture ((stream drei-input-editing-mixin)
 				   gesture type)
+  ;; Drei is not interested in the pointer gestures.
+  (when (typep gesture 'pointer-event)
+    (return-from stream-process-gesture nil 'null))
```
scymtym, line 621: I thought, generally this was supposed to return `gesture` for further processing. But maybe Drei does things differently.

dkochmanski: as I've mentioned in the comment, drei assumes that only key press events end up here, that's what caused the issue. this is basically the processing drei "wants", that is to receive only keyboard events.

Core/clim-basic/extended-streams/stream-input.lisp
scymtym: `clim-demo.drawing-tests::drawing-app-pane` and `drei:drei-gadget-pane` have such methods. Should those be changed to call the next method or are they problematic even before this change due to abstraction mixing?

dkochmanski: mixing abstractions is fine, but it requires additional care, that's what I've meant. A casual CLIM programmer doesn't have to do that. adding cal-next-method in them should be fine. I'll pick up from here tomorrow (afk).

dkochmanski: fixed

dkochmanski:
I've pushed a commit which scraps the last one (which removed the `:after` qualifier from the handle-event defined on seis), and proposes more narrow solution, that is defining stream-append-gesture method for drei-gadget-pane to not queue its events to the input buffer.

scymtym (approved):

PR #1082, opened 2020-10-08
frames: fix redisplay problem with execute-frame-command

ebrasca:


dkochmanski:
as noted on IRC, this fix is copied from: https://github.com/McCLIM/McCLIM/blob/master/Libraries/ESA/esa.lisp#L840

as I can tell, example frame inherits from both esa-frame-mixin and standard-application-frame -- wouldn't that cause doble redisplay of frame which inherits from both standard-frame and esa?

dkochmanski:
this is incomplete solution, closing.

PR #1084, opened 2020-10-13
Replace sb-mop:class-direct-subclasses by closer-mop:class-direct-subclasses

kpoeck:
* so that it works for other lisps than sbcl


dkochmanski (approved):
thanks!

PR #1085, opened 2020-10-14
Use macro to define MAKE-CLIM-*-PANE and cosmetic changes

scymtym:
In the generated `make-clim-*-pane` functions, remove default value of keyword parameters since they were never passed to `MAKE-CLIM-STREAM-PANE`.


dkochmanski:
lgtm, thanks

dkochmanski:
oops, sorry, just now noticed, that a review from @admich was requested..

PR #1086, opened 2020-10-16
Pay attention to the :echo option to with-input-context

heshrobe:
This fixes a minor bug in /Core/clim-core/commands/processor.lisp in the accept presentation method for command.

The method uses with-input-context, but doesn't bind options nor check whether :echo nil is present in options.  If :echo nil is specified (usually from a presentation-translator) then the input should not be placed into the stream.  The fix is shown below (eliding the non-relevant part):

 (flet ((maybe-replace-input (input options)
             (when start-position
               ;; Rescan is nil in case the clicked on thing returned
               ;; something unparseable. -- heshrobe 2020-05-15
               (when (getf options :echo t)
                 (presentation-replace-input stream input type view
                                             :buffer-start start-position
                                             :rescan nil)))
             input)
             ...
      (with-input-context (type)
          (object type event options)
          (if-let ((command (funcall *command-parser* command-table stream)))
            (values (maybe-replace-input (handle-command command) nil) type)
            (simple-parse-error "Empty command"))
        (command
         (values (maybe-replace-input (handle-command object) options) type))))))

### DIFF
--- a/Core/clim-core/commands/processor.lisp
+++ b/Core/clim-core/commands/processor.lisp
@@ -359,11 +359,34 @@
       (with-presentation-type-parameters (command type)
         (command-table-inherits-from-p super-table command-table)))))
 
-(define-presentation-method present
-    (object (type command) stream (view textual-view)
-     &key acceptably for-context-type)
-  (declare (ignore acceptably for-context-type))
-  (funcall *command-unparser* command-table stream object))
+(define-presentation-method accept
+    ((type command) stream (view textual-view) &key)
+  (let ((command-table (find-command-table command-table))
+        (start-position (and (getf options :echo t)
+                             (input-editing-stream-p stream)
+                             (stream-scan-pointer stream))))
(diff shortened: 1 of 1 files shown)


scymtym:
I believe #1072 already addresses this issue.

heshrobe:
Jan,

You’re correct, sorry for not noticing that. 

> On Oct 16, 2020, at 14:35, Jan Moringen <notifications@github.com> wrote:
> 
> 
> I believe #1072 already addresses this issue.
> 
> —
> 

Howard Shrobe

Director CyberSecurity@CSAIL
hes@csail.mit.edu
617-253-7877 work
617-312-6439 cell

scymtym:
> You’re correct, sorry for not noticing that.

No worries. Thanks for helping with open issues.

dkochmanski:
I'm closing this duplicated PR. For the record, in this PR the `present` method was removed instead of the (next top level form) `accept` method.

PR #1087, opened 2020-10-17
command table menu rewrite

dkochmanski:
This PR addresses number of issues encountered when working with menus and keystroke accelerators. Part of said issues are a result of inconsistent specification - it is also modified. The "commands" test suite is updated accordingly (with other fixes along the way).

### DIFF
--- a/Core/clim-core/commands/processor.lisp
+++ b/Core/clim-core/commands/processor.lisp
@@ -22,7 +22,9 @@
                                          &body body)
   (with-gensyms (table)
     `(let* ((,table (find-command-table ,command-table))
-            (,keystroke-var (slot-value ,table 'keystroke-accelerators)))
+            (,keystroke-var (loop for item in (slot-value ,table 'menu)
+                                  for acc = (command-menu-item-keystroke item)
+                                  when acc collect acc)))
        ,@body)))
 
 (defun command-line-command-parser (command-table stream)
--- a/Core/clim-core/commands/processor.lisp
+++ b/Core/clim-core/commands/processor.lisp
@@ -22,9 +22,15 @@
                                          &body body)
   (with-gensyms (table)
     `(let* ((,table (find-command-table ,command-table))
-            (,keystroke-var (loop for item in (slot-value ,table 'menu)
-                                  for acc = (command-menu-item-keystroke item)
-                                  when acc collect acc)))
+            (,keystroke-var
+              (let ((result '()))
+                (map-over-command-table-keystrokes*
+                 (lambda (item table)
+                   (declare (ignore table))
+                   (let ((keystroke (command-menu-item-keystroke item)))
+                     (setf result (adjoin keystroke result :test #'equal))))
+                 ,table)
+                result)))
        ,@body)))
 
 (defun command-line-command-parser (command-table stream)
@@ -100,19 +106,6 @@
 (defvar *partial-command-parser*
   #'command-line-read-remaining-arguments-for-partial-command)
 
-;; This is probably wrong - we should walk the menu rather than inheritance
-;; structure to be consistent with lookup-keystroke-item.
-(defun compute-inherited-keystrokes (command-table)
-  "Return a list containing the keyboard gestures of accelerators defined in
- 'command-table' and all tables it inherits from."
-  (let (accumulated-keystrokes)
-    (do-command-table-inheritance (comtab command-table)
-      (with-command-table-keystrokes (keystrokes comtab)
-        (dolist (keystroke keystrokes)
-          (setf accumulated-keystrokes
-                (adjoin keystroke accumulated-keystrokes :test #'equal)))))
-    accumulated-keystrokes))
-
 (defun read-command (command-table
                      &key (stream *standard-input*)
                           (command-parser *command-parser*)
@@ -124,10 +117,10 @@
         (*partial-command-parser* partial-command-parser))
     (cond (use-keystrokes
            (let ((stroke-result
-                   (read-command-using-keystrokes
-                    command-table
-                    (compute-inherited-keystrokes command-table)
-                    :stream stream)))
+                   (with-command-table-keystrokes (keystrokes command-table)
+                     (read-command-using-keystrokes command-table
+                                                    keystrokes
+                                                    :stream stream))))
              (if (consp stroke-result)
                  stroke-result
                  nil)))
--- a/Core/clim-core/commands/tables.lisp
+++ b/Core/clim-core/commands/tables.lisp
@@ -133,10 +133,11 @@
 
 (defun apply-with-command-table-inheritance (fun command-table)
   (funcall fun command-table)
-  (mapc #'(lambda (inherited-command-table)
-            (apply-with-command-table-inheritance
-             fun (find-command-table inherited-command-table)))
-        (command-table-inherit-from command-table)))
+  (map nil
+       (lambda (inherited-command-table)
+         (apply-with-command-table-inheritance
+          fun (find-command-table inherited-command-table)))
+       (command-table-inherit-from command-table)))
 
 ;;; do-command-table-inheritance has been shipped off to utils.lisp.
 
@@ -346,7 +347,8 @@ designator) inherits menu items."
             (remhash (command-item-name item) (command-line-names table)))
           (remhash command-name commands)))))
 
-(defun map-over-command-table-menu-items (function command-table)
+(defun map-over-command-table-menu-items
+    (function command-table &key (inherited t))
   "Applies function to all of the items in `command-table's
 menu. `Command-table' must be a command table or the name of a
 command table. `Function' must be a function of three arguments,
@@ -371,7 +373,7 @@ examine the type of the command menu item to see if it is
                                   item)))
                    (slot-value table 'menu))))
       (map-table-entries table-object)
-      (when (inherit-menu-items table-object)
+      (when (and inherited (inherit-menu-items table-object))
         (dolist (table (command-table-inherit-from table-object))
           (map-over-command-table-menu-items function table))))
     (values)))
@@ -420,7 +422,9 @@ examine the type of the command menu item to see if it is
                              :key #'command-menu-item-name
                              :test #'string-equal))))))
 
-(defun map-over-command-table-keystrokes (function command-table)
+(defun map-over-command-table-keystrokes
+    (function command-table &key (inherited t))
+  (declare (ignore inherited))
   (let ((command-table (find-command-table command-table)))
     (with-slots (keystroke-accelerators keystroke-items)
         command-table
--- a/Core/clim-core/commands/tables.lisp
+++ b/Core/clim-core/commands/tables.lisp
@@ -58,18 +58,29 @@
   ((command-table-name :reader error-command-table-name
                        :initform nil
                        :initarg :command-table-name))
+  (:report (lambda (object stream)
+             (format stream
+                     "Command table name: ~s~%"
+                     (error-command-table-name object))
+             (apply #'format stream
+                    (simple-condition-format-control object)
+                    (simple-condition-format-arguments object))))
   (:default-initargs :format-control "" :format-arguments nil))
 
-(defmethod print-object ((object command-table-error) stream)
-  (print-unreadable-object (object stream :type t :identity t)
-    (when (error-command-table-name object)
-      (princ (error-command-table-name object) stream))))
+(define-condition command-table-not-found (command-table-error) ()
+  (:default-initargs :format-control "Command table not found."))
 
-(define-condition command-table-not-found (command-table-error) ())
-(define-condition command-table-already-exists (command-table-error) ())
-(define-condition command-not-present (command-table-error) ())
-(define-condition command-not-accessible (command-table-error) ())
-(define-condition command-already-present (command-table-error) ())
+(define-condition command-table-already-exists (command-table-error) ()
+  (:default-initargs :format-control "Command table already exists."))
+
+(define-condition command-not-present (command-table-error) ()
+  (:default-initargs :format-control "Command not present."))
+
+(define-condition command-not-accessible (command-table-error) ()
+  (:default-initargs :format-control "Command not accessible."))
+
+(define-condition command-already-present (command-table-error) ()
+  (:default-initargs :format-control "Command already accessible."))
 
 (defun command-table-designator-as-name (designator)
   "Return the name of `designator' if it is a command table,
--- a/Core/clim-core/commands/tables.lisp
+++ b/Core/clim-core/commands/tables.lisp
@@ -420,6 +420,24 @@ menu item to see if it is `:menu'."
      #'fun table :inherited (and inherited (inherit-keystrokes table))))
   (values))
 
+;;; Unlike map-over-command-table-keystrokes this function descends into
+;;; sub-menus to look for keystrokes. The function accepts (item table).
+(defun map-over-command-table-keystrokes* (function table &key (inherited t))
+  (flet ((map-items (item table)
+           (when-let ((keystroke (command-menu-item-keystroke item)))
+             (funcall function item table)))
+         (map-menus (item table)
+           (declare (ignore table))
+           (when (and (null (command-menu-item-keystroke item))
+                      (eq (command-menu-item-type item) :menu))
+             (let ((table (command-menu-item-value item)))
+               (map-over-command-table-keystrokes* function table
+                                                   :inherited t)))))
+    (let ((inherit (and inherited (inherit-keystrokes table))))
+      (map-over-command-table-menu #'map-items table :inherited inherit)
+      (map-over-command-table-menu #'map-menus table :inherited inherit))
+    (values)))
+
 (defun find-keystroke-item (gesture table
                             &key (test #'event-matches-gesture-name-p)
                               (errorp t))
@@ -435,23 +453,12 @@ menu item to see if it is `:menu'."
 
 (defun lookup-keystroke-item (gesture table
                               &key (test #'event-matches-gesture-name-p))
-  ;; We call FIND-KEYSTROKE-ITEM before descending into sub-menus to ensure
-  ;; that top-level keystrokes are found before nested ones.
-  (multiple-value-bind (item table)
-      (find-keystroke-item gesture table :test test :errorp nil)
-    (when item
-      (return-from lookup-keystroke-item (values item table))))
-  (flet ((fun (item table)
-           (declare (ignore table))
-           (when (and (null (command-menu-item-keystroke item))
-                      (eq (command-menu-item-type item) :menu))
-             (multiple-value-bind (item table)
-                 (lookup-keystroke-item gesture (command-menu-item-value item)
-                                        :test test)
-               (when item
-                 (return-from lookup-keystroke-item (values item table)))))))
-    (map-over-command-table-menu
-     #'fun table :inherited (inherit-keystrokes table))))
+  (map-over-command-table-keystrokes*
+   (lambda (item table)
+     (when-let ((keystroke (command-menu-item-keystroke item)))
+       (when (funcall test gesture keystroke)
+         (return-from lookup-keystroke-item (values item table)))))
+   table :inherited t))
 
 (defun add-keystroke-to-command-table (command-table gesture type value
                                        &key documentation (errorp t))
--- a/Core/clim-core/commands/tables.lisp
+++ b/Core/clim-core/commands/tables.lisp
@@ -328,54 +328,53 @@ designator) inherits menu items."
             (remhash (command-item-name item) (command-line-names table)))
           (remhash command-name commands)))))
 
-(defun map-over-command-table-menu-items
-    (function command-table &key (inherited t))
-  "Applies function to all of the items in `command-table's
-menu. `Command-table' must be a command table or the name of a
-command table. `Function' must be a function of three arguments,
-the menu name, the keystroke accelerator gesture (which will be
-NIL if there is none), and the command menu item; it has dynamic
-extent. The command menu items are mapped over in the order
-specified by `add-menu-item-to-command-table'. `Command-table' is
-a command table designator. Any inherited menu items will be
-mapped over after `command-table's own menu items.
+;;; This internal function is like map-over-command-menu-items, but it maps
+;;; over each element in the command table and (if inherited is t) its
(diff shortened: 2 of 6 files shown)


Documentation/Specification/commands.tex
```diff
@@ -743,13 +758,17 @@
...
 a gesture object, and is handled in the same way as in \cl{find-keystroke-item}.
+
 \ReadOnly
 
 Unlike \cl{find-keystroke-item}, this follows the sub-menu chains that can be
-created with \cl{add-menu-item-to-command-table}.  If the keystroke accelerator
-cannot be found in the command table or any of the command tables from which it
-inherits, \cl{lookup-keystroke-item} will return \cl{nil}.  \arg{command-table}
-is a \term{command table designator}.
+created with \cl{add-menu-item-to-command-table}. If the sub-menu has an
```
scymtym, line 765: Maybe "If a given sub-menu" instead of "If the sub-menu"?

Documentation/Specification/commands.tex
```diff
@@ -743,13 +758,17 @@
...
+associated keystroke that is not \arg{gesture}, the function does not look in
+this command table keystroke accelerators table.
```
scymtym, line 767: Does this mean "in the keystroke accelerators table associated with the sub-menu"?

Core/clim-core/commands/processor.lisp
```diff
@@ -167,6 +169,17 @@
...
+        ((accelerator-gesture
+           (lambda (c &aux (event (accelerator-gesture-event c)))
+             (multiple-value-bind (item item-table)
+                 (lookup-keystroke-item event command-table)
+               (format *debug-io* "~a (in ~a) -- ~a~%"
+                       item;(command-menu-item-type item)
+                       item-table
+                       command-table)))))
+      (read-command command-table :stream stream))
+    #+ (or)
```
dkochmanski, line 175: This should be removed from this PR.

Core/clim-core/commands/tables.lisp
```diff
@@ -221,41 +220,27 @@ designator) inherits menu items."
...
+                      ((null y) nil)
+                      (t (string-lessp x y)))))
+         (setf menu (sort (cons item menu) #'menu-name-lessp
+                          :key #'command-menu-item-name))))
+      (t
+       (check-type after string)
+       (push item
+             (cdr (member after menu
+                          :key #'command-menu-item-name
+                          :test #'string-equal)))))))
```
scymtym, line 243: `(string-equal "NIL" nil) => T` so maybe `menu-name-equal` is needed?

Core/clim-core/commands/tables.lisp
```diff
@@ -377,17 +359,21 @@ examine the type of the command menu item to see if it is
...
-                         (find-menu-item menu-name table :errorp nil))
-                     (command-table-inherit-from table))
-            (when errorp
-              (error 'command-not-accessible :command-table-name
-                     (command-table-designator-as-name table)))))))
+  (loop with table = (find-command-table command-table)
+        for item in (slot-value table 'menu)
+        when (string-equal menu-name (command-menu-item-name item))
+          do (return-from find-menu-item
+               (values item command-table)))
```
scymtym: Why `loop` instead of `when-let ((… (find …`?

Core/clim-core/commands/tables.lisp
```diff
@@ -407,77 +393,76 @@ examine the type of the command menu item to see if it is
...
-
-(defun find-keystroke-item (gesture command-table
+  (let ((table-object (find-command-table command-table)))
+    (flet ((map-table-entries (table)
+             (mapc #'(lambda (item)
+                       (with-slots (menu-name keystroke) item
+                         (when keystroke
+                           (funcall function menu-name keystroke item))))
+                   (slot-value table 'menu))))
+      (map-table-entries table-object)
```
scymtym: Why is this a local function? It is called immediately and only at one point.

Core/clim-core/commands/tables.lisp
```diff
@@ -407,77 +393,76 @@ examine the type of the command menu item to see if it is
...
-    (if errorp
-        (error 'command-not-present :command-table-name
-               (command-table-designator-as-name command-table))
-        nil)))
+                              (errorp t))
+  (setf table (find-command-table table))
+  (loop for item in (slot-value table 'menu)
+        when (funcall test gesture (command-menu-item-keystroke item))
+          do (return-from find-keystroke-item
+               (values item table)))
```
scymtym: Why `loop` instead of `find`?

Core/clim-core/commands/tables.lisp
```diff
@@ -326,54 +326,54 @@ designator) inherits menu items."
...
+;;; ancestors. It does not descend into sub-menus. The mapped function accepts
+;;; two arguments: the item and the command table the item belongs to.
+(defun map-over-command-table-menu (function table &key (inherited t))
+  (setf table (find-command-table table))
+  (flet ((map-table-entries (table)
+           (mapc #'(lambda (item)
+                     (with-slots (menu-name keystroke type) item
+                       (funcall function item table)))
+                 (slot-value table 'menu))))
+    (map-table-entries table)
```
scymtym, line 340: Again, seems like an immediately-only-once-called local function.

Core/clim-core/commands/tables.lisp
```diff
@@ -326,54 +326,54 @@ designator) inherits menu items."
...
+command table or the name of a command table. `Function' must be a function of
+three arguments, the menu name (which will be NIL if there is none), the
+keystroke accelerator gesture (which will be NIL if there is none), and the
+command menu item; it has dynamic extent. The command menu items are mapped
+over in the order specified by `add-menu-item-to-command-table'. `table' is a
+command table designator. Any inherited menu items will be mapped over after
+`command-table's own menu items.
+
+`Map-over-command-table-menu-items' does not descend into sub-menus. If the
+programmer requires this behavior, he should examine the type of the command
```
scymtym, line 356: Can we change "he" to "they" here?

Core/clim-core/commands/tables.lisp
```diff
@@ -326,54 +326,54 @@ designator) inherits menu items."
...
+             (when (or menu-name (eq type :divider))
+               (funcall function menu-name keystroke item)))))
+    (map-over-command-table-menu
+     #'fun table :inherited (and inherited (inherit-menu-items table))))
+  (values))
+
+(defun find-menu-item (menu-name table &key (errorp t))
+  (flet ((fun (item table)
+           (when-let ((item-name (command-menu-item-name item)))
+             (when (string-equal item-name menu-name)
```
scymtym, line 370: Isn't this just `(when (equapl menu-name (command-menu-item-name item)) …` since `nil` is not `equalp` to any string?

Core/clim-core/commands/tables.lisp
```diff
@@ -326,54 +326,54 @@ designator) inherits menu items."
...
+  (flet ((fun (item table)
+           (declare (ignore table))
+           (with-slots (menu-name keystroke type) item
+             (when (or menu-name (eq type :divider))
+               (funcall function menu-name keystroke item)))))
+    (map-over-command-table-menu
+     #'fun table :inherited (and inherited (inherit-menu-items table))))
+  (values))
+
+(defun find-menu-item (menu-name table &key (errorp t))
```
scymtym, line 367: Please keep the parameter name `command-table` to match the specification.

Core/clim-core/commands/tables.lisp
```diff
@@ -436,21 +454,12 @@ menu item to see if it is `:menu'."
                               &key (test #'event-matches-gesture-name-p))
   ;; We call FIND-KEYSTROKE-ITEM before descending into sub-menus to ensure
   ;; that top-level keystrokes are found before nested ones.
```
scymtym: The comment no longer applies.

scymtym (changes requested):
Remarks and requested changes are in individual commit comments.

Core/clim-core/commands/tables.lisp
```diff
@@ -418,7 +420,9 @@ examine the type of the command menu item to see if it is
                              :key #'command-menu-item-name
                              :test #'string-equal))))))
 
-(defun map-over-command-table-keystrokes (function command-table)
+(defun map-over-command-table-keystrokes
+    (function command-table &key (inherited t))
+  (declare (ignore inherited))
```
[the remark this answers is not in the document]
dkochmanski: First I've unified the interface for tests, then implemented it -- in this commit this variable is not (yet) used.

Core/clim-core/commands/tables.lisp
```diff
@@ -344,7 +345,8 @@ designator) inherits menu items."
             (remhash (command-item-name item) (command-line-names table)))
           (remhash command-name (commands table))))))
 
-(defun map-over-command-table-menu-items (function command-table)
+(defun map-over-command-table-menu-items
```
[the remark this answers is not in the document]
dkochmanski: For consistency with other table mapping operators
- map-over-command-table-names
- map-over-command-commands
- map-over-command-table-translators

adding this keyword costs us nothing and it is backward compatible. it allows to access keystrokes and menu items that are specific to this command table (not inherited), or access all including inherited entries.

Tests/commands.lisp
```diff
@@ -60,6 +60,61 @@
...
+           (map-over-command-table-menu-items
+            (lambda (&rest args)
+              (declare (ignore args))
+              (fail "Menu item found (it shouldn't be)."))
+            ct)
+           (block nil
+             (map-over-command-table-keystrokes
+              (lambda (&rest args)
+                (declare (ignore args))
+                (return (pass)))
```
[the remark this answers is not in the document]
dkochmanski: Initial (not commited) version of tests did check as much as possible, however, that exploded in size very quickly, that's why I have decided to check only for specific things -- in this case that the item is found.

Core/clim-core/commands/processor.lisp
dkochmanski: fixed

Tests/commands.lisp
```diff
@@ -5,73 +5,100 @@
...
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;;; command tables
-(define-command-table no-menu-test-table)
+(defmacro with-command-tables (tables &body body)
+  (if (null (cdr tables))
+      `(with-command-table ,(car tables)
+         ,@body)
+      `(with-command-table ,(car tables)
+         (with-command-tables ,(cdr tables)
+           ,@body))))
```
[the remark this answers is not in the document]
dkochmanski: I'm indifferent to that, changed.

Tests/commands.lisp
```diff
@@ -5,73 +5,100 @@
...
+    (let ((acc nil))
+      (do-command-table-inheritance (ct ct.2.*)
+        ;; This macro may traverse the same command table multiple times,
+        ;; because the ancestor may be linked in a few places in the inheritance
+        ;; tree with different :inherit-menu values.
+        (pushnew ct acc))
+      (is (every (lambda (ct) (member ct acc :test #'eq))
+                 (list ct ct.2 ct.2.1 ct.2.2 ct.2.*)))
+      (is (every (lambda (ct) (not (member ct acc :test #'eq)))
+                 (list ct.1 ct.x.x ct.x.y))))))
```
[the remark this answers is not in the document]
dkochmanski: sure, why not

Tests/commands.lisp
```diff
@@ -21,6 +21,14 @@
          (with-command-tables ,(cdr tables)
            ,@body))))
 
+(defmacro with-gestures ((&rest gestures) &body body)
+  `(progn
+     ,@(loop for gesture in gestures
```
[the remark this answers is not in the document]
dkochmanski: fixed as
```
(defmacro with-gestures ((&rest gestures) &body body)
  (loop for (name . spec) in gestures
        collect `(define-gesture-name ,name :keyboard ,spec) into defines
        collect `(delete-gesture-name ,name) into deletes
        finally (return `(progn
                           ,@defines
                           (unwind-protect (progn ,@body)
                             ,@deletes)))))
```

Tests/commands.lisp
```diff
@@ -140,6 +148,126 @@
...
+          (test-look :x sub1 '(com-x) root-menu)
+          (test-find :w sub1 nil nil)
+          (test-look :w sub1 '(com-w) sub1-menu)
+          ;; Entries are inherited
+          (test-find :y sub1 '(cmd-1) root)
+          ;; And may be shadowed
+          (test-find :z root '(cmd-2) root)
+          (test-find :z sub1 '(cmd-3) sub1))))))
+
+;;; This test is similar the previous one, however it examines :inherit-menu
```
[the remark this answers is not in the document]
dkochmanski: fixed

Documentation/Specification/commands.tex
```diff
@@ -183,8 +183,10 @@
 \cl{define-presentation-to-command-translator}.
 
 \item A table that associates keyboard gesture names to menu items (used to
-support keystroke accelerator interactions).  The keystroke accelerator table
-does not contain any items inherited from superior command tables.
+support keystroke accelerator interactions).  By default, the keystroke
+accelerator table does not contain any items inherited from superior command
+tables, although this can be overridden by the \cl{:inherit-menu} option to
+\cl{define-command-table}.
```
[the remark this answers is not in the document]
dkochmanski: fixed

Documentation/Specification/commands.tex
```diff
@@ -243,10 +245,11 @@
...
 
-If \arg{inherit-menu} is \term{true}, the new command table will inherit the
-menu items and keystroke accelerators from all of the inherited command tables.
-If it is \term{false} (the default), no menu items or keystroke accelerators
-will be inherited.
+If \arg{inherit-menu} is \cl{t}, the new command table will inherit the menu
+items and keystroke accelerators from all of the inherited command tables.  If
+it is \term{nil} (the default), no menu items or keystroke accelerators will
+be inherited. If it is \cl{:menu}, only menu items are inherited. If it is
+\cl{:keystrokes}, only keystroke accelerators are inherited.
```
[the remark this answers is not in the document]
dkochmanski: fixed

Documentation/Specification/commands.tex
dkochmanski: fixed

Documentation/Specification/commands.tex
dkochmanski: sure, adjusted

Core/clim-core/commands/tables.lisp
dkochmanski: I'm not sure why I've put string-equal here, I think that ordinary equal will suffice.

fixed accordingly, but not closing the issue in case if you think it is not a good idea.

Core/clim-core/commands/tables.lisp
dkochmanski: it is more readable to me that way -- a single task is encapsulated in a function, and then we call over inherited tables. Same applies to the second function -- do you want me to change that?

Core/clim-core/commands/tables.lisp
dkochmanski: ditto

Core/clim-core/commands/tables.lisp
dkochmanski: sure, done.

Core/clim-core/commands/tables.lisp
dkochmanski: I think that I've made a mistake by using string-equal, I'm changing that calls to simply equal.

Core/clim-core/commands/tables.lisp
dkochmanski: does it matter? it is not a keyword argument.

Core/clim-core/commands/tables.lisp
dkochmanski: fixed

Core/clim-core/commands/tables.lisp
dkochmanski: I don't have anything against find, however

```
(when-let ((item (find gesture (slot-value table 'menu)
                       :test test
                       :key #'command-menu-item-keystroke)))
  (return-from find-keystroke-item
    (values item table)))
```

does not look /much/ better. I'd probably write it with find now, but this code is removed one commit later and is not incorrect, so I'd rather leave it be.

Core/clim-core/commands/tables.lisp
dkochmanski: as noted below, this code is not incorrect and immedietely replaced later (as in - it is just a transient state for clarity)

Tests/commands.lisp
```diff
@@ -140,6 +148,126 @@
       (is (every (lambda (ct) (not (member ct acc :test #'eq)))
                  (list ct.1 ct.x.x ct.x.y))))))
 
+
+(defmacro with-keystroke-item-testers (nil &body body)
+  `(labels ((test-find (gesture table expected-item expected-table)
+              (multiple-value-bind (item found-table)
+                  (find-keystroke-item gesture table :test #'eql :errorp nil)
+                (is (equal (list expected-item expected-table)
+                           (list (and item (command-menu-item-value item)) found-table)))
```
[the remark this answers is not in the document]
dkochmanski: these things come in pairs (expected item should also match the table) -- I think that it is reasonably clear when you look into the test code.

Core/clim-core/commands/tables.lisp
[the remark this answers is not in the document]
scymtym: OK.

Core/clim-core/commands/tables.lisp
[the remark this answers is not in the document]
scymtym: I didn't get that from the commit message. Maybe

> According to the specification, most command table mapping operators accept a keyword INHERITED defaulting to T. Thanks to that it was possible to inhibit inherited objects mapping. 
The functions MAP-OVER-COMMAND-TABLE-KEYSTROKES and MAP-OVER-COMMAND-TABLE-MENU-ITEMS are exceptional in the specification because the keyword parameter is not specified for them.
As a McCLIM extension, MAP-OVER-COMMAND-TABLE-KEYSTROKES and MAP-OVER-COMMAND-TABLE-MENU-ITEMS now also accept the keyword argument.

And maybe add a comment marking the keyword parameter as an extension to the function definitions.

Tests/commands.lisp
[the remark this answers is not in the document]
scymtym: OK, I trust your judgement.

Core/clim-core/commands/tables.lisp
scymtym: The old code used `string-equal` but I'm not sure why.

scymtym: Ah, the spec entry for `add-menu-item-to-command-table` says
> Adds a command menu item to command-table's menu. string is the name of the command menu item; **its character case is ignored**.

Core/clim-core/commands/tables.lisp
scymtym: OK.

Core/clim-core/commands/tables.lisp
scymtym: I mean `map-table-entries` is only called once immediately after its definition. It does not take part in the recursive traversal as a functional argument or anything. In other words: why not replace the `flet` and the call with the body of `map-table-entries`?

Core/clim-core/commands/tables.lisp
scymtym: Right, it's mood since replaced in subsequent commit.

Core/clim-core/commands/tables.lisp
scymtym: But the `when-let` shouldn't be needed regardless since search for `nil` as name wouldn't have worked in the old version either.

But see above comment regarding case-insensitivity of menu item names.

Core/clim-core/commands/tables.lisp
scymtym: Not for execution semantics, but I think it makes sense to use the same names as the spec for users and maintainers.

Core/clim-core/commands/tables.lisp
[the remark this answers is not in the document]
dkochmanski: fixed the commit message according to the suggestion.

Core/clim-core/commands/tables.lisp
dkochmanski: closing (as discussed on irc)

Core/clim-core/commands/tables.lisp
dkochmanski: when-let is needed, because now we store both named and keystroke-only items in the menu. without when-let, call to (find-menu-item nil table) if there are keystroke-only entries would return such entry.

of course it is could be argued, that menu-name being nil is UB, or that we should that it is a string.

Core/clim-core/commands/tables.lisp
dkochmanski: replaced with equalp

Core/clim-core/commands/tables.lisp
dkochmanski: I have no issue with changing that -- my point is that I expect that the compiler will inline it anyway, but it is more readable (at least to me) when that part of the functionality is isolated as a (local) function.

scymtym: Please change it, if you don't mind. I'm not concerned with efficiency either, but abstracting this, in particular a single form, as function signals (to me) that it is going to be used more than once, so this is surprising. It is also longer and more complex than the direct version. A comment above the form could be good, though.

Core/clim-core/commands/tables.lisp
scymtym: Wouldn't validating `menu-name` be better, then?

dkochmanski: it already paid off, while testing the change I've found an issue somewhere else.

validating is fixed and the uncovered issue is fixed in a separate commit 

183aade7 remove-command-from-command-table: remove the matching menu entry

scymtym: Nice! Could you add a test which checks that the menu entry is removed? Should be easy with all the testing machinery you added.

dkochmanski: I've simply extended the test commands.add-and-remove - fixed.

scymtym (approved):

scymtym:
Looks good to me.

PR #1089, opened 2020-10-27
Fix repainting of GENERIC-LIST-PANE

### LINKED ISSUE #1083: List pane does not reset its region(?) when items change
When the items of a `list-pane` change (`(setf clime:list-pane-items)`), the pane doesn't seem to update (at least in the "shrinking" case) its region. This can be seen in the "Drawing Tests" demo: select a category with many tests such as "Ellipse", then select one with fewer tests such as "Rectangle". Observe how 1) the scroll bar of the Tests pane does not fit the new number of items 2) artifacts appear after the existing items when scrolling.

scymtym:
## Fixes and improvements for `generic-list-pane`

* Fill bottom part of the sheet region/viewport region that is not covered by rendered items with the background ink.

* Change space requirements and repaint after `visible-items` is changed.

* Reset `items-origin` after `list-pane-items` is set.

* Remove unnecessary `(setf list-pane-items) :after` method. This method did three things, change space requirements, clear pane background and initiate a repaint, that are already done by the other methods.

* Efficiency improvements: do fewer things for every item: such as calling various readers and setting the text style

## `allocate-space` improvement for `viewport-pane`

The previous `allocate-space` method called `move-and-resize-sheet` and `allocate-space` for the child pane before updating the scroll bars. That order resulted in `move-and-resize-sheet` using the old scroll bar values which, for example if the child shrunk, could move the child to a translation that would normally not be reachable.

This fixes #1083

### DIFF
--- a/Core/clim-core/gadgets/concrete.lisp
+++ b/Core/clim-core/gadgets/concrete.lisp
@@ -996,23 +996,24 @@ response to scroll wheel events."))
 (defgeneric generic-list-pane-item-height (generic-list-pane))
 
 (defmethod generic-list-pane-item-height ((pane generic-list-pane))
-  (+ (text-style-ascent  (pane-text-style pane) pane)
-     (text-style-descent (pane-text-style pane) pane)))
+  (text-style-height (pane-text-style pane) pane))
 
 (defmethod visible-items ((pane generic-list-pane))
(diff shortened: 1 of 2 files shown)


dkochmanski (approved):
Looks good to me. I have no critical remarks.

PR #1091, opened 2020-11-03
Implement modern menu navigation style

dkochmanski:
And fix some issues along the way.


dkochmanski:
code for testing:

```
(define-command cmd ((x string))
  (format t "~a~%" x)
  (finish-output t))

(define-gesture-name :C-x :keyboard (#\x :control))
(define-gesture-name :C-y :keyboard (#\y :control))
(define-gesture-name :C-z :keyboard (#\z :control))

(define-gesture-name :C-1 :keyboard (#\1 :control))
(define-gesture-name :C-2 :keyboard (#\2 :control))
(define-gesture-name :C-3 :keyboard (#\3 :control))

(define-command-table leaf-1 :inherit-from nil
  :menu (("cmd-1" :command (cmd "leaf-1 cmd 1") :keystroke :c-1)
         ("cmd-2" :command (cmd "leaf-1 cmd 2") :keystroke :c-2)
         ("cmd-3" :command (cmd "leaf-1 cmd 2") :keystroke :c-3)))

(define-command-table leaf-2 :inherit-from nil
  :menu (("cmd-1" :command (cmd "leaf-2 cmd 1"))
         ("cmd-2" :command (cmd "leaf-3 cmd 2"))
         ("cmd-3" :command (cmd "leaf-4 cmd 2"))))

(define-command-table root-1 :inherit-from nil
  :menu (("Leaf 1" :menu leaf-1 :keystroke :c-x)
         ("Leaf 2" :menu leaf-2)
         ("help" :divider :bonkers-value)
         ("Anon X" :menu (("cmd-1" :command (cmd "Anon cmd 1"))
                          ("XXXXX" :divider :oh :text-style (nil :bold nil))
                          ("cmd-2" :command (cmd "Anon cmd 2"))
                          (nil     :divider  nil)
                          ("kitten" :menu (("Hi-1" :command (cmd "hueh"))
                                           ("Hi-2" :command (cmd "bleh"))))
                          ("doggie" :menu (("Hi-1" :command (cmd "hueh"))
                                           ("Hi-2" :command (cmd "bleh"))
                                           ("di" :divider :oh)
                                           ("Hi-3" :command (cmd "hue*"))
                                           ("Hi-4" :command (cmd "ble*"))))))))

;;; When the command table has no ancestors there are no defined translators,
;;; even the identity translator. We need to add it manually.
(define-presentation-translator identity-translator
    (t nil root-1 :gesture :select
                  :tester ((object presentation context-type)
                           ;; see the comments around DEFUN PRESENTATION-SUBTYPEP
                           ;; for some of the logic behind this.  Only when
                           ;; PRESENTATION-SUBTYPEP is unsure do we test the object
                           ;; itself for PRESENTATION-TYPEP.
                           (declare (ignore object))
                           (multiple-value-bind (yp sp)
                               (presentation-subtypep (presentation-type presentation)
                                                      context-type)
                             (or yp (not sp))))
                  :tester-definitive nil)
    (object presentation)
  (values object (presentation-type presentation)))

(add-command-to-command-table 'cmd 'root-1 :errorp nil)

(define-application-frame test-menu ()
  ()
  (:pane :interactor :borders nil)
  (:pointer-documentation t)
  (:command-table (test-menu :inherit-from (root-1) :inherit-menu t
                             :menu (("HAH" :command (cmd "boo"))
                                    ("FOO" :command foo)
                                    (nil :divider :bonkers-value))))
  (:menu-bar test-menu))

(define-test-menu-command (foo :name t) ()
  (format t "HI~%"))

(eval-when (:execute :load-toplevel)
  (run-frame-top-level (make-application-frame 'test-menu)))
```

dkochmanski:
rebased against master

Core/clim-core/commands/tables.lisp
```diff
@@ -149,10 +149,8 @@
           (map-func command-table)))))
 
 (defun command-present-in-command-table-p (command-name command-table)
-  (let ((table (find-command-table command-table)))
-    (if (gethash command-name (slot-value table 'commands))
-        table
-        nil)))
+  (let ((ht (slot-value (find-command-table command-table) 'commands)))
+    (gethash command-name ht nil)))
```
scymtym, line 154: This should probably not return two values.

Core/clim-core/panes/composition.lisp
```diff
@@ -754,10 +754,15 @@
...
-      (mapc (curry #'sheet-adopt-child pane) children))))
+    (loop for box in contents
+          for client = (parse-box-content box)
+          for box-pane = (box-client-pane client)
+          collect client into clients
+          when box-pane
+            collect box-pane into children
+          finally
+             (setf (box-layout-mixin-clients pane) clients)
+             (mapc (curry #'sheet-adopt-child pane) children))))
```
scymtym, line 765: If `sheet-adopt-child` must only be called after setting `box-layout-mixin-clients`, please add a comment mentioning that face. Otherwise, immediately calling `sheet-adopt-child` without collecting would seem like a further simplification.

Core/clim-basic/extended-streams/stream-input.lisp
```diff
@@ -365,23 +365,25 @@ keys read."))
...
+                 (accelerator-gesture-p (gesture)
+                   (loop for gesture-name in *accelerator-gestures*
+                           thereis (event-matches-gesture-name-p gesture gesture-name))))
+            (cond
+              ((abort-gesture-p gesture)
+               (signal 'abort-gesture :event gesture))
+              ((accelerator-gesture-p gesture)
+               (signal 'accelerator-gesture :event gesture))
+              (t
+               (return-from stream-read-gesture gesture))))))))
```
scymtym, line 386: How do we know that we must loop instead of returning the gesture?

Core/clim-core/commands/commands.lisp
```diff
@@ -74,8 +74,10 @@
 (defun menu-items-from-list (menu)
   (mapcar
    #'(lambda (item)
-       (destructuring-bind (name type value &rest args) item
-         (apply #'make-menu-item name type value args)))
+       (if (atom item)
+           item
+           (destructuring-bind (name type value &rest args) item
+             (apply #'make-menu-item name type value args))))
```
scymtym, line 80: Please exercise this in one of the examples/tests.

Core/clim-core/commands/commands.lisp
```diff
@@ -101,7 +103,8 @@
      ;; sub-menus.
      (check-type value (or command-table symbol cons))
      (when (listp value)
-       (setf value (menu-items-from-list value))))
+       (let ((items (menu-items-from-list value)))
+         (setf value (make-command-table nil :menu items)))))
```
scymtym, line 107: Please exercise this in one of the examples/tests.

Core/clim-core/frames/frames.lisp
```diff
@@ -901,7 +901,9 @@ frames and will not have focus.
```
scymtym, line 903: Please split this into multiple commits. Maybe header change, visual change, tracking-pointer change and rest.

Core/clim-core/gadgets/menu.lisp
```diff
@@ -196,13 +204,10 @@
          (type (command-menu-item-type item))
          (value (command-menu-item-value item))
          (text-style (command-menu-item-text-style item))
-         (frame (pane-frame client))
-         (manager (frame-manager frame)))
+         (frame *application-frame*))
     (flet ((make-sub-pane (class &rest initargs &key &allow-other-keys)
-             (apply #'make-pane-1 manager frame class
+             (apply #'make-pane class
```
scymtym, line 299: Why this change? I have seen many problems caused by the combination of nested command loops and `*application-frame*`. Using the "hard-wired" frame instead of the special variable seems more robust.

scymtym: In fact, in relation to the other changes: the client of a gadget is usually the frame. Maybe that logic would make sense here as well.

Core/clim-core/gadgets/menu.lisp
```diff
@@ -249,8 +339,15 @@
...
                                          :disarmed-callback 'disarm-menu-button-callback))
                 (rack (make-menu-bar value sub-pane 'vmenu-pane))
-                (border (make-pane 'raised-pane :contents (list rack))))
+                (border (make-pane 'raised-pane :contents (list rack)))
+                (active (remove-if-not (lambda (elt)
+                                         (and (typep elt 'menu-button-pane)
+                                              (gadget-active-p elt)))
+                                       ;; Composite panes adopt children in
+                                       ;; order, so they are in reverse order.
+                                       (reverse (sheet-children rack)))))
```
scymtym, line 348: Is it necessary to cache this? Couldn't `menu-children` do this on demand?

scymtym: In particular since the method for frames is identical.

scymtym (changes requested):
Initial remarks in commit comments.

Core/clim-basic/extended-streams/stream-input.lisp
dkochmanski: This is not explicitly specified. the thing is that what we did before was wrong (because stream-read-gesture returned nil when the condition was not handled).

We must either loop or return the gesture. Looping has a wonderful benefit though -  you may change the command table (and the set of acceleration/abort gestures) dynamically, or simly duly note the value of the accelerator and continue processing where it was.

That will allow baking into our command processing gesture combos (without compromising backward compatibility nor requireing custom command processors like ESA).

Core/clim-core/commands/tables.lisp
dkochmanski: fixed (using nth-value 1 (gethash))

Core/clim-core/gadgets/menu.lisp
dkochmanski: it is not usually the frame. it is the frame only for top-level entries, for subtrees it is the rack pane. I've changed it to use typecase (with a fallback otherwise clause to *application-frame*).

Core/clim-core/gadgets/menu.lisp
dkochmanski: OK. I've changed menu-children to a function too and removed submenu-items function.

Core/clim-core/panes/composition.lisp
dkochmanski: I did that this way because I'm not certain whether the order of operations is important (so I didn't want to change semantics).

Core/clim-core/frames/frames.lisp
dkochmanski: I'm not certain whether it makes sense, the file was simply rewritten to make further changes easier to understand.

dkochmanski: I'd rather squash these three commits, however I've put the header and visual change in two separate commits.

Regarding changing of protocols and the tracking-pointer change -- one depends on another and "splitting" them would rather entail writing a completely new implementation that is based on old protocols and I don't see how that would be an improvement.

Core/clim-core/commands/commands.lisp
dkochmanski: fixed (extended menu-test)

Core/clim-core/commands/commands.lisp
dkochmanski: fixed (extended menu-test)

dkochmanski:
I've answered all remarks (some are waiting for answer).

PR #1092, opened 2020-11-03
Fix name collisions of ad-hoc classes generated by CLX backend for mirroring

scymtym:
* Cosmetic changes, untabifying and file headers

* Also share a single implementation of `maybe-mirroring` between CLX and CLX-fb

* Simplify mirroring decision logic

* Simplify CLX server path parsing


dkochmanski (approved):
I've looked into commits and I've nothing except for some bikeshedding (that I did not include), merging as is.

PR #1093, opened 2020-11-05
New mechanism for looking up the port

dkochmanski:
Introduced change is backward-compatible (fallback method regresses back to symbol plist).

### DIFF
--- a/Backends/CLX-fb/port.lisp
+++ b/Backends/CLX-fb/port.lisp
@@ -5,8 +5,8 @@
 		       clim-clx::clx-basic-port)
   ())
 
-(setf (get :clx-fb :port-type) 'clx-fb-port)
-(setf (get :clx-fb :server-path-parser) 'clim-clx::parse-clx-server-path)
+(defmethod find-port-type ((type (eql :clx-fb)))
+  (values 'clx-fb-port (nth-value 1 (find-port-type :clx))))
 
 (defmethod initialize-instance :after ((port clx-fb-port) &rest args)
   (declare (ignore args))
@@ -17,7 +17,6 @@
   (initialize-clx port)
   (initialize-clx-framebuffer port)
   (clim-extensions:port-all-font-families port))
-
 
 (defun initialize-clx-framebuffer (port)
   (clim-sys:make-process (lambda ()
--- a/Backends/PostScript/package.lisp
+++ b/Backends/PostScript/package.lisp
@@ -19,7 +19,7 @@
 ;;; Boston, MA  02111-1307  USA.
 
 (cl:defpackage #:clim-postscript
-  (:use #:clim #:clim-extensions #:clim-lisp #:clim-postscript-font)
+  (:use #:clim #:clime #:climb #:clim-lisp #:clim-postscript-font)
   (:import-from #:clim-internals
                 #:get-environment-variable
                 #:map-repeated-sequence
--- a/Core/clim-basic/windowing/ports.lisp
+++ b/Core/clim-basic/windowing/ports.lisp
@@ -43,34 +43,33 @@
             (t :clx-ttf))
     :null))
 
-(defun find-default-server-path ()
-  (loop for port in *server-path-search-order*
-        if (get port :port-type)
-           do (return-from find-default-server-path (list port))
-        finally (error "No CLIM backends have been loaded!")))
+(defgeneric find-port-type (port)
+  (:method ((port symbol))
+    (values (get port :port-type)
+            (get port :server-path-parser))))
 
 (defvar *all-ports* nil)
 
 (defun find-port (&key (server-path *default-server-path*))
-  (if (null server-path)
-      (setq server-path (find-default-server-path)))
-  (if (atom server-path)
-      (setq server-path (list server-path)))
-  (setq server-path
-        (funcall (get (first server-path) :server-path-parser) server-path))
-  (loop for port in *all-ports*
-        if (equal server-path (port-server-path port))
-        do (return port)
-        finally (let ((port-type (get (first server-path) :port-type))
-                      port)
-                  (if (null port-type)
-                      (error "Don't know how to make a port of type ~S"
-                             server-path))
-                  (setq port
-                        (funcall 'make-instance port-type
-                                 :server-path server-path))
-                  (push port *all-ports*)
-                  (return port))))
+  (setf server-path (alexandria:ensure-list server-path))
+  (let (port-class)
+    (flet ((try-port (type path)
+             (multiple-value-bind (class parser) (find-port-type type)
+               (when class
+                 (setf port-class class
+                       server-path (if (null parser)
+                                       path
+                                       (funcall parser path)))))))
+      (if (null server-path)
+          (loop for port-type in *server-path-search-order*
+                do (try-port port-type (list port-type))
+                while (null port-class))
+          (try-port (first server-path) server-path))
+      (when (null port-class)
+        (error "No CLIM backends have been loaded!"))
+      (or (find server-path *all-ports* :test #'equal :key #'port-server-path)
+          (first (push (make-instance port-class :server-path server-path)
+                       *all-ports*))))))
 
 (defmacro with-port ((port-var server &rest args &key &allow-other-keys)
                      &body body)
--- a/Core/clim-basic/windowing/ports.lisp
+++ b/Core/clim-basic/windowing/ports.lisp
@@ -49,6 +49,7 @@
             (get port :server-path-parser))))
 
 (defvar *all-ports* nil)
+(defvar *all-ports-lock* (make-lock "All ports lock."))
 
 (defun find-port (&key (server-path *default-server-path*))
   (setf server-path (alexandria:ensure-list server-path))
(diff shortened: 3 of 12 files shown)


Core/clim-basic/windowing/ports.lisp
```diff
@@ -43,34 +43,34 @@
...
-                  (return port))))
+  (unless (listp server-path)
+    (setf server-path (list server-path)))
+  (multiple-value-bind (port-class server-path)
+      (if (null server-path)
+          (loop for port in *server-path-search-order*
+                do (multiple-value-bind (class parser)
+                       (find-port-type port)
+                     (when class
+                       (return (values class (funcall parser (list class))))))
```
scymtym, line 62: I'm pretty sure the parser was called with the port type, not the port class.

Core/clim-basic/windowing/ports.lisp
```diff
@@ -43,34 +43,34 @@
...
-                  (if (null port-type)
-                      (error "Don't know how to make a port of type ~S"
-                             server-path))
-                  (setq port
-                        (funcall 'make-instance port-type
-                                 :server-path server-path))
-                  (push port *all-ports*)
-                  (return port))))
+  (unless (listp server-path)
+    (setf server-path (list server-path)))
```
scymtym, line 55: `ensure-list`?

Core/clim-basic/windowing/ports.lisp
```diff
@@ -43,34 +43,34 @@
...
+                   (error "No CLIM backends have been loaded!"))
+          (multiple-value-bind (class parser)
+              (find-port-type (first server-path))
+            (values class (funcall parser server-path))))
+    (loop for port in *all-ports*
+          if (equal server-path (port-server-path port))
+            do (return-from find-port port))
+    (let ((port (make-instance port-class :server-path server-path)))
+      (push port *all-ports*)
+      port)))
```
scymtym, line 73: maybe
 ```
(or (find port *all-ports* :test #'equal :key #'port-server-path)
     (let ((port …
```

File: Core/clim-basic/windowing/ports.lisp [hunk 4]
```diff
@@ -43,34 +43,34 @@
...
+      (if (null server-path)
+          (loop for port in *server-path-search-order*
+                do (multiple-value-bind (class parser)
+                       (find-port-type port)
+                     (when class
+                       (return (values class (funcall parser (list class))))))
+                finally
+                   (error "No CLIM backends have been loaded!"))
+          (multiple-value-bind (class parser)
+              (find-port-type (first server-path))
```
scymtym, line 66: I think the error reporting in case this port does not exist is now missing.

Backends/CLX-fb/port.lisp
```diff
@@ -5,8 +5,11 @@
 		       clim-clx::clx-basic-port)
   ())
 
-(setf (get :clx-fb :port-type) 'clx-fb-port)
-(setf (get :clx-fb :server-path-parser) 'clim-clx::parse-clx-server-path)
+(defmethod find-port-type ((type (eql :clx-fb)))
+  (multiple-value-bind (type parser)
+      (find-port-type :clx)
+    (declare (ignore type))
+    (values 'clx-fb-port parser)))
```
scymtym, line 12: `(values 'clx-fb-port (nth-value 1 (find-port-type :clx)))`

Backends/PostScript/package.lisp
```diff
@@ -19,7 +19,7 @@
 ;;; Boston, MA  02111-1307  USA.
 
 (cl:defpackage #:clim-postscript
-  (:use #:clim #:clim-extensions #:clim-lisp #:clim-postscript-font)
+  (:use #:clim #:clime #:climb #:clim-lisp #:clim-postscript-font)
```
scymtym, line 22: Can you write `climb:find-port-type` instead of importing this? In my experience, `use`ing to many packages in backends has already let to lots of confusion.

scymtym (changes requested):
Requested changes are in commit comments.

Backends/PostScript/package.lisp
dkochmanski: I can, however I've put it in use because clim-postscript is the only exception among backends that do not use climb, so I thought that consistency will be good here.

Core/clim-basic/windowing/ports.lisp
dkochmanski: done

Core/clim-basic/windowing/ports.lisp
dkochmanski: fixed

Core/clim-basic/windowing/ports.lisp
dkochmanski: fixed

Core/clim-basic/windowing/ports.lisp
dkochmanski: fixed

Backends/CLX-fb/port.lisp
dkochmanski: fixed

Core/clim-basic/windowing/ports.lisp
```diff
@@ -43,34 +43,34 @@
...
+      (if (null server-path)
+          (loop for port-type in *server-path-search-order*
+                do (try-port port-type (list port-type))
+                while (null port-class))
+          (try-port (first server-path) server-path))
+      (when (null port-class)
+        (error "No CLIM backends have been loaded!"))
+      (or (find server-path *all-ports* :test #'equal :key #'port-server-path)
+          (first (push (make-instance port-class :server-path server-path)
+                       *all-ports*))))))
```
dkochmanski, line 73: I have reworked this function to put common logic in flet.

dkochmanski (commented):
change.

Core/clim-basic/windowing/ports.lisp
```diff
@@ -43,34 +43,34 @@
...
+  (setf server-path (alexandria:ensure-list server-path))
+  (let (port-class)
+    (flet ((try-port (type path)
+             (multiple-value-bind (class parser) (find-port-type type)
+               (when class
+                 (if (null parser)
+                     (setf port-class class
+                           server-path path)
+                     (setf port-class class
+                           server-path (funcall parser path)))))))
```
scymtym, line 63: Since the behavior for `class` does not depend on `parser`, maybe like this?
```
(when class
  (setf port-class class
        server-path (if (null parser) 
                        path
                        (funcall parser path))))

File: Backends/PostScript/package.lisp (hunk 6 above)
Comment (scymtym) [reply Backends/PostScript/package.lisp:22]: I would like to move into the opposite direction such that eventually no backend `use`s `climb`, but you decide.

**Review by scymtym — CHANGES_REQUESTED** [review]
A few more comments.

File: Backends/PostScript/package.lisp (hunk 6 above)
Comment (dkochmanski) [reply Backends/PostScript/package.lisp:22]: in that case I'm closing this

File: Core/clim-basic/windowing/ports.lisp (hunk 8 above)
Comment (dkochmanski) [reply Core/clim-basic/windowing/ports.lisp]: done.

**Comment by dkochmanski** [comment]
all remarks addressed.

File: Core/clim-basic/windowing/ports.lisp [hunk 9]
```diff
@@ -43,34 +43,33 @@
...
+      (if (null server-path)
+          (loop for port-type in *server-path-search-order*
+                do (try-port port-type (list port-type))
+                while (null port-class))
+          (try-port (first server-path) server-path))
+      (when (null port-class)
+        (error "No CLIM backends have been loaded!"))
+      (or (find server-path *all-ports* :test #'equal :key #'port-server-path)
+          (first (push (make-instance port-class :server-path server-path)
+                       *all-ports*))))))
```
scymtym, line 72: Can you add locking to this in a separate commit?

scymtym (changes requested):
One more suggestion.

scymtym:
Grepping for `:server-path-parser` finds these:
```
mcclim/Extensions/fonts/freetype.lisp
12: (setf (get :clx-ff :server-path-parser) 'clim-clx::parse-clx-server-path)
mcclim/Extensions/fonts/xrender-fonts.lisp
61: (setf (get :clx-ttf :server-path-parser) 'clim-clx::parse-clx-server-path)
```
Should they be changed as well?

Core/clim-basic/windowing/ports.lisp
dkochmanski: done

dkochmanski:
I've added the method for clx-ff too.

Core/clim-basic/windowing/ports.lisp
```diff
@@ -68,8 +69,9 @@
       (when (null port-class)
         (error "No CLIM backends have been loaded!"))
       (or (find server-path *all-ports* :test #'equal :key #'port-server-path)
-          (first (push (make-instance port-class :server-path server-path)
-                       *all-ports*))))))
+          (with-lock-held (*all-ports-lock*)
+            (first (push (make-instance port-class :server-path server-path)
+                         *all-ports*)))))))
```
scymtym: The critical section starts on the previous line: two threads could perform the `find`, both find nothing, then both (safely but incorrectly) add a new port.

scymtym (changes requested):
Lock scope issue.

Core/clim-basic/windowing/ports.lisp
dkochmanski: right, fixed

scymtym (approved):

scymtym:
Looks good to me.

PR #1094, opened 2020-11-06
DEFINE-APPLICATION-FRAME validates :GEOMETRY option

scymtym:


Core/clim-core/frames/define-application-frame.lisp
```diff
@@ -123,7 +129,8 @@
                               ~R).~@:>"
                              key values (length values)))
                     (check-one-value (first values)))
-                   (* (mapcar #'check-one-value values)))))
+                   (*
+                    (check-one-value values)))))
```
dkochmanski, line 133: Is this change correct? mapcar looks sensible here.

dkochmanski (changes requested):

Core/clim-core/frames/define-application-frame.lisp
scymtym: Both are correct. The `mapcar` version was intended to validate multiple values individually. The new version validates all of `values` at once, which is more powerful and even necessary for example for plist-shaped values. This change does not introduce problems with other options because no option which accepts multiple values had a type specification before this change.

dkochmanski: The name of the function is inaccurate then.  Either change the name of the function, or introduce a second function:
```
(flet ((check-1-values (values)
                        (destructuring-bind (value &rest rest) values
                          (unless (null rest)
                            (error "~@<The option ~S takes a single ~
                              argument (not ~{~S~^ ~} which are ~
                              ~R).~@:>"
                                   key values (length values)))
                          (when (and type (not (typep value type)))
                            (error "~@<The value ~S for option ~S is ~
                                  not of type ~A.~@:>"
                                   value key type)))
                        value)
                      (check-*-values (values)
                        (when (and type (not (typep value type)))
                          (error "~@<The value ~S for option ~S is ~
                                  not of type ~A.~@:>"
                                 value key type))
                        value))
                 (ecase value-count
                   (1 (check-1-values values))
                   (* (check-*-values values))))
```

I think that the latter solution, while more verbose, is more explicit with putting ephasis, that different things are checked depending on the value-count qualifier.

dkochmanski: n.b the fact that :type is used differently for different value-count specifiers may warrant a comment too.

scymtym: I commented and refactored the function to make the argument count distinction clearer.

Core/clim-core/frames/define-application-frame.lisp
```diff
@@ -86,7 +86,21 @@
...
 
+(defun geometry-specification-p (thing)
+  (and (alexandria:proper-list-p thing)
+       (evenp (length thing))
+       (loop for (key value) on thing by #'cddr
+             always (member key '(:left :top :right :bottom :width :height)))))
+
 (defun parse-define-application-frame-options (options)
+  ;; Note that options are always of the form (KEY
+  ;; . ARGUMENTS). Depending on KEY, ARGUMENTS is expected to have one
```
dkochmanski, line 97: this line wrap seems bogus

Core/clim-core/frames/define-application-frame.lisp
```diff
@@ -86,7 +86,21 @@
...
+  ;; of two shapes 1) an arbitrary list, denoted by * in the INFOS
```
dkochmanski, line 98: start each point from a new line

dkochmanski (changes requested):
Only two cosmetic changes in the comment formatting. Otherwise looks OK to me.

Core/clim-core/frames/define-application-frame.lisp
scymtym: Must be Emacs getting confused by the the `.`. Fixed.

Core/clim-core/frames/define-application-frame.lisp
scymtym: Done.

dkochmanski:
lgtm

PR #1096, opened 2020-11-07
invoke-with-output-to-raster-image: don't destroy the port on exit

### LINKED ISSUE #1095: raster backend: can't nest with-output-to-* operators
Steps to reproduce: go to drawing tests, and try to export the test rgb design to a png file (print all tests to png also triggers this error).

Narrowed down test case:

```
(defun drawing-test-raster-image* (filename)
  (mcclim-raster-image:with-output-to-raster-image-file
      (stream filename :width 200 :height 200)
    (draw-rectangle* stream 0 0 100 100)
    (mcclim-raster-image::with-output-to-rgba-pattern
        (stream :width 200 :height 200)
      (draw-rectangle* stream 0 0 200 200))))
```

Drawing operations are not relevant, the inner form errors with 

```
  #<STANDARD-GENERIC-FUNCTION MCCLIM-RENDER-INTERNALS::IMAGE-MIRROR-IMAGE (1)>
when called with arguments
  (NIL).
```

dkochmanski:
Previously the body of the function was wrapped in the macro with-port - that
caused destroyal of all mirrors upon exit. When the function invocation was
nested (i.e to compute some sub-image), the inner function destroyed also
parent's mirror and that lead to error upon exit.

The function had another problem rgb-port did not adopt the top-level sheet by
the graft but rather manually hard-wire it in the method realize-mirror. That
is fixed, so now when the sheet is degrafted, destroy-mirror is called (and
the function port-unregister-mirror is called). Thanks to that we may simply
call find-port to bind the port and before exitting disown the top-level-sheet
- port is not destroyed and there is no resource leakage. Fixes #1095.

### DIFF
--- a/Backends/RasterImage/output-to-image.lisp
+++ b/Backends/RasterImage/output-to-image.lisp
@@ -71,54 +71,56 @@
 
 (defun invoke-with-output-to-raster-image (continuation enter-fn exit-fn server format
                                            &key (width :compute) (height :compute)
-                                                (border-width 0) (recording-p t))
-  (with-port (port server :width 1 :height 1)
-    (let ((top-level-sheet (make-raster-top-level-sheet port format))
-          (vbox (make-instance 'vbox-pane :port port))
-          (border-pane (make-instance 'climi::border-pane
-                                      :port port
-                                      :border-width border-width))
-          (stream (make-raster-image-stream port)))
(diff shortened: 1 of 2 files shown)


dkochmanski:
erm, pushed wrong version. repushed.

dkochmanski:
closing as the new commit has this change also.

PR #1098, opened 2020-11-09
Icons for top-level sheets and frames

scymtym:
* Add a new protocol with documentation that consists of generic functions the names of which are exported from the `clim-extensions` package.

* Add an implementation in the CLX backend based on the `_NET_WM_ICON` property which is described in the "Application Window Properties" section of the "Extended Window Manager Hints" specification [1].

* Add default icon image files based on Pamela Mosiejczuk's (https://gumroad.com/flan_foundry) work.

* Extend the sheet and frame name demo such that the icon for the frame can be changed.

[1] https://specifications.freedesktop.org/wm-spec/1.5/ar01s05.html#idm45766085139216


scymtym:
@lokedhs Could you check with your friend whether she is fine with including the logo file she made in this way?

Core/clim-basic/decls.lisp
```diff
@@ -231,6 +231,20 @@ NEW-NAME must be a string.
 Changing the pretty name of SHEET usually changes the title of the
 window associated with it."))
 
+(defgeneric clime:sheet-icon (sheet)
```
dkochmanski, line 234: please remove package accessors - which symbol belongs to which package is decided in the file packages.lisp (and climi imports exported packages).

Core/clim-core/frames/frames.lisp
```diff
@@ -252,6 +260,15 @@ documentation produced by presentations.")
...
   (clime:note-frame-pretty-name-changed (frame-manager frame) frame new-value))
 
+(defmethod (setf clime:frame-icon) :after (new-value frame)
+  ;; If there is a top-level sheet, set its icon. The port can reflect
+  ;; this change by telling the window manager which might display the
+  ;; new icon somewhere.
+  (when-let ((top-level-sheet (frame-top-level-sheet frame)))
+    (setf (sheet-icon top-level-sheet) new-value))
+  ;; Let client code know.
+  (clime:note-frame-icon-changed (frame-manager frame) frame new-value))
```
dkochmanski, line 283: ditto, no package accessors in climi.

Documentation/Manual/Texinfo/extensions.texi
```diff
@@ -10,11 +11,22 @@
...
+@node Frame and sheet icons
+@section Frame and sheet icons
+
+According to the @clim{} specification, an icon can be defined for an
+application frame class using the @initarg{:icon} initarg of the
+@fmacro{define-application-frame} macro.  @mcclim{} extends this design to
+top-level sheets and adds
+
+@include docstrings/fun-clim-extensions-frame-icon.texi
+@include docstrings/fun-clim-extensions-setf-frame-icon.texi
```
dkochmanski, line 23: also sheet-icon and setf-sheet-icon.texi?

Core/clim-core/frames/frames.lisp
```diff
@@ -75,6 +75,19 @@
...
+                                     *load-pathname*))
+                (icon-pathname (merge-pathnames
+                                (make-pathname :directory '(:relative :up :up :up "data")
+                                               :name name
+                                               :type "png")
+                                file-pathname)))
+           (make-pattern-from-bitmap-file icon-pathname :format :png))))
+  (defvar *default-icon-large* (load-icon "mcclim-logo-round"))
+  (defvar *default-icon-small* (load-icon "mcclim-logo-round-small")))
+
```
dkochmanski, line 90: maybe macrolet, so they are top-level?

dkochmanski (changes requested):
the pr looks clean - that said a few remarks are included.

Core/clim-core/frames/frames.lisp
```diff
@@ -86,6 +100,13 @@
          :reader frame-name)
    (pretty-name :initarg :pretty-name
                 :accessor frame-pretty-name)
+   (%icon :initarg :icon
```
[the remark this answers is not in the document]
scymtym: I use that convention by default, so probably did it here out of habit. That said, I think the convention makes sense and we should switch to using it at some point.

Documentation/Specification/frames.tex
```diff
@@ -200,8 +200,9 @@
 consists of additional arguments to be passed to the function.  The default
 function is \cl{default-frame-top-level}.
 
-\item \cl{:icon} \arg{pixmap} specifies a pixmap to be displayed by the host's
-window manager when the frame is iconified.
+\item \cl{:icon} \arg{pattern} specifies an array pattern to be
+displayed by the host's window manager to visually enhance or
+represent the frame, for example when it is iconified.
```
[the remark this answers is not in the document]
scymtym: > do we want to be able to specify also icon-name? _net_wm_icon_name in x11 terms. It is the string associated with the window when it is iconified (i.e shorter than a name - "firefox" vs "icons for top-level sheets and frames by scymtym * pull request * Mozilla Firefox") - i.e in alt+tab switching of windows.

I'm using Gnome Shell and it doesn't seem to use `_NET_WM_ICON_NAME` but instead something else for the short name (maybe something from the `.desktop` file? who knows).
gnome-shell
`xprop` for the Firefox window says:

```
WM_CLASS(STRING) = "Navigator", "Firefox-esr"
WM_ICON_NAME(STRING) = "Icons for top-level sheets and frames by scymtym · Pull Request #1098 · McCLIM/McCLIM - Mozilla Firefox"
_NET_WM_ICON_NAME(UTF8_STRING) = "Icons for top-level sheets and frames by scymtym · Pull Request #1098 · McCLIM/McCLIM - Mozilla Firefox"
WM_NAME(STRING) = "Icons for top-level sheets and frames by scymtym · Pull Request #1098 · McCLIM/McCLIM - Mozilla Firefox"
_NET_WM_NAME(UTF8_STRING) = "Icons for top-level sheets and frames by scymtym · Pull Request #1098 · McCLIM/McCLIM - Mozilla Firefox"
```
This is basically what we already do: we set all four properties to the "pretty name" of the top-level sheet.

There must, however, be an additional way to compute the name: the "frame and sheet name" demo gets the name "Frame-Sheet-Name-Test" in the Gnome Shell top bar thing. This is not affected by any of the properties mentioned above:

frame-and-sheet-name

Maybe the "non-pretty name" is used here? Do we map that to some X concept? I can't see anything with the X introspection tools. Our code contains `(xlib:set-wm-class window instance-name class-name)` so maybe that's the source.

> do we want the icon-mask? it may be redundant with the transparency, so I think not.

Yes, the specification explicitly says ARGB, so it is not a problem for X. Should a backend need a separate mask, it could still extract the alpha channel of the ARGB image pattern.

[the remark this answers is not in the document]
scymtym: > after a cursory glance at clim tos, they seem to represent the icon as a lambda list: `(&key (icon-name (frame-pretty-name frame)) pixmap clipping-mask)` and destructure it in the backend.

I think it would be better to extend `:pretty-name` instead since the icon is not tied to the frame being iconified nowadays and similarly, a long name and a short name may be useful outside of iconified frames. The syntax could be like `(:pretty-name PRETTY-NAME :icon-name ICON-NAME)` or `(:pretty-name (:default PRETTY-NAME :short SHORT-NAME))` or similar.

Backends/CLX/mirror.lisp
```diff
@@ -25,6 +25,48 @@
...
+           (let ((bytes (icon-property-bytes icons)))
+             (xlib:change-property window :_NET_WM_ICON bytes :cardinal 32))))
+
+  (defmethod port-set-mirror-icon ((port clx-basic-port) mirror icon)
+    (install-icons mirror (list icon)))
+
+  (defmethod port-set-mirror-icon ((port clx-basic-port) mirror (icon sequence))
+    (if (alexandria:emptyp icon)
+        (xlib:delete-property mirror :_NET_WM_ICON)
+        (install-icons mirror icon))))
```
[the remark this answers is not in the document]
scymtym: Changed.

Core/clim-core/frames/frames.lisp
scymtym: Changed.

Documentation/Manual/Texinfo/extensions.texi
scymtym: Fixed.

Core/clim-core/frames/frames.lisp
scymtym: Changed.

Core/clim-basic/decls.lisp
scymtym: Changed.

scymtym:
I think I addressed all remarks.

Documentation/Specification/frames.tex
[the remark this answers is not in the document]
dkochmanski: closing (after discussion on irc).

Core/clim-core/frames/frames.lisp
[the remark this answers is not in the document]
dkochmanski: for now please uphold the existing convention.

Core/clim-core/clim-core.asd
```diff
@@ -1,7 +1,9 @@
 (in-package #:asdf-user)
 
 (defsystem "clim-core"
-  :depends-on ("clim-basic" #+sbcl (:require "sb-introspect"))
+  :depends-on (#+sbcl (:require "sb-introspect")
+               "clim-basic"
+               "mcclim-bitmaps") ; for loading default icons
```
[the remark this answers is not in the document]
dkochmanski: this is still not addressed.

dkochmanski (changes requested):

scymtym:
Sorry, missed three remarks.

> I think that mcclim-bitmaps should be pulled into core instead. while it was optional it could be kept as an "extension", but there is no need to keep these definitions outside of `Core/clim-basic/drawing/image-formats.lisp`

Are you proposing to create the above file and add it to the `clim-core` system? Or `clim-basic` since the `drawing` is there?

dkochmanski:
By "pulling into the core" I mean the `/Core/` directory - yes, the clim-basic system Core/clim-basic/drawing/. It may make sense to create a separate file for xpm image-formats-xpm.lisp (I'm not sure about that).

Core/clim-core/frames/frames.lisp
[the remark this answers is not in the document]
scymtym: Changed.

scymtym:
I would rather not add the image file format support to `clim-basic` in this pull request. I can just remove the default icons instead.

dkochmanski:
that's fine too.

scymtym:
The most recent commit has an alternative solution to the problem. If that solution is acceptable, I would squash the commits in question.

default-icons.lisp
```diff
@@ -0,0 +1,15 @@
...
+                (icon-pathname (merge-pathnames
+                                (make-pathname :directory '(:relative "data")
+                                               :name name
+                                               :type "png")
+                                file-pathname)))
+           (make-pattern-from-bitmap-file icon-pathname :format :png))))
+  (unless climi::*default-icon-large*
+    (setf climi::*default-icon-large* (load-icon "mcclim-logo-round")))
+  (unless climi::*default-icon-small*
+    (setf climi::*default-icon-small* (load-icon "mcclim-logo-round-small"))))
```
dkochmanski, line 15: package accessors are not necessary

Core/clim-core/frames/frames.lisp
```diff
@@ -101,7 +105,10 @@
                 :accessor frame-pretty-name)
    (icon :initarg :icon
          :accessor frame-icon
-         :initform (list *default-icon-large* *default-icon-small*)
+         :initform (append (when-let ((icon *default-icon-large*))
+                             (list icon))
+                           (when-let ((icon *default-icon-small*))
+                             (list icon)))
```
dkochmanski, line 111: (remove nil (list *default-icon-large* *default-icon-small*)) ;?

Core/clim-core/frames/define-application-frame.lisp
```diff
@@ -213,6 +215,10 @@
          (:default-initargs
           :name              ',name
           :pretty-name       ,pretty-name
+          ,@(when icon
+              `(:icon (typecase ,icon
+                        ((or string pathname) (clim:make-pattern-from-bitmap-file ,icon))
+                        (t ,icon))))
```
dkochmanski, line 221: this should still allow specifying :icon nil (if someone wants to /not/ include default icons).

dkochmanski (changes requested):
I'm OK with this alternate approach, I've left some remarks.

Core/clim-core/frames/define-application-frame.lisp
dkochmanski: ah, it does allow it already, nvm.

Core/clim-core/clim-core.asd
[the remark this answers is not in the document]
scymtym: Resolved by loading the default icons in the `mcclim` system.

default-icons.lisp
scymtym: Fixed.

Core/clim-core/frames/frames.lisp
scymtym: Fixed.

Core/clim-core/frames/define-application-frame.lisp
scymtym: Right. Also, I now think that coercing strings and pathnames should maybe be done in an `initialize-instance` method?

dkochmanski: makes sense to me.

scymtym: The macro now uses `icon-supplied-p`, all coercing is now done in the `initialize-instance` method.

Core/clim-core/frames/frames.lisp
```diff
@@ -241,9 +232,20 @@ documentation produced by presentations.")
...
+               (sequence
+                (map 'list #'coerce-to-icon thing))
+               (t
+                thing))))
+    (setf (slot-value obj 'icon) (cond ((not icon-supplied-p)
+                                        nil)
+                                       ((null icon)
+                                        nil)
+                                       (t
+                                        (coerce-to-icon icon)))))
```
dkochmanski: this seems bogus:

(cond ((not supplied-p) nil)
          ((null icon) nil)
          (t …))

is equivalent to

(and icon (coerce-to-icon icon))

or `(if (null icon) nil (coerce-to-icon icon))`

Also I'd rather have this change to initialize-instance squashed to the previous commit where coerce-to-icon was introduced (i.e I don't see a purpose in having that change split in two commits).

dkochmanski (changes requested):

Core/clim-core/frames/frames.lisp
dkochmanski: I see that it is fixed in later commit, so I'm merging this PR.

PR #1099, opened 2020-11-11
Fix FRAME-EXIT for three or more nested command loops

scymtym:
This fixes a regression introduced in 96c3158f. Between 96c3158f and this change, when three or more frames ran in nested command loops and the outermost frame exited, both inner command loops handled the
`frame-exit` condition in the `handler-bind` clause. This caused multiple calls to `disown-frame` of which only the first succeeded.

To reproduce the problem, start demodemo, start three demos, then close the frame of the demo which has been started first.


dkochmanski:
lgtm

PR #1100, opened 2020-11-11
Small fixes for the render extension

scymtym:
* Fix bit twiddling in flipping ink implementation

* Use correct package for `medium-draw-circle*`


dkochmanski:
lgtm

PR #1101, opened 2020-11-11
Fix bordered output styles :UNDERLINE and :CROSSOUT

scymtym:
Before this change, the styles `:underline` and `:crossout` searched the output record to which the style is applied for child output records of type `text-displayed-output-record`.
However, in some cases relevant children are of type `draw-text-output-record`.

To reproduce the problem:
1. Start demodemo
2. Start Drawing Tests
3. Enable the "side-by-side view with the renderer output" option
4. Select the Text » Underlining or Text » Crossing out test
5. Note that the text decorations do not appear in the output of the render backend

### DIFF
--- a/Core/clim-core/bordered-output.lisp
+++ b/Core/clim-core/bordered-output.lisp
@@ -448,37 +448,37 @@
                                        line-dashes)
   (let ((line-style (%%line-style-for-method)))
     (labels ((fn (record)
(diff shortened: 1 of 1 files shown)


dkochmanski:
this change is fine, but I wonder why the renderer output history uses a different class. maybe it would be worth fixing in this PR given that this (potential) issue won't be noticeable after merging.

PR #1102, opened 2020-11-17
graft: extend the core interface

dkochmanski:
- introduce a function graft-pixel-aspect-ratio

- extend functions graft-pixels-per-millimeter and graft-pixels-per-millimeter
  by allowing a key argument orientation

This change is backward compatible opens a path for working with backends
which have non-rectangular pixels.


dkochmanski:
this will be part of the third batch of backend-manual pull requests (the first one is still reviewed)

PR #1104, opened 2020-11-23
Make Shift-Space insert a space in Drei buffers

lokedhs:
After a change which caused the shift key to be treated as a regular modifier key, the self-insertion command will not trigger for shifted space anymore. This fix adds a new command that simply inserts a space and is bound to Shift-Space.


dkochmanski (approved):
lgtm

PR #1106, opened 2020-11-26
blank-area: extend the presentation type

dkochmanski:
Now the presentation type is backed by a class blank-area. Instances have
slots that allow determining where exactly the presentation was accepted.

Moreover the presentation type has a parameter sheet that allows specifying,
that a blank area should be found in a specified.


dkochmanski:
This change implies small non-conformance: the specification of the presentation type suggests, that there is a single "null presentation object", presumably the object stored in the `*null-presentation*` specified below. 

http://bauhh.dyndns.org:8000/clim-spec/23-8.html#_1260

Another drawback of consing a new blank-area instance is that when we have the blank-area context, a new object is consed after each motion event, which seems to be ugly (and that would explain why it was specified to work wit ha single object).

That said, extending the presentation type blank-area with a parameter sheet does is a fairly independent change which a) doesn't cause spec violation, b) doesn't cause excessive consing.

This feature is quite useful when we want to determine the place when a blank area was clicked - that would allow us to define a drag-and-drop translator and "know" both the starting and the finish position. I.e

```
(define-drag-and-drop-translator tr-select-region
    (blank-area command blank-area foo :feedback region-feedback
                                       :tester ((object window)
                                                (and (panep window)
                                                     (eq (pane-name window) 'app1))))
    (object destination-object)
  (let ((x1 (pointer-event-x object))
        (y1 (pointer-event-y object))
        (x2 (pointer-event-x destination-object))
        (y2 (pointer-event-y destination-object)))
    (when (> x1 x2) (rotatef x1 x2))
    (when (> y1 y2) (rotatef y1 y2))
    (let* ((frame *application-frame*)
           (app1 (find-pane-named frame 'app1))
           (pixm (find-pane-named frame 'pixm))
           (dest (with-input-context (nil :override t) ()
                     (accept `(or (blank-area ,app1)
                                  (blank-area ,pixm))))))
      `(com-copy app1
                 ,(pane-name (event-sheet dest))
                 ,x1 ,y1 ,(- x2 x1) ,(- y2 y1)
                 ,(pointer-event-x dest)
                 ,(pointer-event-y dest)))))
```

Note, that readers are purposefully specializing event generic functions to avoid the need to cons a new interface.

dkochmanski:
otoh, consing is not /that/ excessive given we cons each motion event here. that said, we could not cons a new event for sensitivity test but only after the element is "pressed", because we are not highlighting it either way. I'll think about adding that.

Examples/pixmaps.lisp
```diff
@@ -0,0 +1,172 @@
+;;;;  Copyright (c) 2020 Daniel Kochmański
+;;;;
+;;;;    License: BSD-2-Clause.
```
scymtym, line 3: Can this use the usual file header format? That format should work for any license.

Examples/pixmaps.lisp
```diff
@@ -0,0 +1,172 @@
...
+;;; The "Source" pane has displayed example data, the "Target" pane is another
+;;; ordinary pane while the "Pixmap" pane represents the allocated pixmap -
+;;; the pixmap itself is not visible because it is an off-screen window.
+;;; Copying and pasting on the "Pixmap" pane in fact copies from a pixmap and
+;;; pastes to a pixmap, and that pixmap data is copied to the pane with
+;;; MEDIUM-COPY-AREA.
+;;;
+;;; As a bonus, this demo illustrates how to use blank-area for drag-and-drop
+;;; transformations with a McCLIM-specific extension allowing to specify the
+;;; sheet and to obtain both sheet and coordinates of the blank-area object.
```
scymtym, line 27: Can you put some of this into a label pane in the demo frame, like the "file manager" demo does?

Examples/pixmaps.lisp
```diff
@@ -0,0 +1,172 @@
...
+  (with-identity-transformation (stream)
+    (ecase state
+      (:highlight
+       (with-output-recording-options (stream :record nil)
+         (draw-rectangle* stream x0 y0 x y :filled nil
+                                           :line-thickness 3
+                                           :line-dashes t)))
+      (:unhighlight
+       (let ((region (make-rectangle* (- (min x0 x) 2) (- (min y0 y) 2)
+                                      (+ (max x0 x) 2) (+ (max y0 y) 2))))
```
scymtym, line 93: Could these be passed to `draw-rectangle*` directly?

Examples/pixmaps.lisp
```diff
@@ -0,0 +1,172 @@
...
+                      (sheet-medium pane1)))
+         (medium2 (if (typep pane2 'pixmap-sheet)
+                      (sheet-pixmap pane2)
+                      (sheet-medium pane2))))
+    (medium-copy-area medium1 from-x from-y width height medium2 to-x to-y)
+    (map () (lambda (p)
+              (repaint-sheet p +everywhere+))
+         (remove-duplicates (list pane1 pane2)))))
+
+(defun pixmaps (&optional (new-process nil))
```
scymtym, line 166: `new-process` is usually a keyword parameter. Maybe do that here as well for consistency?

scymtym (changes requested):
Some procrastination-level remarks in commit comments.

Examples/pixmaps.lisp
dkochmanski: fixed

Examples/pixmaps.lisp
dkochmanski: they could, but we call repaint-sheet with the region argument anyway, so I think that this is cleaner to cons the region once and use the result.

Examples/pixmaps.lisp
dkochmanski: fixed

Examples/pixmaps.lisp
dkochmanski: fixed

dkochmanski:
rebased and all remarks addressed.

Core/clim-core/presentations/translators.lisp
```diff
@@ -626,7 +623,7 @@ and used to ensure that presentation-translators-caches are up to date.")
      #'(lambda (translator presentation context)
          (return-from find-innermost-presentation-match
            (values presentation translator context)))
-     *null-presentation*
+     (make-blank-area-presentation window x y)
```
dkochmanski, line 626: the event is available in this context, so the blank-area should be constructed with the presentation-object assigned to that event

Core/clim-core/standard-presentations.lisp
```diff
@@ -118,28 +118,44 @@
...
+(define-presentation-type blank-area (&key sheet region))
 
 (define-presentation-method presentation-typep (object (type blank-area))
-  (eq object (presentation-object *null-presentation*)))
+  (and (typep object '(or null pointer-event))
+       (or (eq sheet '*) (eq (event-sheet object) sheet))
+       (or (eq region '*) (region-contains-position-p
+                           region
+                           (pointer-event-x object)
+                           (pointer-event-y object)))))
```
dkochmanski, line 145:
```
(or (null pointer-event)
      (and (typep object pointer-event …) …))
```

otherwise it will error when we call (event-sheet nil)

Core/clim-core/standard-presentations.lisp
```diff
@@ -149,6 +165,33 @@
...
+      *null-presentation*
+      (let ((event (or event
+                       (make-instance 'pointer-event :sheet sheet :x x :y y))))
+        (multiple-value-bind (x1 y1 x2 y2) (bounding-rectangle* sheet)
+          (make-instance 'standard-presentation
+                         :object event
+                         :type `((blank-area :sheet ,sheet :region ,(make-point x y))
+                                 :description "Blank Area")
+                         :view +textual-view+
+                         :x1 x1 :y1 y1 :x2 x2 :y2 y2)))))
```
dkochmanski, line 194: this function could accept only the event (and the caller would be responsible for synthesizing it)

dkochmanski: also this would be easier to read as:

```
(when (null event)
  (return-from … *null-presentation*))
(let ((sheet (event-sheet …) …)
  (multiple-value-bind …))
```

dkochmanski:
I've slightly changed the interpretation of arguments in test-presentation-translator and tweaked presentation-typep to account for *null-presentation* better. I'll test the branch a bit more after the weekend and if nothing comes up - merge it.

PR #1107, opened 2020-11-30
Fix execute-frame-command

dkochmanski:
Also fix with-scaling.


dkochmanski:
I wonder what should happen if the command arrives during some operation spanning multiple events (i.e drag and drop) - should the composition be aborted, or maybe we should (by some means) finish the operation, return that operation's command (if it translates to a command), and during the next iteration of the top level loop instead of calling read-command use that executed command? Sounds non-trivial.

PR #1114, opened 2020-12-17
franz: fix DRAGGING-DRAWING

admich:
Commit 5bc9b44b simplified DRAGGING-DRAWING but introduced a bug: the
the drawing is not visible during the dragging and it is visible
at the end. The required behaviuor is the opposite, the drawing must
be visible during the dragging and must be erased at the end.

This commit fix this. To test try in the listener the following
functions that use DRAGGING-DRAWING:

CLIME:POINTER-PLACE-RUBBER-BAND-LINE*
CLIME:POINTER-INPUT-RECTANGLE*


scymtym (approved):
Looks good to me.

PR #1115, opened 2020-12-21
Add tooltips extension

fmgilham:
Hello, this code was donated to McCLIM by SRI many years ago. It didn't work because of some coordinate weirdness. Seems like it works now except for a problem with scrolling windows.

The tooltips.lisp file has an example at the end that interacts with the "german towns" example in the demos.

To see more general usage, in the example code change the three #+(and) to #-(and) and the three #-(and) to 
#+(and) and the tooltips will come up with any output records.

To see the problem, use the general code with the border styles test. Scroll down near the bottom and move the mouse over the entries. The window will spontaneously scroll back to the beginning.

Apart from this the code looks useful.

### DIFF
--- a/Extensions/tooltips/README.txt
+++ b/Extensions/tooltips/README.txt
@@ -1,0 +1,122 @@
+* Tooltip Documentation
+
+  To use tooltips with clim presentations, define after methods for
+  the clim presentation-method clim:highlight-presentation. When the
+  state argument is :highlight, the method should call draw-tooltip.
+  When the state argument is :unhighlight, the method should call
+  erase-tooltip.
+
+** Example Code
+
+   Here is a worked example that adds tooltips to the German Towns
+   example and the Address Book example.
+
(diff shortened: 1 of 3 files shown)


dkochmanski:
thanks, this looks useful. I think that this code should use some cleanup before merging. Also, I wonder whether it wouldn't work better as an internal interface with better integration. I've managed to trigger a race condition with erase-tooltip.

dkochmanski:
Also, if this is going to be included as an extension, I think that it would deserve a proper demo and documentation (extending german towns example would be enough).

dkochmanski (changes requested):
as mentioned in the comment -- I like what this does, however this requires some work for better integration and discoverability.

fmgilham:
Hello, I cleaned up the tooltip code and added a README .txt file. There is a more polished demo, described in the README.txt file and commented out in the tooltips.lisp file.

Issue #1118, opened 2020-12-27
tracking-pointer keyboard clause


admich:
In tracking-pointer when multiple-window is nil  and the tracked sheet isn't the keyboard focus the keyboard clause is not executed. For this reason in "Tracking Pointer Test" and "DND varius" examples the tracking of the pointer is not terminated when the SPACE key is pressed.

I think to have a solution and I will prepare a PR.

Issue #899, opened 2019-11-09
Drag & Drop across multiple application panes


deterministic-arts:
Please see the attached example. Drag & drop seems to work only within a single application pane: I can drag "mud" onto a "puddle" in the "palette" pane just fine, and the same is true in the "desktop" pane. But when I try to drag a "mud" from the palette pane onto a "puddle" from the "desktop" pane (or vice versa), nothing happens when I release the pointer button. 

While dragging the object, the operation seems to happen, even across the panes (i.e., while holding the button, I am dragging the object along with the pointer, highlighting of the drop site happens, etc.) but when I release the button, the command is not executed in the cross-panes case.

admich:
I have the same problem. 
The probelm is that the tracking pointer code (when multiple-window is t) receive the pointer events two times: one in the tracked sheet coordinate and another in the coordinate of the sheet where the pointer is over. The first is useless because with it it is not possible to find the presentation under the pointer. With the second the code is able to highlight the right presentations but of the last release button event only the first event is received and so it isn't possible to find the presentation to throw.

PR #1119, opened 2020-12-27
tracking-pointer: intercept keyboard event

admich:
If a :keyboard clause is present in the tracking-pointer body than
also the keyboard events needs to be tracked.

Fix: #1118

### DIFF
--- a/Core/clim-basic/extended-streams/pointer-tracking.lisp
+++ b/Core/clim-basic/extended-streams/pointer-tracking.lisp
@@ -191,6 +191,7 @@
               ;; the other hand, we pass events for TRACKED-SHEET (or
               ;; all events if MULTIPLE-WINDOW is true) to TRACK-EVENT.
               do (cond ((not (or multiple-window
+                                 (and (typep event 'keyboard-event) (keyboard-handler state))
                                  (eql tracked-sheet (event-sheet event))))
                         ;; Event is not intercepted.
                         (handle-event (event-sheet event) event))


Core/clim-basic/extended-streams/pointer-tracking.lisp
```diff
@@ -191,6 +191,7 @@
               ;; the other hand, we pass events for TRACKED-SHEET (or
               ;; all events if MULTIPLE-WINDOW is true) to TRACK-EVENT.
               do (cond ((not (or multiple-window
+                                 (and (typep event 'keyboard-event) (keyboard-handler state))
                                  (eql tracked-sheet (event-sheet event))))
```
dkochmanski, line 195: I think that a solution that fits better how the tracking-pointer is implemented would invovle:
- extending with-pointer-grabbed to accept `:keyboard t` (that is, change port-grab-pointer - for clx that would be adding the event to the event mask)
- wrapping the body in `with-input-focus` on a sheet

dkochmanski (changes requested):

Core/clim-basic/extended-streams/pointer-tracking.lisp
admich: I think taht  the second suggestion (`with-input-focus`) is better: in principle to track the pointer on a single window grabbing the pointer isn't necessary . 
Anyway I think that that `port-grab-pointer` needs a review (see #1120 and #899)

admich:
I close this PR because I follow the suggestion of dkochmanski and I open a new PR #1164  using `with-input-focus`.

PR #1146, opened 2021-02-05
Don't execute a disabled command

### LINKED ISSUE #1144: Disabled command
In McCLIM when a command is not enabled it can still be executed from keystroke or clicking on a command presentation.
In the following application after the `com-disable-1` is executed the `com-print-1` is not available from the interactor line but it can still be executed with the keystroke C-1 or with click on the previous "Print 1" on the interactor pane.
```
(in-package :clim-user)

(define-application-frame disable-commands ()
  ()
  (:panes
    (inter :interactor :width 400 :height 200))
  (:menu-bar nil)
  (:pointer-documentation t)
  (:layouts (default
                (vertically ()
                  inter))))

(defun disable-commands (&key new-process)
  (setf *frame* (make-application-frame 'disable-commands :width 1000 :height 800))
  (if new-process
      (clim-sys:make-process (lambda () (run-frame-top-level *frame*)))
      (run-frame-top-level *frame*)))

(define-disable-commands-command (com-print-1 :name t :keystroke (#\1 :control))
    ()
  (format (frame-query-io *application-frame*) "One"))

(define-disable-commands-command (com-print-2 :name t :keystroke (#\2 :control))
    ()
  (format (frame-query-io *application-frame*) "Two"))

(define-disable-commands-command (com-enable-1 :name t)
    ()
  (setf (command-enabled 'com-print-1 *application-frame*) t))

(define-disable-commands-command (com-disable-1 :name t)
    ()
  (setf (command-enabled 'com-print-1 *application-frame*) nil))
```

admich:
The spec (26.6.1) say that COMMAND presentation type represent
commands  accessible in COMMAND-TABLE presentation parameter and enabled
in *APPLICATION-FRAME*. The last requirement was not checked before this commit.

The spec (27.4) say that in LOOKUP-KEYSTROKE-COMMAND-ITEM
"only keystrokes that map to an enabled application command will be
matched."

In this way  disabled commands aren't executed.

Fix #1144 issue

### DIFF
--- a/Core/clim-core/commands/tables.lisp
+++ b/Core/clim-core/commands/tables.lisp
@@ -502,12 +502,14 @@ menu item to see if it is `:menu'."
                          (otherwise nil))))
     ;; Return a literal command, or create a partial command from a
     ;; command-name.
-    (return-from lookup-keystroke-command-item
-      (substitute-numeric-argument-marker
-       (if (symbolp command)
-           (partial-command-from-name command command-table)
-           command)
-       numeric-arg)))
+    (when (or (null *application-frame*)
+              (command-enabled (command-name command) *application-frame*))
+      (return-from lookup-keystroke-command-item
+        (substitute-numeric-argument-marker
+         (if (symbolp command)
+             (partial-command-from-name command command-table)
+             command)
+         numeric-arg))))
   gesture)
 
 (defun map-over-command-table-translators
--- a/Core/clim-core/commands/processor.lisp
+++ b/Core/clim-core/commands/processor.lisp
@@ -346,7 +346,9 @@
 
 (define-presentation-method presentation-typep (object (type command))
   (and (consp object)
-       (command-accessible-in-command-table-p (car object) command-table)))
+       (command-accessible-in-command-table-p (car object) command-table)
+       (or (null *application-frame*)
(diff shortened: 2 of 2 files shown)


Core/clim-core/commands/tables.lisp
```diff
@@ -502,12 +502,15 @@ menu item to see if it is `:menu'."
...
     ;; command-name.
-    (return-from lookup-keystroke-command-item
-      (substitute-numeric-argument-marker
-       (if (symbolp command)
-           (partial-command-from-name command command-table)
-           command)
-       numeric-arg)))
+    (when (or (null *application-frame*)
+              (command-enabled (if (consp command) (command-name command) command)
+                                  *application-frame*))
```
dkochmanski, line 507: command-name already checks whether the comand is a cons or not, so we may skip the check here.

dkochmanski (changes requested):
putting aside discussion on #clim, this looks fine to me (I had one remark). We may first merge this and later  (eventually) add a new parameter for the command.

Core/clim-core/commands/tables.lisp
admich: Done.

dkochmanski (approved):

dkochmanski:
merged, thanks!

PR #1149, opened 2021-02-07
command-tables: fix the inherited menu items

admich:
Without this fix a command table inherit the menu-item of the ancestors
also when `:inherit-menu` is `:keystrokes`


dkochmanski:
lgtm, thanks!

PR #1164, opened 2021-03-18
tracking-pointer: keyboard events

admich:
This PR fix the handling of keyboard events during the tracking of the pointer.
This PR supersedes the PR #1119 following the suggestion of dkochmanski of using with-input-focus.

Fix: #1118


dkochmanski:
lgtm, thanks!

PR #1165, opened 2021-03-19
Pointer

admich:
This PR didn't introduce new features and didn't fix any bugs. With this PR McCLIM follows better the specification:

1) The pointer events have the slot pointer bound to a pointer object and not to 0 as before
2) The pointer sheet is stored in a slot of the pointer and not of the port
3) `pointer-modifier-state` is removed because it  isn't in the specification and it is not used

### DIFF
--- a/Backends/Null/port.lisp
+++ b/Backends/Null/port.lisp
@@ -27,8 +27,8 @@
 
 (defclass null-port (basic-port)
   ((id)
-   (pointer :accessor port-pointer :initform (make-instance 'null-pointer))
-   (window :initform nil :accessor null-port-window)))
+   (window :initform nil :accessor null-port-window))
+  (:default-initargs :pointer (make-instance 'null-pointer)))
 
 (defmethod find-port-type ((type (eql :null)))
   (values 'null-port 'identity))
--- a/Core/clim-basic/extended-streams/stream-input.lisp
+++ b/Core/clim-basic/extended-streams/stream-input.lisp
@@ -640,7 +640,6 @@ known gestures."
   ((port :reader port :initarg :port)
    (state-lock :reader state-lock :initform (make-lock "pointer lock"))
    (button-state :initform 0 )
-   (modifier-state :initform 0)
    (sheet :initform nil :initarg :sheet :accessor pointer-sheet)))
 
 (defgeneric pointer-sheet (pointer))
@@ -648,8 +647,6 @@ known gestures."
 (defgeneric (setf pointer-sheet) (sheet pointer))
 
 (defgeneric pointer-button-state (pointer))
-
-(defgeneric pointer-modifier-state (pointer))
 
 (defgeneric pointer-position (pointer))
 
@@ -665,15 +662,6 @@ known gestures."
 (defmethod pointer-button-state ((pointer standard-pointer))
   (with-lock-held ((state-lock pointer))
     (slot-value pointer 'button-state)))
-
-(defmethod pointer-modifier-state ((pointer standard-pointer))
-  (with-lock-held ((state-lock pointer))
-    (slot-value pointer 'modifier-state)))
-
-(defmethod pointer-update-state
-    ((pointer standard-pointer) (event keyboard-event))
-  (with-lock-held ((state-lock pointer))
-    (setf (slot-value pointer 'modifier-state) (event-modifier-state event))))
 
 (defmethod pointer-update-state
     ((pointer standard-pointer) (event pointer-button-press-event))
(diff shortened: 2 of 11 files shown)


Backends/Null/port.lisp
```diff
@@ -27,8 +27,8 @@
 
 (defclass null-port (basic-port)
   ((id)
-   (pointer :accessor port-pointer :initform (make-instance 'null-pointer))
-   (window :initform nil :accessor null-port-window)))
+   (window :initform nil :accessor null-port-window))
+  (:default-initargs :pointer (make-instance 'null-pointer)))
```
dkochmanski, line 31: what is this null-port-window? I don't see any mention in the commit message about it, nor it is referenced elsewhere in this commit.

Core/clim-core/presentations/translators.lisp
```diff
@@ -457,17 +457,11 @@ and used to ensure that presentation-translators-caches are up to date.")
         (loop for context in input-context
               do (mopscp context presentation)))))
 
-(defun window-modifier-state (window)
-  "Provides default modifier state for presentation translator functions."
-  ;; There are many things which may go wrong (NULL window, sheet
-  ;; without a port etc). We clamp errors to 0 meaning "no modifiers".
-  (or (ignore-errors (pointer-modifier-state (port-pointer (port window)))) 0))
```
dkochmanski, line 464: I understand that it was always factually 0, but should it be? shouldn't we update the pointer modifier state?

Core/clim-basic/extended-streams/stream-input.lisp
```diff
@@ -666,15 +663,6 @@ known gestures."
...
     (slot-value pointer 'button-state)))
 
-(defmethod pointer-modifier-state ((pointer standard-pointer))
-  (with-lock-held ((state-lock pointer))
-    (slot-value pointer 'modifier-state)))
-
-(defmethod pointer-update-state
-    ((pointer standard-pointer) (event keyboard-event))
-  (with-lock-held ((state-lock pointer))
-    (setf (slot-value pointer 'modifier-state) (event-modifier-state event))))
```
dkochmanski, line 678: I can tell from this commit, that the pointer-modifier-state could change by means of pointer-update-state (compare to the previous commit where you remove window-modifier-state).

Core/clim-core/presentations/translators.lisp
dkochmanski: or use port-modifier-state. I imagine that this is used for the pointer documentation, so when control is pressed I'd expect that the modifier state is set up accordingly. It seems from the next commit, that pointer-modifer-state may change by means of the method pointer-update-state.

dkochmanski (changes requested):

Backends/Null/port.lisp
admich: I don't know what is `null-port-window`, it  was already there I didn't add it, I only removed `pointer` slot because I add it to `basic-port`. 
I think it is there because clx port have a `clx-port-window` where it is stored the xwindow of the root.
Anyway if you want we can remove it, it is never used and it is not in the specification.

Core/clim-core/presentations/translators.lisp
admich: I was in doubt to use `port-modifier-state` or remove this function, I opt for the second because now the pointer documentation works good despite this function return always 0 so I deduce that the pointer documentation machinery doesn't need this function.

Core/clim-basic/extended-streams/stream-input.lisp
admich: Yes but `pointer-update-state` is never called so `pointer-modifier-state` never changed. The modifier state is already returned by `port-modifier-state` that works and its defined in Franz manual, so I remove `pointer-modifier-state` that seems a not used, not working and not defined in spec. replica of  `port-modifier-state`. 

 The only place where `pointer-modifier-state` was used was `window-modifier-state` that I removed (instead of modify it with the using of `port-modifier-state`) for the reason written in the previous comment.

Backends/Null/port.lisp
dkochmanski: right, I've misread the diff - sorry.

Core/clim-core/presentations/translators.lisp
dkochmanski: currently we synthesize the motion event for the pointer documentation when necessary.

dkochmanski:
merged after rebasing, thanks!

PR #1167, opened 2021-03-25
Frame managers

dkochmanski:


dkochmanski:
@admich in this commit I've changed the protocol class of a menu frame to the `application-frame`, I think that you were mentioning a need for that.

PR #1168, opened 2021-03-27
stream-write-string is supposed to return the string passed to:

kpoeck:
Current clasp fails with the address-book demo displaying the names of the addresses. The reason is obviously that `(write-string "Scott McKay")` returns nil, while present expects a string.

This happens since 
`(defmethod stream-write-string ((stream standard-extended-output-stream) string
                                &optional (start 0) end)...` returns nil

Looking at the documentation for stream-write-string it says:
```lisp
— Generic Function: sb-gray:stream-write-string stream string &optional start end
This is used by write-string. It writes the string to the stream, optionally delimited by start and end, which default to 0 and nil. The string argument is returned.
````


dkochmanski:
Thanks, I agree that this is an omission. I don't think that this deserves an explicit comment, so I've removed it.

PR #1169, opened 2021-04-04
Small patches 2

death:


dkochmanski:
lgtm, thanks!

dkochmanski:
I've reverted a90bcce91f4104dd5736e24f10465c6f4f04abb9 because it causes a regression in accepting values:
- run menu test
- go to "about/more" menu item
- type in the accepting field something
- press `C-[` to confirm

without this commit it does work.

death:
Confirmed.

I don't yet understand why this happens. It seems `read-command-using-keystrokes` passes a gesture to `ensure-complete-command` (instead of returning it? the spec says it may return a gesture) but that's not the whole story.

The additional gesture also creates issues for `complete-input` and `complete-input-rescan`, because they expect gestures to be characters and `vector-push-extend` them to a string.  Likely they should test before pushing.

dkochmanski:
Yes, it seems unclear what's going on - ot is likely some deeper issue with in stream-input,maybe the function process-gesture?

Wysłano z aplikacji ProtonMail

-------- Oryginalna wiadomość --------
‎ 10 kwi 2021, 10:20, death napisał(a):

death:
There seems to be plenty of overloading going on in McCLIM (and CLIM spec) with regards to concepts like gestures, gesture names, characters, and events.

What happens is that `finalize-query-record` chooses one of the activation gesture names (`:newline`), passes it to `character-gesture-name` (not a terribly good name; it seems to return a character or a key name corresponding to the gesture name) which returns `:kp-enter`, which is then set as the `activation-gesture` of the standard input editing stream.

Later on `activation-gesture-p` gets called with this value, and returns false. It uses `gesture-matches-spec-p` to try and match `:kp-enter` with each activation gesture name. Since the latter (the "spec") is a symbol, `event-matches-gesture-name-p` is called. According to the CLIM spec, this function takes a device event. In McCLIM, however, it seems characters may also be passed.

The function `event-matches-gesture-name-p` calls `event-matches-gesture-p` with the physical gestures corresponding to the gesture name. The `event-matches-gesture-p` has a method specializing on `character` that attempts to find a matching physical keyboard gesture by comparing against the physical gesture's device-supplied name.  Since we have a symbol `:kp-enter`, however, the default method gets called and just returns false.

We could add a method to `event-matches-gesture-p` that specializes on symbols and does the same as the method specializing on characters. This fixes your test case. It may also make sense to warn about unexpected "event" object in the default method.

If we're to discourage overloading concepts, we could start by adding `check-type` forms to functions from the CLIM spec, one by one, and fix the ensuing breakage. Likely the right thing to do, but can take a lot of work.

PR #1172, opened 2021-04-07
attempt to make accepting-values dialogs work

death:
The relevant test-case is `ACCEPTING-WITH-LIST-PANE-VIEW`.

First, `REPLAY` should not call `PANE-VIEWPORT-REGION` on streams that are not panes. With this change, `ACCEPTING-WITH-LIST-PANE-VIEW` works when not run in its own window.

When an own-window is used, gadget-related events were not handled until the `ACCEPTING-VALUES` loop ended. In this patch I attempt to fix this by dispatching frame events in `ACCEPTING-VALUES-DEFAULT-COMMAND`. Is this the right fix?


dkochmanski:
the pane viewport fix doesn't seem to be necessary anymore (we have a sheet specialization that returns nil when there is no viewport); I've tested the issue on master and it seems that gadget events are handled just fine when it is in its own window, so I assume that it was fixed somewhere else in the meantime. I'm closing the pr, but if you feel that the issue is still not addressed then please reopen. and sorry for the long delay.

Issue #1171, opened 2021-04-06
Menus flash and some items can't be selected if user clicks without releasing


john-a-carroll:
```
(in-package :clim-user)

(define-application-frame menutest ()
    ()
  (:pane :application))

(define-menutest-command com-item ()
  (format *trace-output* "~&com-item~%"))

(define-command-table file-command-table
                      :menu (("File 1" :command com-item)
                             ("File 2" :command com-item)
                             ("File 3" :command com-item)
                             ("File 4" :command com-item)))

(define-command-table edit-command-table
                      :menu (("Edit 1" :command com-item)
                             ("Edit 2" :command com-item)
                             ("Edit 3" :command com-item)
                             ("Edit 4" :command com-item)))

(define-command-table windows-command-table
                      :menu (("Windows 1" :command com-item)
                             ("Windows 2" :command com-item)
                             ("Windows 3" :command com-item)
                             ("Windows 4" :command com-item)))

(define-command-table menutest
                      :menu (("File" :menu file-command-table)
                             ("Edit" :menu edit-command-table)
                             ("Windows" :menu windows-command-table)))

(run-frame-top-level (make-application-frame 'menutest))
```

The menus behave very nicely when they are opened with a mouse button click and immediate release. However, if the user clicks without releasing and then drags across the menu bar, the menus flash disconcertingly and after a while it becomes impossible to select items from some of them because the wrong menu gets displayed.

dkochmanski:
The issue comes from the fact, that:

1) we grab a sheet in tracking-pointer with multiple-window (port-grabbed-sheet is t)
2) the sheet is pressed so it receives event copies

I think that the correct solution will be to not treat a pressed sheet especially (i.e gadgets that want it to be treated specially will need to pointer-grab the sheet), but this change will be quite impactful:
- it changes behavior of distribute-event
- it requires identifying all gadgets that depend on the pointer being grabbed
- extra care needs to be put for sheets that doesn't have a direct mirror

dkochmanski:
when I simply comment out distribution of the event copy in the method distribute-event then menu bar doesn't flicker and everything works as expected (but for example sliders doesn't work as before).

dkochmanski:
I've pushed a conservative change that addresses this issue. It may be superseded by #1177 in the future.

PR #1177, opened 2021-04-13
Admich grab pointer

admich:
This PR improve pointer tracking and drag and drop that now works on multiple window.

Fix: #899


dkochmanski:
This pr will also fix "flashing" in menu bar, as reported in #1171

dkochmanski:
I've noticed a regression: when I have a slider and drag it, and there is another mirror along the way, then dragging stops. I think that's one of a few valid cases for single-window grabbing.

xxx

dkochmanski:
also notice, that a few operators allow specifying multiple-window as an argument; I think that we should get rid of the "pressed-sheet" copy distribution but that needs to be replaced with grabbing that is correctly handled in both full and single mirror sitations.

admich:
The problem isn't single/full mirror, I didn't try but I think that the problem remains also with a single mirror port. The problem is that slider never grab the pointer. I see two possible solutions:

1) change slider code in such a way that when handle button-press it starts to tracking the pointer in multiple window way. Of course we need to review also other gadgets that could have similar problems.
2) At the button press the pressed sheet automatically grab the pointer (as I understand it is what x11 does). 

One problem to solve it is that sometimes for the grabbing sheet is useful the original event in the coordinate of the original pointer sheet (e.g. when you want to do some stuff with presentation), and sometimes is useful the event in the grabbing sheet coordinate (like in the slider case). So we need to understand how and when the code choose between this two cases.

dkochmanski:
I've mentioned single/full mirror because assuming that we use x11 grabbing then:
1. un unmirrored sheet is grabbed
2. x11 grabs the mirror of its ancestor
3. we need to distribute events not to the mirrored ancestor but to the grabbed sheet
I think that we should rely on the interface port-grab-pointer to some extent (and make it accept multiple-window argument).

Regarding solution 1. you mention: that is the purpose of the `pressed-sheet` slot; that said, I think that we should resign from the pressed sheet and rely on grabbing and correctly distribute the event.

PR #1197, opened 2021-06-14
Dot-based graph layout

daewok:
Add support for using external tools that speak the [DOT language](https://www.graphviz.org/doc/info/lang.html) to perform graph layout.

In order to use, simply use the newly defined `:dot-digraph` graph type. If you also use your own `arc-drawer`, note that an additional keyword argument `:splines` is also provided. This is a list of `dot-spline` instances that describe the laid out edges as cubic B-splines.

Adds several systems to the Extensions folder.

+ `mcclim-dot/core` provides all the data structures and GFs needed.
+ `mcclim-dot/graphviz` adds the capability to use the [Graphviz](https://www.graphviz.org/) software suite as an external process to perform the layout.
+ `mcclim-dot` depends on both the above systems.

This separation is done so that it's possible to use only the core data structures and GFs and provide your own DOT layout tool (such as via CFFI or some other interface).

The [cl-dot](https://github.com/michaelw/cl-dot) project is used for their DOT data structures and ability to serialize.

Before this is ready for review, I'd like to address the following:

+ [x] Add support for labels on edges.
+ [x] Stop using internal symbols from cl-dot. Preferably by getting them to export the appropriate symbols.
+ [ ] ~~Add support for non directed graphs.~~
+ [x] Determine if passing the points of a cubic B-spline (DOT's specified output) as-is to McCLIM's Bezier interface is dodgy. (Seems to be reasonable on all the tests I've done).

Potential future changes include:

+ [ ] Exposing support for more of DOT's layout options. For example, support for subgraphs to ensure certain nodes are always laid out together.


Extensions/dot/mcclim-dot.asd
```diff
@@ -0,0 +1,38 @@
...
+  :license "MIT"
+  :description "Support for using DOT based graph layout engines."
+  :depends-on ("mcclim-dot/core" "mcclim-dot/graphviz")
+  :in-order-to ((test-op (test-op "mcclim-dot/test"))))
+
+(defsystem "mcclim-dot/core"
+  :license "MIT"
+  :description "Core DOT routines."
+  :depends-on ("mcclim"
+               "mcclim-bezier"
```
dkochmanski, line 13: you may want to use a more complete alternative that has been merged recently into core: draw-bezigon* et al.

daewok: Nice. Done.

Extensions/dot/mcclim-dot.asd
```diff
@@ -0,0 +1,37 @@
...
+  :license "MIT"
+  :description "Support for using DOT based graph layout engines."
+  :depends-on ("mcclim-dot/core" "mcclim-dot/graphviz")
+  :in-order-to ((test-op (test-op "mcclim-dot/test"))))
+
+(defsystem "mcclim-dot/core"
+  :license "MIT"
+  :description "Core DOT routines."
+  :depends-on ("mcclim"
+               (:version "cl-dot" "0.9")
```
dkochmanski, line 13: if I understand this correctly, all needed symbols are available in the library; perhaps to not depend on a version that is not yet available in ql, this commit (c4e486f5a611dda6ff5dd90aa1f3f5726fe00821) should be reverted? I don't have a strong opinion about this.

dkochmanski (changes requested):
I've tested this after merging against upstream master and it works fine for me. I was a bit confused, because the error when the "dot" command is not found is an obscure one (something that NIL is not of type string down the line), so perhaps this should be handled earlier (i.e probe-file command or something) and error with more intelligible message.

dkochmanski:
missing dependencies are now available in quicklisp so after some extra testing I've merged this pull request, thanks!

PR #1201, opened 2021-07-07
frames: graft method for application-frame

admich:
The CLIM Specification says that GRAFT is defined for application
frames but before this commit the method wasn't defined.


dkochmanski:
lgtm, thanks!

PR #1202, opened 2021-07-10
frame-panes and pointer-documentation fix

admich:
ADOPT-FRAME-PANES creates (when required) the menu and
pointer-documentation pane, but the frame-panes weren't updated
accordingly. In this way the function UPDATE-FRAME-PANE-LISTS didn't
find the pointer-documentation pane that therefore didnt't work.


dkochmanski:
lgtm, thanks!

PR #1203, opened 2021-07-12
Geometry improvements

dkochmanski:


dkochmanski:
after talking with the pr author I'm convinced that this is what we need, lgtm! ;)

PR #1209, opened 2021-08-20
Geometry improvements

dkochmanski:
Implement the ellipse polygonalization.


dkochmanski:
converting to draft until I figure out whether we could speed this up (and compare with approximation with bezier, then polygon).

dkochmanski:
going through bezier to polyline gives us various glitches because of how clx render this thing. perhaps when we switch to xrender that will make more sense.

PR #1211, opened 2021-08-31
Frame state

admich:
With this PR the frame state is updated when the frame state change due to user interaction with the  window manager.
Furthermore it is possible to start an application frame in shrunk state.

### DIFF
--- a/Backends/CLX/input.lisp
+++ b/Backends/CLX/input.lisp
@@ -223,9 +223,14 @@
                              :sheet sheet
                              :x x :y y :width width :height height))))
       (:map-notify
-       (make-instance 'window-map-event :sheet sheet))
+       (if (and (typep sheet 'top-level-sheet-pane)
+                (eq (frame-state (pane-frame sheet)) :shrunk))
+           (make-instance 'window-manager-deiconify-event :sheet sheet)
+           (make-instance 'window-map-event :sheet sheet)))
       (:unmap-notify
-       (make-instance 'window-unmap-event :sheet sheet))
+       (if (= 3 (car (xlib:get-property window :WM_STATE)))
+           (make-instance 'window-manager-iconify-event :sheet sheet)
+           (make-instance 'window-unmap-event :sheet sheet)))
       (:destroy-notify
        (make-instance 'window-destroy-event :sheet sheet))
       (:motion-notify
--- a/Backends/CLX/input.lisp
+++ b/Backends/CLX/input.lisp
@@ -228,7 +228,7 @@
            (make-instance 'window-manager-deiconify-event :sheet sheet)
            (make-instance 'window-map-event :sheet sheet)))
       (:unmap-notify
-       (if (= 3 (car (xlib:get-property window :WM_STATE)))
+       (if (= +icccm-iconic-state+ (car (xlib:get-property window :WM_STATE)))
            (make-instance 'window-manager-iconify-event :sheet sheet)
            (make-instance 'window-unmap-event :sheet sheet)))
       (:destroy-notify
--- a/Backends/CLX/mirror.lisp
+++ b/Backends/CLX/mirror.lisp
@@ -1,5 +1,9 @@
 (in-package #:clim-clx)
 
+(defconstant +icccm-withdrawn-state+ 0)
+(defconstant +icccm-normal-state+ 1)
+(defconstant +icccm-iconic-state+ 3)
+
 (defclass clx-mirror ()
   ((window
     :initarg :window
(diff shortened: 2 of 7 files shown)


Backends/CLX/input.lisp
```diff
@@ -223,9 +223,14 @@
...
                              :x x :y y :width width :height height))))
       (:map-notify
-       (make-instance 'window-map-event :sheet sheet))
+       (if (and (typep sheet 'top-level-sheet-pane)
+                (eq (frame-state (pane-frame sheet)) :shrunk))
+           (make-instance 'window-manager-deiconify-event :sheet sheet)
+           (make-instance 'window-map-event :sheet sheet)))
       (:unmap-notify
-       (make-instance 'window-unmap-event :sheet sheet))
+       (if (= 3 (car (xlib:get-property window :WM_STATE)))
```
dkochmanski, line 231: perhaps we should define appropriate constants (i.e in clx) for states - given they are standard across different x11 implementations - of course if they are not already defined, if they are then we just use them.

Backends/CLX/mirror.lisp
```diff
@@ -122,6 +122,21 @@
...
+       (let ((hints (xlib:wm-hints window)))
+          (setf (xlib:wm-hints-initial-state hints) :iconic)
+          (setf (xlib:wm-hints window) hints)
+          (xlib:map-window window)
+          (xlib:display-force-output (clx-port-display port))))
+      ((eq 1 (car (xlib:get-property window :WM_STATE)))
+       (xlib:iconify-window window (clx-port-screen port))
+       (xlib:display-force-output (clx-port-display port)))
+      ((eq 2 (car (xlib:get-property window :WM_STATE)))
+       (warn "The sheet is already shrunk")))))
```
dkochmanski, line 142: please access the property only once by using let:
```
(let ((wm-state (car (xlib:get-property window :wm_state)))) …)
```

dkochmanski: also, do we want to signal a warning here? perhaps leaving this as a no-op without further notification would be less engaging for the programmer without much demerit.

Backends/CLX/mirror.lisp
```diff
@@ -122,6 +122,14 @@
     (xlib:display-force-output (clx-port-display port))))
 
 (defmethod port-shrink-sheet ((port clx-basic-port) (sheet mirrored-sheet-mixin))
-  (when-let ((mirror (sheet-direct-mirror sheet)))
-    (xlib:iconify-window (window mirror) (clx-port-screen port))
-    (xlib:display-force-output (clx-port-display port))))
+  ;; Not all x11 window-managers iconify windows (in particular many
```
dkochmanski, line 125: this method change illustrates a certain problematic pattern with the clim spec where user interactions and port interactions are somewhat contrived (I'll use shrink frame as an example, but this applies i.e to the window configuration event and resize-sheet too):

User: shrink-frame -> shrink-sheet -> port-shrink-sheet
Port: (handle-event iconify-event) -> shrink-sheet -> port-shrink-sheet

Using both leads to a situation where we may end in an infinite loop, because calling port-shrink-sheet could generate iconify-event that will lead to calling port-shrink-sheet again etc etc. In order to avoid such infinite loop CLIM-TOS has a keyword argument in a few functions names `port-did-it` that is `T` when the port generates the event and `NIL` when it is a result of the user interaction to distinguish both situations. McCLIM on the other hand (in the case of the configuration event) binds a dynamic variable `*configuration-event-p*` that inhibits calling `port-set-mirror-geometry` (in order to avoid the infinite loop). I wonder what uniform approach to this problem we should adopt.

Regarding this case, as noted it is wasteful to go through get-property twice in case when it is the port what leads to this call (via handle-event).

dkochmanski (changes requested):
Thanks, I have minor stylistic remarks (we may define constants in clim-clx too instead of clx, I guess it would be fine either way) and an open discussion regarding situations where we have insufficient api to distinguish between port-triggered and user-triggered window state changes.

dkochmanski:
other than that this pr look good to me.

Backends/CLX/mirror.lisp
admich: done --also 2 was the wrong constant, anyway now there is +icccm-iconic-state+ constant

Backends/CLX/mirror.lisp
admich: Yes, for the moment I did not introduce any of the suggested approach (anyway there isn't infinite loop in this case), and I don't have opinion about what is better. 

I also don't understand exactly what is the right things to do in main gf (shrink-frame) and what to do in note-...... .

I also tried to check at the end of shrink-frame if the application is really shruk but it no so easy because the x11 mesages are intercepted from another thread (the port thread).

Backends/CLX/input.lisp
admich: I didn't find the constant in clx so I defined them in clim-clx

Backends/CLX/mirror.lisp
dkochmanski: OK, let's leave this problem to the future selves.

dkochmanski (approved):

dkochmanski:
lgtm, thanks!

PR #1212, opened 2021-09-08
clim-examples: graph-toy: Prevent error upon exiting.

### LINKED ISSUE #1192: closing "Data Graph Toy" demo lands in debugger
with error:
```lisp
  #<STANDARD-GENERIC-FUNCTION CLIM:PANE-NEEDS-REDISPLAY (4)>
when called with arguments
  (NIL).
   [Condition of type SB-PCL::NO-APPLICABLE-METHOD-ERROR]
See also:
  Common Lisp Hyperspec, 7.6.6 [:section]

Restarts:
 0: [RETRY] Retry calling the generic function.
 1: [CLEAR-PANE-TRY-AGAIN] Clear the output history of the pane and reattempt forceful redisplay.
 2: [CLEAR-PANE] Clear the output history of the pane, but don't redisplay.
 3: [SKIP-REDISPLAY] Skip this redisplay.
 4: [ABORT] Return to application command loop.
 5: [RETRY] Retry SLIME REPL evaluation request.
 --more--

Backtrace:
  0: ((:METHOD NO-APPLICABLE-METHOD (T)) #<STANDARD-GENERIC-FUNCTION CLIM:PANE-NEEDS-REDISPLAY (4)> NIL) [fast-method]
  1: (SB-PCL::CALL-NO-APPLICABLE-METHOD #<STANDARD-GENERIC-FUNCTION CLIM:PANE-NEEDS-REDISPLAY (4)> (NIL))
  2: ((:METHOD CLIM:REDISPLAY-FRAME-PANE :AROUND (CLIM:APPLICATION-FRAME T)) #<CLIM-DEMO::GRAPH-TOY {1002D855C3}> CLIM-DEMO::MAIN-DISPLAY) [fast-method]
  3: ((:METHOD CLIM:HANDLE-EVENT (CLIM-DEMO::GRAPH-TOY CLIM-DEMO::NEW-VALUE-EVENT)) #<CLIM-DEMO::GRAPH-TOY {1002D855C3}> #<CLIM-DEMO::NEW-VALUE-EVENT {1005CAEA03}>) [fast-method]
  4: (CLIM-EXTENSIONS:SIMPLE-EVENT-LOOP #<CLIM-DEMO:DEMODEMO {100733ADC3}>)
```

johnlorentzson:
The graph toy demo would cause an error from trying to redisplay after
the frame had been closed. It now checks whether the frame is active
before handling an event.

Fixes #1192 
Please let me know if this is an undesirable solution. I considered doing a `RETURN-FROM` at the top to avoid changing the indentation but I thought that would be worse.


Examples/graph-toy.lisp
```diff
@@ -99,21 +99,22 @@
...
-            do (setf (elt (val-array frame) i) (elt (val-array frame) j)))
-      (setf (fill-pointer (val-array frame)) xval-count))
-    (redisplay-frame-pane frame 'main-display)
-    (incf (last-yval frame) (- (random 9) 4))
-    (when (member (frame-state frame) '(:enabled :shrunk))
-      (schedule-event (frame-top-level-sheet frame)
-                      (make-instance 'new-value-event :sheet frame
-                                                      :val (last-yval frame))
-                      0.5))))
+  (when (eq (frame-state frame) :enabled)
```
nsrahmad, line 102: `frame-state` could return 4 possible values, `:disowned`, `:disabled`, `:enabled` or `:shrunk`. We definitely do not want to handle events while `:disowned` or `:disabled`. You are sure about handling events when `:enabled`, then what do you think should happen when the frame state is `:shrunk`?

Examples/graph-toy.lisp
```diff
@@ -99,21 +99,22 @@
...
+    (let ((val-array-len (length (val-array frame)))
+          (xval-count (max-xvals frame)))
+      (when (>= val-array-len (* xval-count 2))
+        (loop for i from 0 to (- xval-count 1)
+           for j from (- val-array-len xval-count)
+           do (setf (elt (val-array frame) i) (elt (val-array frame) j)))
+        (setf (fill-pointer (val-array frame)) xval-count))
+      (redisplay-frame-pane frame 'main-display)
+      (incf (last-yval frame) (- (random 9) 4))
+      (when (member (frame-state frame) '(:enabled :shrunk))
```
nsrahmad, line 113: I guess this answers the above question. But if we already doing this check at the start, do we really need to check again here?

nsrahmad (changes requested):
Hi, what you did here is definitely better than `return-from`.

johnlorentzson:
When I saw the code the problem looked so simple that I didn't bother reading the entire method. My new commit is what I should have done in the first place.

Examples/graph-toy.lisp
```diff
@@ -104,15 +104,15 @@
...
     (when (>= val-array-len (* xval-count 2))
       (loop for i from 0 to (- xval-count 1)
-            for j from (- val-array-len xval-count)
-            do (setf (elt (val-array frame) i) (elt (val-array frame) j)))
+         for j from (- val-array-len xval-count)
+         do (setf (elt (val-array frame) i) (elt (val-array frame) j)))
       (setf (fill-pointer (val-array frame)) xval-count))
-    (redisplay-frame-pane frame 'main-display)
-    (incf (last-yval frame) (- (random 9) 4))
     (when (member (frame-state frame) '(:enabled :shrunk))
```
nsrahmad, line 110: So the current implementation is like:
1. update the frame state
2. check if frame state is valid
3. redisplay and schedule event

But suppose I scheduled the event, and then destroyed the frame. When the event will be handled, it will try to update the frame state before checking for valid state. Do we really want that? or should we check for valid state before doing anything in the method?

In code, should I get in trouble If I do this:
```lisp
;;; After loading clim-examples

(let ((frame (clim:make-application-frame 'clim-demo::graph-toy)))
  (clim-sys:make-process 
   (lambda () (clim:run-frame-top-level frame)))
  (sleep 0.3)
  (clim:destroy-frame frame)
  (setf (clim-demo::val-array frame) nil))
```

nsrahmad (changes requested):

Examples/graph-toy.lisp
johnlorentzson: I felt what I did was a reasonable solution, but since this is an example program I guess the root problem should be solved, not the bug in this particular program.

I personally think that scheduled events should be cancelled when the frame is destroyed but I'm not familiar enough with the event system to know for sure. Does that sound like a reasonable solution?

nsrahmad: That is reasonable.

nsrahmad:
All right, Thanks for fixing this bug @johnlorentzson .

PR #1217, opened 2021-10-07
improve the command table menu inheritance

dkochmanski:


nsrahmad:
@dkochmanski About `command-not-present` vs `command-not-accessible`, you seem to interpret it as if a command is not present in current command-table and :errorp is true, then it should signal `command-not-present`. If it possibly also looks into inherited command-tables, and the command is not found (and errorp...) then it should signal `command-not-accessible`.

Whereas my interpretation was:
If a command is not present in current command-table or any inherited command-tables, and errorp is true, then it should signal `command-not-present`.
if a command is present in current command-table or any inherited command-tables, but the command is currently disabled, and errorp..., then it should signal `command-not-accessible`.

dkochmanski:
Hey, thanks for taking a look into proposed changes. I take my understanding of a difference between `present` and `accessible` from the introduction to [`27.2 Command Tables`](http://bauhh.dyndns.org:8000/clim-spec/27-2.html)

> We say that a command is present in a command table when it has been added to that command table. We say that a command is accessible in a command table when it is present in that command table or is present in any of the command tables from which that command table inherits.

This is additionally reinforced by the specification of operators, i.e `find-command-from-command-line-name` is specified to signal `command-not-accessible` but `remove-command-from-command-table` signals `command-not-present` (because find looks also in inherited tables while remove removes only commands that are in the command table).

nsrahmad:
@dkochmanski Sorry about that, I misremembered something. you are right, it is actually very clear.

PR #1222, opened 2021-10-20
ttf fonts: map default text styles with cl-dejavu fonts

dkochmanski:
This saves us from a frequent issue that the system fonts can't be
found. cl-dejavu is available in quicklisp and contains prepackaged
dejavu fonts.

Changes:
- the variable *families/faces* always contains absolute pathnames
- fonts are autoconfigured in eval-when (not in asd file)
- ad-hoc fc-match implementation is removed as unnecessary

The variable *truetype-font-path* is preserved because it is used to
list the "system" fonts (and i.e to customize this path to load the
"user" fonts).

### DIFF
--- a/Backends/CLX/fonts-xrender.lisp
+++ b/Backends/CLX/fonts-xrender.lisp
@@ -16,9 +16,6 @@
 
 ;;;; Notes
 
-;;; You might need to tweak mcclim-truetype:*families/faces* to point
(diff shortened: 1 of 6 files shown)


nsrahmad:
We do need to think about what happens when we create an executable? what if that executable in then run in different environment? should we try to adjust path at runtime?

PR #1233, opened 2021-11-03
New command parser

dkochmanski:
~This merge request~ The first two commits don't introduce any ground-breaking changes in the behavior - it is purely for the internals. When a command is defined only one parser is defined and functions like `command-line-read-remaining-arguments-for-partial-command` utilize it.


dkochmanski:
(should be rebased and merged after #1232).

PR #1248, opened 2021-12-08
Use uppercase M for em conversion and allow real column values

yitzchak:
In modern typesetting an em is often just the absolute font size. This does not look good in pretty printing where using the traditional definition of the width of an uppercase "M" is better.

Also, the functions `pprint-indent` and `pprint-tab` allow Real-valued column values. The Gray stream protocol does not make any statement about the return value of `stream-line-column`. Returning a Real-valued column number will make proportional typesetting have better alignments.


dkochmanski:
looks good, thanks! I'll merge it when all tests pass.

PR #1259, opened 2022-01-19
[draft] Menu keyboard navigation

admich:
With this PR I add keyboard navigation to menu (menu as menu-choose not as menu gadget where it already exist).

To activate the keyboard navigation the user must use `menu-choose-from-drawer` with a drawer function that
returns three values:
1.  a list of navigable presentations,
2.  the default presentation (or nil),
3.  the keyword `:kbd-nav`
If the third value isn't `:kbd-nav` the keyboard navigation isn't activated and the values returned by the drawer function aren't used.

If this PR is merged in master is also necessary to document new features in the McCLIM manual.

### DIFF
--- a/Core/clim-core/dialogs/menu-choose.lisp
+++ b/Core/clim-core/dialogs/menu-choose.lisp
@@ -358,20 +358,37 @@ maximum size according to `frame')."
   "A list of gesture names that serve as additional abort gestures for
 `menu-choose-from-drawer'.")
 
-;; Spec function.
+;; McCLIM add the possibility to navigate between the items of a menu
+;; with the keyboard. To activate the keyboard navifgation the user
+;; must use MENU-CHOOSE-FROW-DRAWER with a drawer function that
+;; returns three values:
+;; 1. a list of navigable presentations,
+;; 2. the default presentation (or nil),
+;; 3. the keyword :kbd-nav
+;; If the third value isn't :kbd-nav the keyboard navigation isn't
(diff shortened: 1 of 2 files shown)


admich:
Possible improvements:
1. change used gesture (see #1260)
2. Now the highlighted item with the pointer and with the keyboard have the same visual aspect, maybe it is better to differentiate (e.g. for the keyboard selected item only an underline instead of a box)

Core/clim-core/dialogs/menu-choose.lisp
```diff
@@ -111,22 +110,26 @@
...
+                                         item :style
+                                         '(:sans-serif nil nil)))
+                              (with-output-as-presentation
+                                  (stream
+                                   item
+                                   `((,name ,@params)
+                                     :description ,(getf (menu-item-options item) :documentation)
+                                     ,@options)
+                                   :single-box t)
+                                (funcall item-printer item stream))))))
```
dkochmanski, line 129: I'm not sure whether draw-standard-menu should collect presentations, perhaps it should be frame-manager-menu-choose by making a closure over item-printer (and then estabilish the context with accelerators around the call to menu-choose-from-drawer).

dkochmanski (changes requested):
I've made an initial review with some preliminary comments. I'd still want to have another pass with it we talk through what was signaled here. Some of remarks are rather an opening for a discussion than suggestions for change.

Also, please rebase on top of master (doing that locally had no conflicts).

PR #1261, opened 2022-01-20
Add :resynchronize-every-pass t to address book accepting-values

alanz:
Context:running the Examples/address-book.lisp application.

Without this setting, with SBCL on Linux, Entering information for the
"Name" field and pressing ENTER takes you to the "Address" field, but
the "Name" field becomes blank.

With this setting the already entered values continue to be displayed.


dkochmanski:
makes sense, thanks!

PR #1264, opened 2022-01-26
present complex number

admich:
Two changes:

1. The imaginary unit i is printed
2. The inferior presentations of the real part and imaginary part are
sensitive so they can be used for typed input.


dkochmanski:
two questions:
- what's the point of making real and imag parts sensitive?
- perhaps more terse form would be better, like 1+2i  3-2i (without spaces)

admich:
1.  why not? For examples: after a complex has been presented maybe you want to accept a real and you can click on the imag or real part to have exactly the same real number. Or you want to accept another complex and maybe you need a complex with the same imag part but a different real part of the previous presented complex.
2. Without space is aesthetically better but with space the presentation is acceptably in the sense that without further code the following works:
```
(accept-from-string 'complex (present-to-string 'complex #C(3 2)))
```

dkochmanski:
the "default" accept method still doesn't "eat" I, so if we ever fix incremental parsing from strings (i.e accept will return correctly where it finished), then we'll have garbage in front of the reminder.

When I have presented a complex number I don't think that accepting its both parts as reals makes semantic sense too. Unless there is a convincing use case for this change I don't think that we should incorporate it.

admich:
I agree with you that it is necessary to change also accept method in order to not have garbage.

For the other point i think that if the specification allow nested presentations it is good to use them. 

Anyway I close this PR and and I will open an issue to remember us to improve complex presentation because present #C(1 2) as 1 2 is ugly.

admich:
I have tried opengenera and there the complex are presented as: `#C(1 2)` with the real and imaginary parts that are  selectable for accept real.

PR #1265, opened 2022-01-30
Fix various

dkochmanski:


dkochmanski:
failures in the algebra test are caused by fiveam lingering behind master, I use there `for-all*` macro that has been merged a few days back.

PR #1270, opened 2022-02-18
core: read-command-using-keystrokes: sanitize the accelerator result

dkochmanski:
LOOKUP-KEYSTROKE-COMMAND-ITEM may return the gesture if no command can be found
under the keystroke. We don't want to pass a gesture to ENSURE-COMPLETE-COMMAND
so just in case we check the result. This allows associating keystrokes with
arbitrary actions with add-keystroke-to-command-table :function.


admich:
Maybe we can define a macro `define-keystroke-action` that expand to  `add-keystroke-to-command-table`

PR #1271, opened 2022-02-19
window-viewport-position

admich:
The implementation was wrong. From the specification window-erase-viewport:

> Returns two values, the x and y position of the top-left corner of the CLIM
> stream pane window's viewport. If the window is not scrollable, this will return
> the two values 0 and 0.


dkochmanski:
lgtm, thanks

PR #1274, opened 2022-03-12
command-accessible-in-command-table-p: specification conformance

admich:
For the specification COMMAND-ACCESSIBLE-IN-COMMAND-TABLE-P must
return the command table in which the command was found or NIL if the
command was not found.

Before this commit it return T or NIL.

With this commit also COMMAND-PRESENT-IN-COMMAND-TABLE-P return the
command table or NIL. I think this adhere to the specification where
is written that COMMAND-PRESENT-IN-COMMAND-TABLE-P returns true if the
command is present in the command table but the specification doesn't
say that it must return T.

### DIFF
--- a/Core/clim-core/commands/tables.lisp
+++ b/Core/clim-core/commands/tables.lisp
@@ -164,8 +164,10 @@
           (map-func command-table)))))
 
 (defun command-present-in-command-table-p (command-name command-table)
-  (let ((ht (slot-value (find-command-table command-table) 'commands)))
-    (nth-value 1 (gethash command-name ht))))
+  (let ((table (find-command-table command-table)))
+    (let ((ht (slot-value table 'commands)))
+      (when (nth-value 1 (gethash command-name ht))
+        table))))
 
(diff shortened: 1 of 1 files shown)


admich:
Eventually if we want that `command-present-in-command-table-p` continues to return `t` we can use the following code but I think that the code in the commit is better: 

```
(defun command-present-in-command-table-p% (command-name command-table)
  (let ((table (find-command-table command-table)))
    (let ((ht (slot-value table 'commands)))
      (when (nth-value 1 (gethash command-name ht))
        table))))

(defun command-present-in-command-table-p (command-name command-table)
  (when (command-present-in-command-table-p% command-name command-table)
    t))

(defun command-accessible-in-command-table-p (command-name command-table)
  (or (command-present-in-command-table-p% command-name command-table)
      (some #'(lambda (table)
                (command-accessible-in-command-table-p
                 command-name
                 (find-command-table table)))
            (command-table-inherit-from (find-command-table command-table)))))
```

PR #1279, opened 2022-05-05
Refactor render

dkochmanski:


dkochmanski:
Still much to do, but this branch is tested against regressions and a complete chunk that fixes numerous issues, so I'm going to merge it now.

For now we replace eager repaint-sheet with dispatch-repaint, so coordinate swizzling will corrupt output for a short period of time (until the command is executed). I have a plan how to have a cake and eat it too.

PR #1291, opened 2022-05-27
A variant of drawing-benchmark; scale/pan, multiple inks at once, recording on/off.

mgsouth:
Selectable sets of scale/translate transforms, multiple inks visible at
same time (it makes a difference), output recording on/off.

Will be using this as a test case for an issue.


Examples/xforms-benchmark.lisp
```diff
@@ -0,0 +1,210 @@
...
+          (setf (aref inks i j k)
+                (make-rgb-color (float (/ (1+ i) color-planes))
+                                (float (/ (1+ j) cols))
+                                (float (/ (1+ k) rows)))))))
+
+    ;; XFORM-ITERATIONS should be an even number for each loop; this will
+    ;; prevent having an unmatched set of do/undo transformations.
+    (do* ((loop-cnt 0 (1+ loop-cnt))
+          (xform-iterations 0 (ash 1 (* loop-cnt 2)))
+          (s nil))
```
dkochmanski, line 107: please remove unused variable

Examples/xforms-benchmark.lisp
```diff
@@ -0,0 +1,210 @@
...
+
+        ((> xform-iterations max-transforms))
+
```
dkochmanski, line 110: there is no need for empty lines in the single top-level form (applies to other places too)

Examples/xforms-benchmark.lisp
```diff
@@ -0,0 +1,210 @@
...
+      ;; Setup--concatenate a bunch of scaling & panning transformations.
+      (setf (sheet-transformation stream) +identity-transformation+)
+      (window-clear stream)
+      ;;(finish-output stream)
```
dkochmanski, line 114: why is this commented out?

Examples/xforms-benchmark.lisp
```diff
@@ -0,0 +1,210 @@
...
+              (let* ((x (+ 20 (* xn 100)))
+                     (y (+ (* yn 100) 50)))
+                (draw-rectangle* stream
+                                 x y (+ x 50) (+ y 50)
+                                 :ink (aref inks (mod n color-planes) xn yn)
+                                 :filled t)
+                (incf n)
+                ;;(draw-text* stream "ABCDEFG" x (+ y 70) :ink ink)
+                ;;(incf n)
+                ))))
```
dkochmanski, line 186: some leftover test comments

Examples/xforms-benchmark.lisp
```diff
@@ -0,0 +1,210 @@
...
+;;;
+;;; (c) Copyright 2022 by Michael South <msouth@msouth.org>
+;;;
+;;; Based on "drawings-benchmark.lisp"
+;;;  (c) Copyright 2006 by David Lichteblau <david@lichteblau.com>
+;;;
+;;; ---------------------------------------------------------------------------
+;;;
+
+(in-package #:clim-demo)
```
dkochmanski, line 13: please create a separate package for the demo

dkochmanski (changes requested):
thanks, I like the idea of adding more options to benchmarks. that said I think that a good way forward would be to also preserve previous features (used inks and shapes make a difference too). I've left a few comments, but I did not review this demo in great detail, so only outstanding things are noticed.

PR #1292, opened 2022-05-28
Implement pointer scrolling handling for generic-option-pane

### LINKED ISSUE #1214: implement scroll interaction over the option pane gadget
Currently the option pane is activated only after pressing it - then it is possible to select the appropriate option in the popup.
This ticket requires writing a handle-event method specialized on the class `generic-option-pane` and `pointer-scroll-event` that:

- on scroll up selects the option above (if any, for topmost it is a no-op)
- on scroll down selects the option below (if any, for bottmost it is a no-op)

The pane in question is defined in `Core/clim-core/gadgets/concrete.lisp:1273`.  The value should be done as
 `(setf (gadget-value pane) <the-new-value> :invoke-callback t)`.

Notes:
- the gadget specification: http://bauhh.dyndns.org:8000/clim-spec/30-4.html#_1845
- this gesture should work probably only for the option panes in exclusive mode
- the code in the module checks whether the value slot is bound - this should be handled too
- in `Examples/demodemo.lisp` there is `option-test` near the bottom, it may be a good start point for tests

defaultxr:
Fixes #1214.

This seems to be the clearest way to implement this feature, however it does traverse the `items` 3 separate times (first for `position`, then for `length`, and finally for `nth`). It may be possible to reduce this to just once with a `loop` or similar, but that idea felt like "premature optimization" to me, especially for a gadget that seems like it won't usually have a number of items high enough that efficiency would be a real issue. Additionally, the `loop` form would probably be a lot more complex an implementation than this.


Core/clim-core/gadgets/concrete.lisp
```diff
@@ -1574,6 +1574,20 @@ if INVOKE-CALLBACK is given."))
...
 
+(defmethod handle-event ((pane generic-option-pane) (event pointer-scroll-event))
+  (when (list-pane-exclusive-p pane)
+    (let ((delta-y (pointer-event-delta-y event))
+          (items (list-pane-items pane)))
+      (if (slot-boundp pane 'value)
+          (let* ((old-position (position (gadget-value pane) items
+                                         :test (list-pane-test pane)
+                                         :key (list-pane-value-key pane)))
+                 (new-position (alexandria:clamp (+ old-position (signum delta-y)) 0 (1- (length items)))))
```
dkochmanski, line 1585: the symbol clamp is imported to the climi package already.

Core/clim-core/gadgets/concrete.lisp
```diff
@@ -1574,6 +1574,20 @@ if INVOKE-CALLBACK is given."))
...
+            (unless (= old-position new-position)
+              (setf (gadget-value pane :invoke-callback t) (nth new-position items))))
+          (when (and items (plusp delta-y)) ;; if no item is selected and we're scrolling down, select the first one.
```
dkochmanski, line 1588: please put the comment above the form, or use single `;` for "after-line" comments

dkochmanski (changes requested):
I don't mind a few lists traversals, if it bites us then we could optimize. I had two stylistic remarks, but otherwise looks good to me. thanks!

defaultxr:
Implemented your corrections, thanks.

Core/clim-core/gadgets/concrete.lisp
```diff
@@ -1574,6 +1574,20 @@ if INVOKE-CALLBACK is given."))
...
   (disarm-gadget pane))
 
+(defmethod handle-event ((pane generic-option-pane) (event pointer-scroll-event))
+  (when (list-pane-exclusive-p pane)
+    (let ((delta-y (pointer-event-delta-y event))
+          (items (list-pane-items pane)))
+      (if (slot-boundp pane 'value)
+          (let* ((old-position (position (gadget-value pane) items
+                                         :test (list-pane-test pane)
+                                         :key (list-pane-value-key pane)))
```
dkochmanski, line 1584: a brief testing reveals a small defect. I've tried `option-test` from clim examples and I've landed in the debugger when I've scrolled the second box, that's because the value NIL is not of the expected type NUMBER (old-position is bound to NIL here).

Here's why this is happening:
- the option-test demo has a bug, because the second pane is defined as:
```
(option-pane-2 :option-pane
                         :value "Option 1"
                         :items '("Option 1" "Option 2" "Option 3" "Option 4" "Option 6" "Option 7")
                         :value-changed-callback (constantly nil))
```

since the default test is EQL, the current value is not the member of ITEMS. That's a bug in a demo, however I'd like for this code to not land the user in a debugger here.

I've also checked, and the value in a value-gadget is always bound (by default to NIL), so checking whether the slot is bound is not necessary. how about

```
(if-let ((old-position …))
  (scroll)
  (initiate)
```

?

also we could drop the test for whether delta-y is plusp in the second test and simply set to car items, that perhaps would be more intuitive?

dkochmanski (changes requested):

dkochmanski:
also, I've just checked in the specification - items are a sequence (not necessarily the list) -- while `position` and `length` are fine to use, `nth` should be replaced with `elt` and `(and items (plusp ...))` should be replaced with `(and (emptyp items) …)`

dkochmanski:
please import it.

defaultxr:
Alright, I've implemented those suggestions and reformatted the code a bit, let me know if you spot any other issues.

dkochmanski (approved):

dkochmanski:
lgtm, thanks!

PR #1298, opened 2022-07-03
Wayland Backend early proof-of-concept

cmack:
# DO NOT MERGE

This is an early proof of concept of a Wayland backend which I intend to implement using EGL and OpenGL at first.

I've written more detailed notes in a README.org in `Backends/Wayland` but this PR is intended for early review so that I can correct my lapses in understanding earlier than later.


Backends/Wayland/package.lisp
```diff
@@ -0,0 +1,39 @@
+(in-package #:common-lisp-user)
+
+(uiop:define-package #:clim-wayland
+    (:use #:clim #:clim-lisp #:clim-backend)
+  (:local-nicknames (#:alx #:alexandria)
+                    (#:wlc #:com.andrewsoutar.cl-wayland-client)
+                    (#:wl-core #:com.andrewsoutar.cl-wayland-client/core)
+                    (#:wl-egl #:com.andrewsoutar.cl-wayland-client/egl-core)
+                    (#:xdg #:com.andrewsoutar.cl-wayland-client.protocol/stable/xdg-shell/xdg-shell))
```
dkochmanski, line 9: new extensions and backends should be named as `mcclim-foo`, in this case `mcclim-wayland`

Backends/Wayland/package.lisp
```diff
@@ -0,0 +1,39 @@
+(in-package #:common-lisp-user)
+
+(uiop:define-package #:clim-wayland
```
dkochmanski, line 3: please don't use uiop:define-package - in mcclim codebase uiop is rather discouraged (although there are a few uses in asd definitions and in the listener)

Backends/Wayland/package.lisp
```diff
@@ -0,0 +1,39 @@
...
+    (:use #:clim #:clim-lisp #:clim-backend)
+  (:local-nicknames (#:alx #:alexandria)
+                    (#:wlc #:com.andrewsoutar.cl-wayland-client)
+                    (#:wl-core #:com.andrewsoutar.cl-wayland-client/core)
+                    (#:wl-egl #:com.andrewsoutar.cl-wayland-client/egl-core)
+                    (#:xdg #:com.andrewsoutar.cl-wayland-client.protocol/stable/xdg-shell/xdg-shell))
+  (:import-from #:climi
+                #:standard-port
```
dkochmanski, line 11: head-ups - standard-port functionality will be merged with basic-port, so this class won't exist soon (see the branch modules)

Backends/Wayland/input.lisp
```diff
@@ -0,0 +1,119 @@
...
+  (format *debug-io* "close toplevel event ~s~%" *application-frame*)
+  ;; QQQQ this feels terrible but it seems *application-frame* isn't
+  ;; bound. Extremely hacky way to get top-level-frame-sheet
+  (let* ((sheet (frame-top-level-sheet
+                 (first (frame-manager-frames
+                         (find-frame-manager :port *wayland-port*)))))
+         (clim-event (make-instance 'window-manager-delete-event
+                                    :sheet sheet)))
+    ;; (break)
+    (distribute-event *wayland-port* clim-event)))
```
dkochmanski, line 66: doesn't this event indicate in any way which window is closed? if it does, then you may do a reverse lookup to check what sheet is associated with the mirror.

Backends/Wayland/mirror.lisp
```diff
@@ -0,0 +1,110 @@
+(in-package #:clim-wayland)
+
+;;; Mirror protocol?
+
+(defclass wayland-mirror ()
+  ((window :initarg :window :reader window))
+  (:default-initargs :window (alx:required-argument :window)))
```
dkochmanski, line 7: it may be worth to also manage in this instance the sheet associated with the mirror, this way the reverse lookup will involve only invoking a reader.

Backends/Wayland/port.lisp
```diff
@@ -0,0 +1,332 @@
...
+
+(defmethod port-set-mirror-geometry
+    ((port wayland-port) (sheet mirrored-sheet-mixin) region)
+  (alx:when-let ((mirror (sheet-direct-mirror sheet)))
+    (with-bounding-rectangle* (x1 y1 x2 y2 :width w :height h) region
+      (with-bounding-rectangle* (ox1 ox2 oy1 oy2) (sheet-mirror-geometry sheet)
+        (let ((window (window mirror)))
+          (when (some #'/=
+                      (list x1 y1 x2 y2)
+                      (list ox1 oy1 ox2 oy2))
```
dkochmanski, line 213: if wayland doesn't accept real numbers (i.e width as a ratio or a float), then you should check these here I think.

Backends/Wayland/port.lisp
```diff
@@ -0,0 +1,332 @@
...
+
+(defclass wayland-graft (graft)
+  ())
+
+(defmethod dispatch-event
+    ((sheet mirrored-sheet-mixin) (event window-configuration-event))
+  (format t "dispatch-event window-cfg-event: Not sure what should happen here...~%")
+  ;; I'm not sure if the is the right way. For now, let's handle it like an
+  ;; immediate-event-mixin
+  (handle-event sheet event))
```
dkochmanski, line 243: why the default dispatch-event doesn't work for you?

Backends/Wayland/port.lisp
```diff
@@ -0,0 +1,332 @@
...
+                                  :orientation orientation
+                                  :units units)))
+
+(defclass wayland-frame-manager (standard-frame-manager)
+  ((mirroring :initarg :mirroring
+              :initform nil
+              :reader mirroring)
+   (class-gensym :initarg :class-gensym
+                 :initform (gensym "WAYLAND-")
+                 :reader class-gensym)))
```
dkochmanski, line 310: I wouldn't cargo-cult the frame manager, using standard-frame-manager directly should be fine (then only top level sheet mixin gets a mirror by default)

dkochmanski (commented):
I took only a cursory glance, so these remarks are rather shallow.

Backends/Wayland/port.lisp
cmack: If remember right, I think it was distributing the event to an object which didn't implement the protocol. I'll look into this again.

cmack:
> I took only a cursory glance, so these remarks are rather shallow.

Thanks for this feedback! It's plenty for the next immediate actions

Backends/Wayland/mirror.lisp
```diff
@@ -0,0 +1,110 @@
...
+
+(defmethod realize-mirror ((port wayland-port) (sheet mirrored-sheet-mixin))
+  (format t "realizing mirror~%")
+  (let (
+         ;; QQQQ I'm still having conceptual ignorance on the lifecycle of graft
+         ;; mirror instantiation and realize mirror. I hope it's not too
+         ;; cargo-culted from CLX.
+        (mirror (sheet-direct-mirror (graft port))))
+    ;; we'd like this set earlier than the basic-port specialization
+    (setf (climi::%sheet-direct-mirror sheet) mirror)))
```
dkochmanski, line 73: graft is made with make-graft; it doesn't need mirror realization, because it always exists (it is the screen representation).

realize-mirror is called, when a mirrored sheet (usually top-level-sheet) is adopted by a graft (or a sheet that has graft as an ancestor).

Backends/Wayland/port.lisp
```diff
@@ -0,0 +1,332 @@
...
+
+(defmethod handle-event
+    ((sheet wayland-graft) (event window-configuration-event))
+  ;; Is this where I should set the egl-window width and swap buffers?
+  ;; Should this be handled by the medium? or te mirror?
+  (alx:when-let ((mirror (sheet-mirror sheet)))
+    (format *debug-io* "handle-event window-cfg-event: top level reconfigure? Should I resize all the children?~%")
+    ;; QQQQ Should I use port-set-mirror-geometry or is there some other way
+    ;; to trigger it?
+    ;; (break)
```
dkochmanski, line 253: why configuration event is called on a graft? it should be called on a top level sheet. and the handle event is already implemented in the core. it calls internal function update-mirror-geometry.

Backends/Wayland/port.lisp
```diff
@@ -0,0 +1,332 @@
...
+    (unless (eql +everywhere+ (window-event-region event))
+     (with-bounding-rectangle* (x1 y1 x2 y2 :width w :height h)
+         (window-event-region event)
+       (with-bounding-rectangle* (ox1 ox2 oy1 oy2)
+           ;; QQQQ why is this -5 1 -5 1 still
+           (sheet-mirror-geometry sheet)
```
dkochmanski, line 259: first - with-bounding-rectangle* accepts x1 y1 x2 y2, not x1 x2 y1 y2;
mirror [-5, -5] -> [-4, -4] (width/height = 1) represents a mirror that has no visible region. that works because it hides before its ancestor mirror.

sheet-mirror-geometry is modified by a handler in the core for this event, so the new dimension is set only after the appropriate method is called. new mirror geometry is returned by the function port-set-mirror-geometry (or whatever it is called), and indirectly propaged by the method mentioned earlier.

Backends/Wayland/port.lisp
```diff
@@ -0,0 +1,332 @@
...
+                     (list ox1 oy1 ox2 oy2))
+           (format *debug-io* "resize surface geometry new:~s current:~s~%"
+                   (list x1 y1 x2 y2)
+                   (list ox1 oy1 ox2 oy2))
+           ;; TODO: determine if we need CLX's ROUND-COORDINATE
+           (xdg:xdg-surface-set-window-geometry (slot-value (port sheet) 'surface) x1 y1 w h)
+
+           (wl-egl:wl-egl-window-resize
+            (wayland-egl-mirror-window mirror) w h x1 y1)
+           ;; Is this overkill?
```
dkochmanski, line 271: pretty much yes.

cmack:
@oliverdelancey I'm working hard at it as work and family life allow :) I've been reading up on the documentation and source myself trying make sure I'm not bypassing the core protocol in "bad" ways. There are also some things in the wayland client package I've been working to upstream.

I'm happy to have as much code review as you or anyone is willing to give. I'll try to push more of my local progress tonight.

Backends/Wayland/package.lisp
cmack: done

Backends/Wayland/package.lisp
cmack: done

Backends/Wayland/package.lisp
cmack: I think this is done upstream now and I've since rebased so that it's included.

oliverdelancey:
Awesome! Also I wasn't looking to rush you, good projects take time :) Just wanted to let you know that there's at least one person out here who is excited about the prospect of McCLIM + Wayland.

Backends/Wayland/README.org
```diff
@@ -0,0 +1,146 @@
...
+
+** Environment
+You'll need development library for wayland and mesa. For debian/ubuntu based distributions, it usually means the =*-dev= packages. On Guix, I'm doing this in a dedicated dev environment:
+#+begin_src sh
+guix shell sbcl mesa wayland wayland-protocols libxkbcommon
+#+end_src
+
+I needed to start emacs in a way it could find the wayland protocal specs and libraries. You might not need this if you aren't sandboxing your environment:
+#+begin_src sh
+LD_LIBRARY_PATH="$GUIX_ENVIRONMENT/lib" emacs &
```
timmydo, line 36: I think setting LD_LIBRARY_PATH on Guix is considered to be a hack. Guix binds $LIBRARY_PATH to where the libs are symlinked. You could put something like this in your sbclrc (or do something more fancy with LIBRARY_PATH):
```
(ql:quickload :cffi)
(push (format nil "~a/.guix-profile/lib/" (uiop:getenv "HOME")) cffi:*foreign-library-directories*)
```

cmack: Thanks. Yes, I agree it's mostly a quick one-off solution for now. If it is a "hack," it's been an extremely useful one for me in the past when debugging OpenGL calls in lisp with gDebugger circa 2006-8 or so.

I'll give it a try soon, though.

PR #1301, opened 2022-07-26
Define indirect-ink-ink as a generic function

contrapunctus-1:
This will allow subclassing `indirect-ink` to provide different implementations of `indirect-ink-ink`, which would be desirable for the indirect-ink-based theme library I'm working on, with (hopefully) no impact on McCLIM.

### DIFF
--- a/Core/clim-basic/drawing/design.lisp
+++ b/Core/clim-basic/drawing/design.lisp
@@ -272,8 +272,9 @@
 (defun indirect-ink-p (design)
   (typep design 'indirect-ink))
 
-(defun indirect-ink-ink (design)
-  (check-type design indirect-ink)
+(defgeneric indirect-ink-ink (design))
+
+(defmethod indirect-ink-ink ((design indirect-ink))
   (with-slots (dynamic-variable-symbol default-ink) design
     (if (boundp dynamic-variable-symbol)
         (symbol-value dynamic-variable-symbol)
--- a/Core/clim-basic/drawing/design.lisp
+++ b/Core/clim-basic/drawing/design.lisp
@@ -84,7 +84,7 @@
 ;;;
 ;;;   INDIRECT-INK                                                    [class]
 ;;;   INDIRECT-INK-P ink                                              [predicate]
-;;;   INDIRECT-INK-INK ink                                            [function]
+;;;   INDIRECT-INK-INK ink                                            [generic function]
 ;;;
 ;;;      Indirect inks are underspecified in the standard. Class is exported for
 ;;;      specialization. Predicate INDIRECT-INK-P is defined. Initargs:


dkochmanski:
why the current implementation is insufficient?

contrapunctus-1:
Each indirect ink defined involves defining two special variables. For instance, I have 17 inks at the moment, and this number is likely to grow. While this can be done concisely by e.g. defining the special variables using a `macrolet` and a `loop`, that does not lend to very readable code. Doing it in the idiomatic, readable, and explicit way entails that I litter my code with ≥34 `defvar`s and place ≥34 symbols in my `(defpackage ... (:export ...))`. In any case, client code must import at least as many symbols as the inks they desire.

I'd prefer to implement lookup in a hash table, encapsulated in a `theme` class, so both me and any client application deals with far fewer symbols.

dkochmanski:
if you have 17 inks, then you still need to export 17 inks, I don't see how using a hash table mitigates that. the symbol behind the indirect ink is a detail that is not important for the code client. I imagine, that you think about something like:

```
(defmethod indirect-ink-ink ((ink what-is-the-specialization?))
  (gethash ink (color-table *theme*)))
```

vs the default method, where the concrete ink is stored in the ink instance (i.e in a default slot, or in a dynamic variable).

but the user still needs to know variables like ~*primary-foreground*~ or ~*primary-background*~ etc (reiterating - 17 exports is necessary).

Then there is the question: how do you customize the theme?
I imagine, that what you want to do is (possibly with some syntactic sugar):

```
(setf (gethash *primary-foreground* (color-table *theme*)) +red+)
```

but how is this better than (possibly with added accessor):
```
(setf (slot-value *primary-foreground* 'default-ink) +red+)
```
?

re gensym, we allow the dynamic variable to be nil in that case (a little of code).
I don't principally object making `indirect-ink-ink`, but I don't see much utility in that (compared to making the current indirect ink implementation more customizable).

contrapunctus-1:
The idea was to use keywords as hash table keys, so rather than 17 or 34, _no_ symbols need be exported for inks. That would be the solution I'd like best.

Additionally, making `indirect-ink-ink` generic would be the most flexible solution that could be provided (for use cases beyond my theme library) within the constraints of McCLIM's indirect ink protocol.

> re gensym, we can allow the dynamic variable symbol to be nil in that case (a little of code).

Something like this?

```lisp
(defclass indirect-ink (design)
  ((dynamic-variable-symbol :initarg :symbol
                            :initform nil)
   (default-ink :initarg :default
                :initform (error "required")
                :accessor default-ink)))

(defun indirect-ink-ink (design)
  (check-type design indirect-ink)
  (with-slots (dynamic-variable-symbol default-ink) design
    (if (and dynamic-variable-symbol
             (boundp dynamic-variable-symbol))
        (symbol-value dynamic-variable-symbol)
        default-ink)))
```

It could work and would be better than the current situation of `ink × 2` exported symbols, but not as nice as this PR, which would result in `ink × 0` symbols for my use case, and more flexibility for any other use cases.

PR #1303, opened 2022-08-03
frame-managers: use make-pane-1 instead of make-instance

admich:
When there is menu or pointer-documentation-pane in a frame the top
level vrack layout pane was created using MAKE-INSTANCE instead of
MAKE-PANE-1, in this way it was created an instance of the abstract
class and not an instance of the concrete class choosed by the frame
manager.


dkochmanski:
lgtm, thanks!

PR #1307, opened 2022-08-16
format-items: fix default printer

admich:
In format-items the default for printer cannot be provided in the lambda list
otherwise when the actual printer function is constructed in the let
form the case when presentation-type is supplied and printer isn't
supplied is never selected.


dkochmanski:
lgtm, thanks!

PR #1327, opened 2022-12-31
define :complete gesture as #\tab not :TAB

admich:
otherwise the TAB completion doesn't work.

### DIFF
--- a/Core/extended-input/standard-gestures.lisp
+++ b/Core/extended-input/standard-gestures.lisp
@@ -17,7 +17,7 @@
 
 (define-gesture-name :abort         :keyboard             (#\c :control))
 (define-gesture-name :clear-input   :keyboard             (#\u :control))
-(define-gesture-name :complete      :keyboard             (:tab))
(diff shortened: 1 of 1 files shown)


dkochmanski:
I don't mind changing that to the character later, but first I'd like to troubleshot why key names doesn't work here.

admich:
I think that the problem is that when McCLIM check if the event is a :complete event  with `event-matches-gesture-name-p` the event is no more the raw event because the following method of `stream-process-gesture` has returned only the character:

```
  (defmethod stream-process-gesture
    ((stream standard-extended-input-stream) gesture type)
  (declare (ignore type))
  (typecase gesture
    (key-press-event
     (if-let ((character (and (zerop (event-modifier-state gesture))
                              (event-char gesture))))
       (values character 'character)
       (values gesture (type-of gesture))))
    (pointer-event
     (values gesture (type-of gesture)))
    (character
     (values gesture 'character))))
```

PR #1330, opened 2023-01-14
map-over-frames: always return nil

admich:
MAP-OVER-FRAMES must always return NIL also when FRAME-MANAGER keyword argument is provided.

Otherwise, for example, FIND-APPLICATION-FRAME doesn't work with FRAME-MANAGER keyword argument provided.


dkochmanski:
good point, lgtm, thanks.

PR #1331, opened 2023-01-14
fix clx-frame-manager adopt-frame for :dock frame

admich:
The xwindow is in the variable WINDOW and not MIRROR


dkochmanski:
lgtm, thanks

